PR #7, opened 2013-07-15
Replace "adds a restriction restricting" with "further restricts" for better wording

### LINKED ISSUE #6: Phrasing "adds a restriction restricting" on front page
I think the wording  "adds a restriction restricting" is ugly and adds more perceived complexity to GPL licenses. It is found in index.html at: "V3 is similar to V2, but adds a restriction restricting use in hardware that forbids software alterations.".

bxt:
I replced "adds a restriction restricting" with "further restricts" in the index page, for better wording, hope to fix #6.


haacked:
Thanks a lot! :sparkles:

PR #9, opened 2013-07-15
Add license name to MIT license

ajepst:
It's very helpful/timesaving to people considering making use of a project for the license text to have the title, so they don't have to do a search to confirm what the license is. Therefore, the portion of the MIT license that gets copied to clipboard should have the license title.


haacked:
:thumbsup: and :sparkles:

PR #11, opened 2013-07-15
Use Affero GPL text instead of GPLv3

rahearn:
The full license text on the Affero GPL page was regular GPLv3 instead of AGPL.

Copied AGPL code directly from http://www.gnu.org/licenses/agpl-3.0.txt


benbalter:
Great catch. Thanks for taking the time to bundle this up as a pull request.

PR #14, opened 2013-07-15
Licenses/ should link to the no-license license, not to /no-license

benbalter:
When you click the "no license" license from /licenses/ it should go to the license text, not to the /no-license/ meta page linked from the index.

This is a result of a permalink conflict, now resolved.

### DIFF
--- a/licenses/no-license.html
+++ b/licenses/no-license.html
@@ -1,6 +1,6 @@
 ---
 layout: license
-permalink: /no-license/
+permalink: /licenses/no-license/
 class: license-types
 title: No License
 


haacked:
Need to fix index.html too since it links to "no-license" but probably should link here, right?

benbalter:
I'd think we'd want the text "What if I don't want to choose a license? You don't have to." to point to the ["here's what happen if you don't choose a license page"](http://choosealicense.com/no-license/), and from the list of licenses, we'd want to point to the ["no-license" license text + attributes](http://choosealicense.com/licenses/no-license), no?

Other option would be to collapse /no-license into the sidebar of the "no-license" license.

May also make sense to link from /no-license to /licenses/no-license?

haacked:
@benbalter you are correct sir!

PR #16, opened 2013-07-15
Remove include-license (covered by include-copyright)

dpen2000:
I think this line should have been removed in 3c29d5da87e560c3cc39c36753c58a60e79664ce


benbalter:
Great catch, and thanks for taking the time to dig through the history to ferret things out.

/cc @Haacked

PR #15, opened 2013-07-15
WTFPL

sabas:
Popular license and valid alternative to public domain.
OpenStreetMap editors Potlach2 and iD are two examples of software with this license.

(I'm pulling this for the lulz, I don't think it will be merged :-) )

PR #17, opened 2013-07-15
The WTFPL, also a choice

patosar:


jpmckinney:
@superpatosainz This duplicates #15. Close?

anaisbetts:
Dupe of #15

benbalter:
https://github.com/github/choosealicense.com/pull/210 which also implements, is currently open for discussion.

anaisbetts:
> This license was intentionally left off of the list of choices, because one of the "Whatever you want" things to do, is "Sue the author for liability", which Seems Bad™. Thanks for submitting it though!

Still true ^^  - we shouldn't encourage people to choose licenses that open them up to legal liability

benbalter:
> we shouldn't encourage people to choose licenses that open them up to legal liability

:100:% agree. The idea proposed in https://github.com/github/choosealicense.com/issues/273 (which would need to be resolved before it could be added), would be a class of hidden licenses, that would not be recomended, but, if for example, you come across a WTFPL-licensed project, we could at least provide a google-able page to explain what the use can and can't do (and why they shouldn't choose it for the reasons you mentioned). But again, wouldn't be presented as an option / recomendation anywhere.

anaisbetts:
@benbalter Into it

PR #18, opened 2013-07-15
Update eclipse.txt - include disclose-source

dpen2000:
I think this was a mistake removal in 3c29d5da87e560c3cc39c36753c58a60e79664ce


dpen2000:
Will resubmit. This would reintroduce previously removed line.

PR #19, opened 2013-07-15
Update eclipse.txt - include disclose-source

dpen2000:
I think this was a mistake removal in 3c29d5da87e560c3cc39c36753c58a60e79664ce. I'm not a legal expert though.

### DIFF
--- a/licenses/eclipse.txt
+++ b/licenses/eclipse.txt
@@ -10,6 +10,7 @@ how: Create a text file (typically named LICENSE or LICENSE.txt) in the root of 
 source: http://www.eclipse.org/legal/epl-v10.html
 
 required:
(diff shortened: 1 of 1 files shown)


haacked:
Ack! I think you're right. Let me double check.

haacked:
Yep.

> A Contributor may choose to distribute the Program in object code form under its
> own license agreement, provided that:
> ...
>  states that source code for the Program is available from such Contributor, and informs licensees how to obtain it in a reasonable manner on or through a medium customarily used for software exchange.

I deleted the wrong line in that commit. Looks like submitted the other half of the fix here: 73c1efe44f95144f4bad243baa144ccc7d7bd479

Thanks and nice catch!

PR #22, opened 2013-07-15
It's about licenses: let's get them right!

steko:
"Creative Commons 3.0 license" is unclear. The link is correct though.


haacked:
Thanks! :sparkles: We got it right in the footer of the site, but not the readme. Thanks for the PR!

PR #23, opened 2013-07-15
Strip html from description when displaying in meta tag

dpen2000:
This fixes the issue of description of the BSD license (which includes links) showing at top of: http://choosealicense.com/licenses/bsd/


haacked:
Thanks for all the help @dpen2000! Another fine PR.

PR #8, opened 2013-07-15
Add description for the Affero GPL license

waweber:
This will add a description to the Affero GPL entry in the licenses page.

Signed-off-by: Wes Weber wweber@wesweber.com

PR #24, opened 2013-07-15
Added Note About Affero GPL to Disclose Source message.

aatxe:
#8 adds a description of the AGPL license. This goes with that by adding a note about the AGPL (similar to the one for the LGPL) inside the Disclose Source info message.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -12,7 +12,7 @@ rules:
       description: Indicate significant changes made to the code.
       label: State Changes
     disclose-source:
-      description: Source code must be made available when distributing the software. In the case of LGPL, the source for the library (and not the entire program) must be made available.
+      description: Source code must be made available when distributing the software. In the case of LGPL, the source for the library (and not the entire program) must be made available. In the case of the Affero GPL (AGPL), the definition of distributing is expanded to include networked access.
       label: Disclose Source
     library-usage:
       description: The library may be used within a non-open-source application. 


haacked:
Seems legit. /cc @hoolio @benbalter 

@aaronweiss74 did you try this locally? I'm curious what the tooltip looks like with extra verbage. Screenshot?

I'm a little concerned because #32 also adds to this category.

benbalter:
Between this and #32, that tool tip is about to become longer than the MIT license itself. Comments on #32, but think license specific stuff is best left associated with the license. Tool tips are for what licenses have in common, not what makes them different.

aatxe:
I think @benbalter is probably right. I jumped on the change for consistency with the LGPL notice, but I did have the same concern in the back of my mind about its length.

@Haacked I didn't try it locally, but if you still want to see it, I can do so.

adriank:
+1

_config.yml
```diff
@@ -12,7 +12,7 @@ rules:
       description: Indicate significant changes made to the code.
       label: State Changes
     disclose-source:
-      description: Source code must be made available when distributing the software. In the case of LGPL, the source for the library (and not the entire program) must be made available.
+      description: Source code must be made available when distributing the software. In the case of LGPL, the source for the library (and not the entire program) must be made available. In the case of the Affero GPL (AGPL), the definition of distributing is expanded to include networked access.
```
afeld, line 15: Isn't the addition more appropriate under the `distribution` heading?

haacked, line 15: Yeah, that kind of makes sense. It allows us to keep the primary description brief.

haacked:
Closing this as I'm convinced by @benbalter's argument. :smile:

PR #25, opened 2013-07-15
Remove "private-use" rule

jpmckinney:
All licenses permit private use, so this information doesn't help in choosing a license. See #1 for more discussion.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -34,9 +34,6 @@ rules:
     sublicense: 
       description: You may grant a sublicense to modify and distribute this software to third parties not included in the license.
       label: Sublicensing
-    private-use: 
-      description: You may use and modify the software without distributing it.
-      label: Private Use
     patent-grant:
       description: This license provides an express grant of patent rights from the contributor to the recipient.
       label: Patent Grant
@@ -60,7 +57,4 @@ rules:
     sublicense:
       description: You may not grant a sublicense to modify and distribute this software to third parties not included in the license.
       label: Sublicensing
-    private-use:
-      description: You may not use or modify the software without distributing it.
-      label: Private Use
 
--- a/_config.yml
+++ b/_config.yml
@@ -60,4 +60,7 @@ rules:
     sublicense:
       description: You may not grant a sublicense to modify and distribute this software to third parties not included in the license.
       label: Sublicensing
+    private-use:
+      description: You may not use or modify the software without distributing it.
+      label: Private Use
 
--- a/licenses/agpl.txt
+++ b/licenses/agpl.txt
@@ -21,6 +21,7 @@ permitted:
 forbidden:
   - no-liability
   - no-sublicense
+  - private-use
 
 using:
   Some Project: "#"
--- a/licenses/agpl.txt
+++ b/licenses/agpl.txt
@@ -21,7 +21,6 @@ permitted:
 forbidden:
   - no-liability
   - no-sublicense
-  - private-use
 
 using:
   Some Project: "#"
--- a/licenses/artistic.txt
+++ b/licenses/artistic.txt
@@ -18,7 +18,6 @@ permitted:
   - modifications
   - distribution
   - sublicense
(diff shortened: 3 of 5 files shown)


jpmckinney:
Hmm, on second thought, probably not 100% correct, as private use is allowed under Affero, you just can't privately modify it without also distributing the changes. Maybe just remove "private use" as a "permitted" action, and find another way to efficiently communicate Affero's obligations?

aatxe:
Whether or not the AGPL allows private use depends heavily on your definition of private. The only special requirement it has is that if the software is networked (or used in a networked application) that anyone who can connect to or in some way remotely access said software must have access to the source. If you have a modified server derived from something under the AGPL and only ten people can see it, those ten people have to have source access.

aatxe:
Both #8 and #24, the latter of which is mine, attempt to address problems with how the AGPL is displayed.

haacked:
Do you think "private modifications" is closer to the spirit?

jpmckinney:
@aaronweiss74 I think we commented at same time (see my above comment). This pull request should probably just stick to removing the "Permitted: Private Use" clauses, which are currently confusing (see #1). +1 #8 and #24 

@Haacked Private modifications is closer, but I'm not expert to say if it's "close enough."

aatxe:
@Haacked The problem is really that the word private is misleading here. The purpose of the Affero license is simply to expand the definition of "access" to include networked access. 

@jpmckinney you're correct, we did indeed. I'm in agreement on just remove the permits private use thing.

haacked:
I would not object to that. :thumbsup:

jpmckinney:
K, updated pull request and changed issue title/description.

haacked:
Thanks @jpmckinney I'm going to let it sit for a bit to allow others time to comment before merging it.

ghost:
> All licenses permit private use, so this information doesn't help in choosing a license.

Agree, private-use is misleading. It's not accurate for AGPL either.
I think that if disambiguation pages are the way to go, it's a better place to express the difference. If not, or additionally, the note on AGPL could contain a phrase on it.

afeld:
:thumbsup:

haacked:
@benbalter any thoughts on this?

benbalter:
Glad to revisit, but I believe the original intention of the `private-use`
field was to clarify what you can and can't do with the software. Yes, we
all know that X license permits private use, but for the uninitiated,
that's not a logical conclusion, or even if it is, we'd want to lower the
cognitive burden.

haacked:
Thanks @benbalter! So any thoughts on `private-use` ass applied to Affero?

benbalter:
AGPL permits private use, just like the rest. The difference with AGPL is it redefines the triggering threshold for distribution to include hosted services, which GPL does not.

It looks like AGPL doesn't have a blurb, a simple, "This is GPL, but distribution also includes hosted services" or similar should be enough to disambiguate IMHO, but I'd move that to a different issue.

haacked:
Ok, so it seems like the change we need is to _add_ private use to AGPL.

jpmckinney:
If that's the way we're going, we need to add private use to all the licenses. Right now, it's just on public domain, no license, artistic license.

haacked:
@jpmckinney: yep, I did that here: https://github.com/github/choosealicense.com/issues/115

afeld:
It seems more useful to point out the exception to the rule, rather than having some attribute on all but one.

benbalter:
> It seems more useful to point out the exception to the rule, rather than having some attribute on all but one.

@Afeld, see the reasoning [above](https://github.com/github/choosealicense.com/pull/25#issuecomment-27343365)

jpmckinney:
I'm happy to close this in favor of #115. (Or a maintainer can do that for me.)

haacked:
Once we merge #115 it'll close this one automatically because I added the comment `closes #25`. I just need someone to review and give me dat :squirrel:

PR #26, opened 2013-07-15
Adds a way to copy the license without using ZeroClipboard

elcuervo:
This is an alternative way of using the "Copy to clipboard" button when the user does not have Flash installed.
Selects the license when the browser does not have Flash installed so people can easily copy & paste it.


haacked:
This might be a neat feature to build into ZeroClipboard if it's not already there. /cc @jonrohan

haacked:
Thanks for the PR! I'm having trouble testing this out. I tried it on Chrome on a Mac and Chrome on Windows.

I went to `about:plugins` and disabled Flash and when I click the button, I get the following errors:

```
Uncaught TypeError: Cannot read property 'style' of undefined ZeroClipboard.js:1
ZeroClipboard.reposition ZeroClipboard.js:1
ZeroClipboard.setCurrent ZeroClipboard.js:1
_elementMouseOver
```

In looking at the code, it looks like it selects the text, but it doesn't tell the person they still need to right click and copy it. Users might not realize they're getting the "fallback" behavior so it might be worth notifying them in this case that because Flash is disabled or uninstalled, they need to right click on the selected text and copy it.

jonrohan:
It's not currently in there. GitHub does/did something similar with the url field on repos. But this got some complaints.

elcuervo:
@Haacked I don't have flash installed in my system so I tested it without having to disable flash. I know about the need to copy the text anyways. I didn't add the tooltip. It's a hacky solution for a bigger problem, I know.
One "solution" I can think about it's catching the copy command if nothing is selected to copy the license... just thinking out loud here.
I sent this as a "good enough" solution since the copy button does not do anything when flash is not installed.
Should I continue with this PR or should we look for another approach?

haacked:
@elcuervo so in #45 @hobarrera mentioned that there's an HTML 5 clipboard API. I think it'd be great if we used that if the browser supports it, then fallback to ZeroClipboard, and then fallback to text selection.

elcuervo:
I thought about the html5 clipboard but since it not openly supported I went to the text selection.
Then let's do this:
- HTML5 clipboard API
- ZeroClipboard
- Text selection + tooltip with info on how to copy&paste

That looks good?  

## 

elCuervo

haacked:
Yes. :+1:

haacked:
@elcuervo I thought about this some more and it seems to me that the code as-is doesn't make anything worse. So I think it might make sense to merge this and make a separate PR for HTML5 clipboard API support.

elcuervo:
Roger that. I'll be adding the tooltip just to be more user-friendly in the next PR.

## 

elCuervo

haacked:
Yeah, I found a nice flash detection library that can detect the "disabled" state here: http://www.featureblend.com/javascript-flash-detection-library.html

Might be nice to use that. If you're up for it, I can test it. Otherwise when I find some time, I'll try it. But it might take a while.

elcuervo:
I will. That's better than my hacky way of checking for Flash. Thanks.

## 

elCuervo

JamesMGreene:
Hey all!

I'm already working on getting some better Flash detection and version checking built into the JS side of ZeroClipboard. It looks very similar to that of the referenced page, and most of it was extracted from the ultimate Flash JS lib: [swfobject](https://github.com/swfobject/swfobject/blob/master/swfobject/src/swfobject.js).  I'm pretty busy with moving and some [paying :dollar:] side work this month but I'm hoping to have my incoming PR done and merged within the next week or so, after which point we should be able to cut a new beta release if you'd like to try it out.

If you're seeking capabilities, then I would argue that the correct order is:
1. [ZeroClipboard](https://github.com/zeroclipboard/ZeroClipboard)
   -  As much as we all hate Flash, ZeroClipboard can copy without a user-initiated _browser_ "copy" event.
2. [HTML5 Clipboard API](http://www.w3.org/TR/clipboard-apis/)
   - It can copy based on a user-initiated _browser_ "copy" event but requires no other interaction as you can still programmatically select the text you want to insert. It also exists in oldIE with a _slight_ variation, if you care about oldIE.
3. **Bonus Option**: [Trello Clipboard](http://stackoverflow.com/a/17528590/471696) technique
   - Coming in a very close runner-up to the HTML5 Clipboard API is Trello's clever clipboard hack, which I recently mentioned in zeroclipboard/ZeroClipboard#171 strictly for documentation purposes.  Its major disadvantage is that it only works if the _browser_ "copy" event is triggered by keyboard (e.g. `Ctrl`+`C`).
4. Programmatic text selection + user-initiated _browser_ copy event.
   - This is the lowest common denominator... unless you're unkind and don't even pre-select the text for them! :astonished: If you're worried about the user shifting focus and unselecting the selected text, you can just add a button to programmatically select it again for them.

In all honesty, though, if you are going to be showing them the license text anyway (?), then you may even want to consider cutting Options 2 and 3 altogether as they don't buy you a whole lot more than Option 4 considering their added complexity.

If you're basing your order on something other than capabilities, then feel free to go in whatever order you so choose. :v:

P.S. Even though ZeroClipboard may never be able to use the HTML5 Clipboard API given [their small but fundamental functionality differences](http://lists.w3.org/Archives/Public/public-webapps/2013JulSep/0061.html), I am still planning to add some event facades into ZeroClipboard that would allow the user to use the same callback function for both ZeroClipboard and the HTML5 Clipboard API.  See zeroclipboard/ZeroClipboard#105 for an initial brain dump of code.

PR #27, opened 2013-07-15
Use travis for ci to verify builds work before merging

benbalter:
This will alert us via the Merge API if pull requests fail to build either due to Jekyll or CoffeeScript errors.


afeld:
[Doesn't seem to be building](https://travis-ci.org/github/choosealicense.com/pull_requests) for pull requests...changed the hook to use my token to see if that does the trick.

PR #31, opened 2013-07-15
Make consistent with featured change

haacked:
We made a change here: fb5546db7fe2855269a8796f4eec4ea9f0589dea

This should reflect that change.


benbalter:
:+1:

PR #32, opened 2013-07-15
point out weak file-based copyleft of MPL

Waquo:
The current description makes the MPL come off as if it had GPL-strength copyleft, which it doesn't.

http://www.mozilla.org/MPL/2.0/FAQ.html#virality

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -12,7 +12,7 @@ rules:
       description: Indicate significant changes made to the code.
       label: State Changes
     disclose-source:
-      description: Source code must be made available when distributing the software. In the case of LGPL, the source for the library (and not the entire program) must be made available.
+      description: Source code must be made available when distributing the software. In the case of LGPL, the source for the library (and not the entire program) must be made available. In the case of the MPL, only files explicitly placed under the MPL and changes to them must be made available, not new files added to the project nor the entire program.
       label: Disclose Source
     library-usage:
       description: The library may be used within a non-open-source application. 


haacked:
Seems legit. /cc @benbalter @hoolio 

#24 also adds text to "disclose source". It'd be nice not to make the tooltip super giant. There is a benefit to having it all in there though. If I was looking at one license's "disclose source" requirements, I might see that another has conditions I like better.

But we also want these tooltips to be short and sweet.

So an alternative might be to have _custom_ "disclose source" categories. That way each license has text specific to it.

```
disclose-source-gpl:
   description: current description + affero networked device clause.
   label: "disclose source"

disclose-source-mpl:
   description: this one
   label: "disclose source"

disclose-source-lgpl:
   description: this one
   label: "disclose source"

```

Would love to hear from others.

benbalter:
I'd :-1: making bespoke tooltips for each license. The whole point of the site is to simplify the process and create common points of comparison between the licenses, not to go down the long tail of nuanced interpretations.

If there's a license-specific note that is absolutely vital to understanding a license (e.g., to distinguish AGPL from vanilla GPL) lets put it on the license page. Either in the description, or as a note in the sidebar.

I'd really rather not get in the habit of making tool tips a :christmas_tree: for every legal edge case.

Alternatively, some sort of GPL disambiguation mechanism would solve.

haacked:
> If there's a license-specific note that is absolutely vital to understanding a license (e.g., to distinguish AGPL from vanilla GPL) lets put it on the license page.

So rather than put it in the tooltip, we put it in the license page? Maybe what we need to do is make the tooltip more general. For example,

> Source code must be made available when distributing the software. In some cases this might only cover the code under the license itself. In others it includes the whole work that makes use of the software. Check the license for more details.

aatxe:
@Haacked maybe something like...

> All or a portion of the source code must be made available to those with access to the software following distribution. The license determines whether the required disclosure is of all or a portion of the source. Definitions of "made available" and "distribution" are also set by the license. Check the license for more details.

I think it would better express the differences (and resultant lack of clarity) between GPL and AGPL (#24) while still remaining generic enough to describe other differences (e.g. MPL and GPL).

ghost:
I think this is an issue of reciprocity, not source disclosure. All licenses requiring it require the code to be under the same license one way or the other. Source disclosure under a license forbidding forking and so on is not an option.

Perhaps...

> Reciprocity
> Modifications of the software must be shared under an open source license, if the software is made available. In some cases only modifications to the source code itself must be shared, in other cases it includes code using the software. The requirement to share the code also depends how the software is made available. Check the license for more details.

I still don't think it expresses it well enough... Perhaps the choice of @aaronweiss74 of the phrase before last is better.

haacked:
Sorry for my long absence from this discussion. @benbalter or @talniv, any thoughts on updating the tool tips? In particular, look at the previous two comments and let me know what you think we should do if anything. :smile:

benbalter:
The issue is improvement vs. derivative work. How about:

> Source code must be made available when distributing the software. In some cases this might only cover **the original project and its improvements, in others it may also include projects which rely on the licensed code**. Check the license for more details.

or something?

ghost:
Well, on coverage, as accurate as any; it doesn't quite cover AGPL, which is not triggered at distribution, but network service access to modified code. Not much can be done I think, on that, it's for the details of each license to express it.

On the copyleft condition, source code made available is not accurate: source code available under a proprietary license is not a fulfillment of this condition. It's not "disclosure" and it's not "availability", it's "available under the same license"; or available under an open source license.

This is an important point, and IMO it _should_ be well documented and well understood. If it's not expressed here, in the source disclosure property, then another place for it is "sublicensing" property, which is very inaccurate and odd currently. I'll make a PR to fix that...

bkeepers:
This pull is no longer able to be merged. If you want to push a fix for the merge conflicts, we can reopen this.

PR #42, opened 2013-07-16
Leave out dev things.

spraints:
In #27, I noticed that `script/*` ended up [on the built site](http://choosealicense.com/script/cibuild). This removes them from the published site.


haacked:
Thanks @spraints!

PR #49, opened 2013-07-16
Darker text for better contrast - fixes #37

### LINKED ISSUE #37: Readability of license texts
The contrast of text to background when displaying a license text should be improved.

The text itself has a slightly darker grey as the background.

For the interested visitor and reader the license text should have either a darker text colour or bolder/thicker letters.

BTW: Thanks for that service! Especially the overview of the different licenses is really helpful.

jasonlong:
I think @torbjoernk has a fair point in #37 saying that the text contrast could be better. This pull darkens the body text a bit.

Before:
screen shot 2013-07-16 at 2 18 58 pm

After:
screen shot 2013-07-16 at 2 18 49 pm

### DIFF
--- a/css/application.css
+++ b/css/application.css
@@ -56,7 +56,7 @@ article, aside, details, figcaption, figure, footer, header, hgroup, menu, nav, 
 
 body {
   background: #f5f1ec url(../images/bg.jpg);
-  color: #7e7974;
+  color: #5c5855;
   font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
   font-size: 14px;
   line-height: 1.4;
@@ -233,7 +233,7 @@ strong {
   -ms-border-radius: 3px;
   -o-border-radius: 3px;
   border-radius: 3px;
-  color: #7e7974;
+  color: #5c5855;
   padding: 16px;
 }
 .see-more {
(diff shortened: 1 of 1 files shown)


cobyism:
Would it make sense to take things a step further and give each license a dedicated boundary around the license’s content? I’m thinking something along the lines of this:

2013-07-16 at 7 51 pm

I feel like it needs something like this to ensure that the license appears as though it is _clearly_ the most important piece of content on the page, and I feel like contrast (using the background as well as the foreground colour) is the way to make that happen. Thoughts?

jasonlong:
I kinda prefer it as is, but I can see your point @cobyism. If we did do something like this, I probably wouldn't go pure white though. On git-scm.com, I used a very light beige for the content container which looks nice. But I'd also rather not have this site start to look too similar to git-scm.

cobyism:
Yeah, not suggesting you go with that exact mockup by any means, just throwing an idea out there—it’s equally easy to argue that an enclosure isn’t needed here at all. It’s your site at the end of the day, so go with your gut :smiley:

haacked:
Let's try this out for now and we can tweak more later if we think it would benefit. :)

jasonlong:
Sounds good. Thanks for the feedback @cobyism. Maybe we'll circle back to this at some point.

Issue #54, opened 2013-07-17
Better handling of angle brackets in licenses


haacked:
The licenses we included are intended to be plain text. But since we render them on a website, we have issues with angle brackets. There are two cases we've seen for angle brackets.
1. To denote a template.

```
copyright <your name> <year> <month>
```
1. To denote a URL

```
for more info, visit <http://unlicense.org>
```

As of #53 we plan to manually replace angle brackets with square brackets. The problem we've encountered is that these licenses templates are the same one we use in .com when you pick a license and they are dropped as-is into a repository without the benefit of Jekyll processing.

So what I propose we consider, if possible:
1. License templates have the license in exactly the original plain text format.
2. When rendering to the web, replace angle brackets with &lt; and &gt;
3. Unless the contents start with "http:" or "https:" in which case we render it as a link. The copy and paste mechanism should still get the original text.

I have a feeling the inclusion of the angle brackets in the source might be a problem for Jekyll though, is that right @benbalter?

bhuga:
Turns out we need to do something here. I'm having trouble with the license picker when [http://...] is the same as [variable]. Bleh.

> License templates have the license in exactly the original plain text format.

These vary, so I would call this a non-goal. We need them standard; some had `{}` in the originals. I propose:
1. `[name]` is a variable
2. `<http://...>` is a URL. Rendered in web as a link, and as plain text with our without `<>` in copy-paste.
3. `{text}` is a non-interpolated variable in license text, akin to https://github.com/github/choosealicense.com/blob/gh-pages/licenses/gpl-v2.txt#L360

benbalter:
The delimiter is arbitrary, just need to decide on a standard.

> [name] is a variable

:+1:

> {text} is a non-interpolated variable in license text,

Got to double check that that doesn't trip up Jekyll, but I think we'll be fine (Jekyll uses `{{`)

> http://...

I don't want to get in the business of doing more work than we have to. Jekyll has a built in filter `xml_escape` which will convert everything to HTML entities for us, or we could just find and replace the `<`s as `&lt;`s. But don't know that we have to make it a hyperlink.

Otherside of the coin, if we start mixing HTML with plain text, is the copy button. The copy button will grab the HTML version (I believe) so we'd have to convert back to plain text on the fly (which is a pain).

I'd recommend leaving URLs as non-html, non-links for simplicity sake, but we may have to swap the HTML entities back to text before they hit the clipboard, which is significantly easier.

/cc @holman because .com

[...]

PR #55, opened 2013-07-17
Rework brackets

bhuga:
Potential solution to #54.

### DIFF
--- a/_includes/sidebar.html
+++ b/_includes/sidebar.html
@@ -4,7 +4,7 @@
     <div class='how-to-apply callout'>
       <h5>How to apply this license</h5>
       <p>
-        {{ page.how | replace:"<<","[" | replace:">>","]" }}
+        {{ page.how  }}
       </p>
       {% if page.note %}
       <p class="note">
--- a/README.md
+++ b/README.md
@@ -41,6 +41,16 @@ Licenses sit in the `/licenses` folder as markdown (`.md`) files. Each license h
 * `note` - The note field in the sidebar (optional)
 * `how` - How to use the license, also in the sidebar
 * `required`, `permitted`, `forbidden` - bulleted list of rules applicable to the license (see below)
+* `filename` - The filename to be created on GitHub.com when a repository is initialized with this license.
+
+The licenses on choosealicense.com are regularly imported to GitHub.com to be used as the list of licenses available when creating a repository. When we create a repository, we will replace certain strings in the license with variables from the repository. These can be used to create accurate copyright notices. The available variables are:
+
+* `[fullname]` - The full name or username of the repository owner
+* `[login]` - The repository owner's username
+* `[email]` - The repository owner's primary email address
+* `[project]` - The repository name
+* `[description]` - The description of the repository
+* `[year]` - The current year
 
 # Rules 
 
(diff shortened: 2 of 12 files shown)


haacked:
Looks good to me. @benbalter any concerns?

benbalter:
@bhuga It looks like the variable delimiter was changed in babcaa2a18043ea74aec05b2424fa6c469ee7115. Did you also change things on the .com side as well?

benbalter:
Also need to mirror the content filter in [`_includes/sidebar.html`](https://github.com/github/choosealicense.com/blob/gh-pages/_includes/sidebar.html#L7)

bhuga:
@benbalter I took it out so that I wouldn't have to change things on the .com side, and then ran into the variable/URL collision noted earlier. This is trying to bring the license files and .com into sync with no bugs.

I actually just removed the content filter in the sidebar. the `how:` fields of the YAML front matter don't have anything to replace.

benbalter:
As long as things don't :boom: when we re-vendor thing into .com (which we should do following the merge), :+1:.

Thanks for taking the time to do the deep dive on this one. Not the most exciting problem to tackle, but important to get right and will make things a lot easier going forward.

bhuga:
Yeah, little things like `<<url>>` on .com and missing FSF URLS here are hard to miss. Hopefully with this convention in place we'll be good for easy-peasy `cp licenses/*` in the future. Going to go ahead and merge this and I will try to get the .com pull updated today too.

gjtorikian:
Is there a list somewhere of all potential `{ }` variables? I'm doing some [lazy `gsub`ing right now](https://github.com/gjtorikian/addalicense.com/blob/master/server.rb#L92-L94) but I see new variables like `name of copyright owner` in this PR (which should probably just be `fullname` or something, no?).

bhuga:
@gjtorikian you would change to subbing on `[variable]` after this pull. `{variable}` is no longer meant to be interpolated.

I just found a few more cases where we were interpolating variables which were wrong. The GPL actually has an inducement to use itself at the end that I fixed, and I changed the artistic license to be correctly filled out on repo creation. Quick :eyes: on those changes are :cool:

benbalter:
Could punt to a different PR, but would be awesome to document A) what brackets we interpolate, and B) what variables we replace on the .com side, perhaps in the "adding a license" section of the readme so we don't have to rehash this again.

bhuga:
I'll do the README here. This was already one pull, let's do it right. Need to run just now though.

bhuga:
Okay, README is done. Any more thoughts?

benbalter:
:sparkles: :shipit:

_includes/sidebar.html
```diff
@@ -4,7 +4,7 @@
     <div class='how-to-apply callout'>
       <h5>How to apply this license</h5>
       <p>
-        {{ page.how | replace:"<<","[" | replace:">>","]" }}
+        {{ page.how  }}
```
haacked, line 7: Nitpick nanny here: There's 2 spaces after "how" there. :trollface:

bhuga, line 7: Fix pushed. Thanks for noticing. :dart:

gjtorikian:
What does .com do when a repo doesn't have, for example, a description? Replace the variable with an empty string?

bhuga:
Should be. It's actually whatever mustache does.

PR #59, opened 2013-07-19
Add license property 'include-copyright-binary'.

marnix:
Add an additional license property: "Must this license be part of a binary distribution?"

To make this clear, rename 'include-copyright' to 'include-copyright-source', so that we
can call the new property 'include-copyright-binary'.

Note that this commit does not yet set this property for the existing licenses-- that requires actually reading them. :-)

Background: In the course of my work it is very useful to know whether the binary form of someone else's product actually must be accompanied by a license, or whether it may be redistributed without.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ The licenses on choosealicense.com are regularly imported to GitHub.com to be us
 
 # Rules 
 
-* Rules (the license's properties) are stored as a bulleted list within the licenses YAML front matter. A full list of rules can be found in the repository's `_config.yml` file. Each rule has a name e.g., `include-copyright`, a human-readable label, e.g., `Copyright inclusion`, and a description `Include the original copyright with the code`. To add a new rule, simply add it to config.yml and reference it in the appropriate license. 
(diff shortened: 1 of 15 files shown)


afeld:
Until we have licenses on the site that require licenses in binaries, I think this PR could be simplified to a change of the description of `include-copyright`:

> Include a copy of the license and copyright notice with the source code.

marnix:
Thanks for that feedback!  There are already such licenses on the site.  At least the BSD licenses (both 2-clause and 3-clause) state that "Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following disclaimer in the documentation and/or other materials provided with the distribution."

That said, I agree with clarifying the text on (what is currently called) include-copyright.  Do you want a separate pull request for that?

afeld:
Nah, I would just add `include-copyright-binary` to the appropriate licenses in this PR and we'll make sure we have a lawyer verify them.

haacked:
Hmm, I'd rather go with adapting the text to `include-copyright`.

Perhaps:

> Include a copy of the license and copyright notice with the source code. Some licenses also require including a copy of the license with the binaries.

/cc @benbalter, @talniv

ghost:
I'd put the problem this way: "in the course of choosing a license, would I want to know if the license requires explicitly copyright and license for distributions in binary form?". Yes, I think the detail can be relevant for the choice.

benbalter:
Closing as stale, but feel free to continue the discussion or submit a PR if there's still interest in seeing this.

PR #60, opened 2013-07-19
Delete deploy-time-generated app.js file.

marnix:
It seems to me that file app.js needn't be under source control, since it is automatically generated from app.coffee when the site is generated.


benbalter:
@marnix, thanks for going through things so thoroughly and for taking the time to submit the Pull Request. Always a fan of DRYing up code whenever possible.

> app.js needn't be under source control, since it is automatically generated from app.coffee when the site is generated.

As it turns out, Jekyll doesn't actually support pre-processors other than Markdown, at least not yet. (see https://github.com/mojombo/jekyll/issues/819 and its associated pull request https://github.com/mojombo/jekyll/pull/1278).

For now, and because most of the community-contributed edits are going to be to content (not the javascript), we've just been [compiling the Coffeescript locally](https://github.com/github/choosealicense.com/blob/gh-pages/script/cibuild#L5) at build time, and pushing the compiled code to Pages.

Not ideal for a large-scale project, and theoretically, we could use something like [jekyllbot](https://github.com/benbalter/jekyllbot) or [jekyll-hook](https://github.com/developmentseed/jekyll-hook) as a build bot to auto-commit the compiled file, but for our purposes, quick and dirty gets the job done here, after all [it hasn't changed much](https://github.com/github/choosealicense.com/commits/gh-pages/javascripts/app.coffee).

Again, appreciate the issue being raised and for the Pull Request, but for the time being, it looks like `app.js` needs to stay in version control.

marnix:
OK, that's clear, thanks for the detailed explanation!

PR #63, opened 2013-07-22
Using SPDX standard names and IDs for licenses

### LINKED ISSUE #43: Following SPDX standard for license names / IDs
Hi,
In order to avoid ambiguity and facilitate interactions with other projects, it would be nice to use the de facto standard for license names and IDs as defined by the SPDX project. see https://spdx.org/licenses/
Cheers, 
Camille

### LINKED ISSUE #64: Using GPL is too vague, there are GPLs other than the GNU GPL
The site refers to "GPL" but there are various GPLs, not just the GNU GPL.  GNU say to refer to it as "GNU GPL":  http://www.gnu.org/licenses/gpl-faq.html#WhatDoesGPLStandFor

BTW, it may be worth spelling out GPL once as many mistakenly think the G stands for GNU whereas it's General.  This is probably another reason they avoid what they think to be tautological:  GNU GPL.

camillem:
Hi, 
Here are some changes for adhering more closely to the SPDX standard.
What I did :
- Changed the names of the files to "[SPDX ID].txt"
- In the files, changed the value of "title:" to [SPDX name] and the value of "permalink:" to "/licenses/[SPDX ID]/"
- Updated the links to /licenses/GPL-3.0/ in LGPL-3.0.txt "description:" and "how:" sections.
  Points to notice : 
- the "no-licence.txt" file was left as is.
- For the GPL family, I didn't take into account the "only" or "or later" variations.
  Cheers, 
  Camille

Fixes https://github.com/github/choosealicense.com/issues/64. Fixes https://github.com/github/choosealicense.com/issues/43.


tekkub:
Using the full name in the title causes the [list page](http://choosealicense.com/licenses/) to kinda explode:

image

It would probably be worth it to shorten these down so that they fit in a single line if possible (we could drop the "Public License" off some or all if needed).  We could make sure that the description always starts with the full name, so the user sees it and can search for it:

> Version 3.0 of the GNU Lesser General Public License is...

tekkub:
We could also go so far as to include SPDX ID after the full name in the description, sorta like how the Mozilla one is live right now...

> Version 3.0 of the GNU Lesser General Public License (LGPL-3.0) is...

camillem:
Hi @tekkub, 
I shortened the license names. Regarding your suggestion to add the full name followed by the ID, I think it's a good idea, but at the moment, most of the licenses don't have a description.
@Haacked IMO the most important part is the spdx ID in the URL. Would you agree with that?

haacked:
@camillem having it in the URL is fine, but they should all be lowercase.

haacked:
Actually, I take that back. I think the URL should be simple and clean. It's the Title that matters for the correct name. Also, the URLs already out there and we don't want to break any existing URLs.

camillem:
@Haacked  : I understand that cool URIs don't change, but considering the relative youth of the website, I don't think it would break (m)any things to change the current URL in order to follow the /licenses/{SPDX ID} pattern. A quick search with https://www.google.fr/search?q=link:choosealicense.com/licenses gives me only five results (and we wouldn't change this url) and about zaro results for the individual licenses (https://www.google.fr/search?q=link:http://choosealicense.com/licenses/apache/ ). But maybe there are things that I overlooked? 
And more important: at the moment, I don't think it's possible to get the license of a repo via Github's API (please correct me if I'm wrong), but that's something that might happen in the future. Should such evolution happen, it would be really usefull to have SPDX compliant data.
And as the data regarding the licenses seems to be shared between choosealicense.com and github itself (considering the menu at https://github.com/new) , I imagine it would make things easier to be also SPDX consistent at choosealicense.com.

Regarding the names of the licenses, as @tekkub mentioned, some of the full SPDX names would spread on too many lines in the current layout. Hence, I proposed shortened names, that would still relate to the original SPDX name, bring more consistency and some disambiguation elements (like the version number for the Affero GPL).

ghost:
I think that since an API for licenses seems available (which is great!) it's more relevant that URLs would be predictable programatically, and using SPDX IDs would do that.

benbalter:
Breaking the URLs this early on is minor, and don't know that people are deep linking to licenses. Are there any numbers on SPDX usage?

All for standards, but agree with @Haacked that URLs should be lower case. We can expose SPDX-name as a field in the API if that's a problem.

Is there any value to the title being SPDX'd as well?

ghost:
I don't know numbers on SPDX usage, sorry. I can point a few organizations and projects using or referring SPDX, if it helps:
- OSI list (URLs have been updated to use SPDX IDs http://opensource.org/licenses/alphabetical)
- Fossology OS analysis tool. https://lwn.net/Articles/556850/
- Debian copyright and licenses file format (partial compatibility only. http://dep.debian.net/deps/dep5/)
- Others according to SPDX (https://spdx.org/SPDX-Tools, to note BlackDuck, ohloh.net backing company)
- License database. https://licensedb.com

IMHO use of SPDX standard from the start is worth considering for any new licensing tool. (or at least license list IDs)

> All for standards, but agree with @Haacked that URLs should be lower case. We can expose SPDX-name as a field in the API if that's a problem.
> 
> Is there any value to the title being SPDX'd as well?

I suppose you mean SPDX ID, for name, and name, for title? Sorry, I got somewhat confused.

Well, on further thought, IDs as field in the API may suffice, considering that choosealicense.com doesn't intend to serve all open source licenses text anyway and there are other sites doing that or in progress... I don't think the displayed title needs to be the name.

I don't know how far the Github project(s) intend to go, or are open to, on licensing support and interoperability with tools, though. Far-off example: a RDF/RDFa snippet documenting the license in README.md or gh-pages of a project. Rather counter-example: addalicense.com site is duplicating the license texts locally, also referenced with custom, non-standard, filenames.

It may be only me, I feel that while there are relatively old projects realized in this area, it is still an uncharted territory as well, and progressing as we speak.

haacked:
> Breaking the URLs this early on is minor

I disagree. Unless we have a really good reason to break them, I don't think we should. Especially if by now, people are linking to the license pages.

Also, for a content site, what's the benefit of following SPDX. So far, we haven't even discussed this. Just because a standard exists isn't really a good enough reason. Is this THE STANDARD? Are most license foundations etc. on board with it? And how does it apply to a content site vs an API?

It looks like SPDX is focused on data exchange. I hate changing sites geared towards humans to make the site work better for computers. Perhaps it makes more sense to expose a JSON API that follows SPDX.

jlovejoy:
@Haacked - the goal of SPDX is to become the standard for how to report and exchange copyright and licensing information for software, particularly open source software.  I believe that adoption is percolating and we will see a marked uptick soon.  
In regard to the SPDX License List - there is much more adoption of that, than of the full spec currently.  We have identified some of the companies and projects using the SPDX license short identifiers, but I suspect there are many more than we know of.

benbalter:
@camillem any interest in still seeing this forward? Sounds like breaking the URLs are a no go, but titles and/or exposing the metadata via the API makes sense to me.

tekkub:
Pages do support redirects now, so we could do this without breaking existing links.

benbalter:
> Pages do support redirects now, so we could do this without breaking existing links.

Oh man. Great point. :+1: if you're interested in updating the pull request.

benbalter:
Downcased filenames and added redirects over in https://github.com/github/choosealicense.com/pull/209.

camillem:
@benbalter Sorry for the late reply, and yes still very interested indeed. :+1:

PR #69, opened 2013-07-26
Users should be warned of Apache/GPLv2 compatibility concerns

ncoghlan:
Choosing to use the Apache license instead of MIT means that some "GPLv2 only" projects may choose not to use the software. This is rarely something that users of a permissive license actually want.

While the legal situation is murky (the ASF and FSF disagree about whether the licenses are actually incompatible), this is not just a theoretical problem from a social perspective: https://github.com/twitter/bootstrap/issues/2054

### DIFF
--- a/index.html
+++ b/index.html
@@ -29,7 +29,7 @@ description: A site to provide non-judgmental guidance on choosing a license for
       The <a href="licenses/apache/">Apache License</a> is a permissive license similar to the MIT License, but also provides an express grant of patent rights from contributors to users.
     </p>
     <p>
-      As the Apache license is not compatible with the GPLv2 license, some open source projects cannot use software under the Apache license.
+      Some "GPLv2 only" open source projects refuse to include code under the Apache license, as they consider the two licenses to be incompatible.
     </p>
     <p>
       <strong>Apache</strong>, <strong>SVN</strong>, and <strong>NuGet</strong> use the Apache&nbsp;License.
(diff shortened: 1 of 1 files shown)


gstein:
There is no "APL" ... you likely mean Apache License, v2.0. Or "ALv2".

And note that the "theoretical" incompatibility is a statement/belief of the FSF. The Apache Software Foundation does **not** believe they are incompatible.

ncoghlan:
Thanks for pointing that out, Greg (both the incorrect acronym and the fact that the ASF disagrees with the FSF's assessment that the licenses are incompatible).

I've updated the description and the proposed change to the home page accordingly, attempting to highlight that this is more of a social issue, rather than a legal one.

dalb8:
This is particularly important in the case of Android apps, 95% of which contain some Apache licensed code as the support libraries at least are under that licence.

If the copyright header suggestions become more explicit the option of GPLv2+ can be mentioned, allowing for the entire work to be distributed as GPLv3.

haacked:
> I've updated the description and the proposed change to the home page accordingly, attempting to highlight that this is more of a social issue, rather than a legal one.

I'm hesitant to include this on the front page. There are probably many "social issues" we could get into. For example, I can name very large influential tech companies (ok company) that refuse to incorporate GPL because they feel it is "viral". Should we also include a note on that?

Perhaps we need to think about a way to add notes to the actual license page or another more "holistic" approach to bringing out this type of peripheral information.

Issue #68, opened 2013-07-25
Internationalization / i18n / multilingual site


benbalter:
Would love to see about baking in I18N support to choosealicense.com proper. See #67 and #62

We already have the bulk of the strings in a single file (`_config.yml`), so it should just be a matter of abstracting out some of our variables, as far as I can tell.

Would love to take this on starting in August if there's interest. Would be a great exercise to set a standard for Jekyll L10N, and can iterate on some of the lessons learned form https://github.com/CMSgov/healthcare.gov.

@parkr any interest?
@dhcole any insight / words of wisdom?

liks79:
It's really good!
I hope 'choosealicense.com' support I18N.
It is very valuable work for open source beginner!

weakish:
> @parkr any interest?

@parkr _was_ [not interested in supporting i18N in Jekyll core](https://github.com/mojombo/jekyll/issues/389).

parkr:
This looks like a great candidate for use of the [`_data` directory](https://github.com/mojombo/jekyll/pull/1003#issuecomment-21119251) proposed by @benbalter in https://github.com/mojombo/jekyll/pull/1003.

weakish:
Just using subdomain?

A quick and simple (maybe not very elegant approach) is just using subdomains (with separate repositories) .

For example:
- Korean: `http://ko.choosealicense.com`
- Chinese:  `http://zh.choosealicense.com`

Thus, we only need to make sure that:
- Translated sites and main site should link to each other.
- github/{lang}.choosealicense.com has the proper CNAME file (and the related DNS Zone configurations)

Done.

And in future we can switch to a more elegant solution  once Jekyll has better support of I18N.

parkr:
It'd be way easier to handle if everything were in one repo and simply to ship every translation under `/LOCALE/` (e.g. `/en/`, `/ko/`) with pages for each. It's a lot of duplications, but right now it is the best option, given that Jekyll has no means of handling this.

haacked:
@parkr Yeah, I'm kind of leaning towards that approach. @benbalter seem good to you?

liks79:
How about using URI separation? (`http://choosealicense.com/{lang}`)

Fo example :
- Korean : `http://choosealicense.com/ko`
- Chinese: `http://choosealicense.com/zh`

parkr:
@liks79 That's exactly what I was proposing. :)

liks79:
[...]

PR #72, opened 2013-07-26
Chinese translation

weakish:
Chinese localization of choosealicense.com

Web: http://choosealicense.gitcafe.com/


haacked:
This translates the existing site. If I accepted this, it would overwrite the English site. We need a better approach.

haacked:
See #68.

Issue #73, opened 2013-08-05
Artistic License 2.0 does not require Disclose Source


dan2097:
Assuming you do:
 ensure that installation of your Modified Version does not prevent the user
installing or running the Standard Version. In addition, the Modified Version
must bear a name that is different from the name of the Standard Version.

then the Artistic License 2.0 does not require Disclose Source

sources:
http://choosealicense.com/licenses/artistic/
http://osdir.com/ml/licenses.open-source.general/2007-03/msg00055.html

afeld:
Curious that [TLDRlegal](http://www.tldrlegal.com/l/ARTISTIC) also designates the license as Disclose Source.  I'm far from an expert on this topic, but is there some language in the license itself that makes it ambiguous?

haacked:
It's a little confusing. Here's what section 4 a) says:

> Distribution of Modified Versions of the Package as Source
> (4) You may Distribute your Modified Version as Source (either gratis or for a
> Distributor Fee, and with or without a Compiled form of the Modified Version)
> provided that you clearly document how it differs from the Standard Version,
> including, but not limited to, documenting any non-standard features,
> executables, or modules, and provided that you do at least ONE of the following:
> 
> (a) make the Modified Version available to the Copyright Holder of the Standard
> Version, under the Original License, so that the Copyright Holder may include
> your modifications in the Standard Version.

Then 6 comes along and says:

> (6) You may Distribute a Modified Version in Compiled form without the Source,
> provided that you comply with Section 4 with respect to the Source of the
> Modified Version.

I think the fact is there are some cases where you kind of are required to distribute with source and others where you are not.

I haven't taken the time to map this out. @hoolio or @benbalter, I think a lawyerly eye here would really help.

PR #78, opened 2013-08-19
Artistic license is not copyleft

ghost:
- Artistic license is not copyleft, it doesn't require disclose source. Reported in #73
- Also added a proposal for the description, following the information provided by @dan2097

### DIFF
--- a/licenses/artistic.txt
+++ b/licenses/artistic.txt
@@ -5,7 +5,7 @@ permalink: /licenses/artistic/
 
 source: http://opensource.org/licenses/Artistic-2.0
 
-description: A license that&#8217;s heavily favored by the Perl community.
+description: A license that&#8217;s heavily favored by the Perl community. Provided that modified versions do not prevent the user from running the standard version, this license does not require distributions of modified versions to provide the source code.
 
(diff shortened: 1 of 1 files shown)


benbalter:
> A license that&#8217;s heavily favored by the Perl community. Provided that modified versions do not prevent the user from running the standard version, this license does not require distributions of modified versions to provide the source code.

All for this, would love to see if we can't tighten it up a bit. Don't know that we need to explicitly note that it's not copyleft (we don't do that on the other permissive licenses):

> Heavily favored by the Perl community, the Artistic license requires that modified versions of the software do not prevent users from running the standard version.

ghost:
@benbalter Updated as noted. It appears that in some reading, Artistic is taken as if it would be copyleft, but I guess it's only some mistake, and you're right, the description is not the place to clarify that... :smile: The removal of disclose source fixes it.

benbalter:
:+1:, @Haacked ?

benbalter:
also, @engelnyst thanks for taking the time to look into all this. Not super glamorous, but incredibly important and incredibly helpful. Glad to see your contributions. Thank you.

haacked:
thumbs-up-peyton

Thanks! Sorry for the delay!

Issue #66, opened 2013-07-23
GPL and disclose of changes


Lennie:
_The GPL (V2 or V3) is a copyleft license that requires others who modify your code to disclose their changes if they redistribute it in source or binary form_

I'm not sure this explains very clearly what the GPL does.

Basically the GPL makes sure the user of the 'product' will also receive the source and/or modifications. It does not mean the changes need to be made publicly available, only be made available to the user.

So I propose to change the text slightly:
_The GPL (V2 or V3) is a copyleft license that requires others who modify your code to make the code and changes available to whomever they distribute it to_

Or something along those lines.

haacked:
Hmm, mind submitting a pull request with the exact wording you think so we can iterate on it?

PR #79, opened 2013-08-19
Short description of GPL on homepage

ghost:
Small update of the language in GPL short presentation to convey the copyleft requirement some better. I hope it's better... Also it follows @Lennie's suggestion on downstream recipients ("to whomever you distribute to"), although I'm not sure that particular part, while correct, is essential to the license.

The short description should IMHO cover the essentials, one needs to know, it's way too short for specific details... This is probably not yet ideal. Thanks for taking a look though.

### DIFF
--- a/index.html
+++ b/index.html
@@ -36,7 +36,7 @@ description: A site to provide non-judgmental guidance on choosing a license for
     <a href="licenses/gpl-v2"><img height='69' src='images/circular@2x.png' width='72'></a>
     <h3><a href="licenses/gpl-v2">I care about sharing improvements.</a></h3>
     <p>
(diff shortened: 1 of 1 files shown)


benbalter:
All for clarification, but don't know that I'd support anything that lengthens the summary.

> anyone distributing the code, or a derivative work of the code, to make the complete source available under the same terms to whomever they distribute to

Perhaps:

> anyone who distributes your code or a derivative work to make the source available under the same terms

ghost:
Shortened as noted. Thank you, IMO it's better for the first part of GPL short description, it conveys the main point.

benbalter:
I'm :+1:, @Haacked ?

Lennie:
It makes me happy, push it to the website already ;-)

Then you can close this and at least https://github.com/github/choosealicense.com/issues/66

haacked:
Love it! Same length as before, but more clear! 
obama-applause

PR #80, opened 2013-08-19
"What if none of these works for me" already implies the other options.

ghost:
"What if none of these works for me" already implies the other options: if none of these work, the full licenses page will show all the other options available.

Removed the redundant support for non-OSS choice. It was also duplicate, advertised in both "retaining all my rights" form as in "what if I don't want any license" form.
"Retaining all my rights" is an inaccurate and politically charged statement, too, IMO: it says that open source licensors would "lose" something, which is false. They offer rights to everyone else, yes, but the copyright holder doesn't "lose" anything, they continue to have any rights, including to make an additional non-open source distribution if so they want. (with their own code of course).
Please consider this proposal to clean up the homepage, removing this (misleading, IMO), statement.

### DIFF
--- a/index.html
+++ b/index.html
@@ -44,7 +44,5 @@ description: A site to provide non-judgmental guidance on choosing a license for
   </li>
 </ul>
 <div class="see-more callout">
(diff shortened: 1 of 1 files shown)


afeld:
All but the potential redundancy point were fixed by #89.

PR #81, opened 2013-08-20
Compress images

pathawks:
Compress images using [ImageOptim](https://github.com/pornel/ImageOptim) version 1.4.3


haacked:
True story. One of the compressions introduced an ugly pixel artifact. Then I scratched the artifact off my screen. :stuck_out_tongue: 

What were the size savings of the compression?

benbalter:
@cobyism any thoughts?

cobyism:
:+1: Yep, worth doing.

@Haacked This many:

2013-08-20 at 2 52 pm

Issue #48, opened 2013-07-16
Proposed modified workflow: make permissive/copyleft and patents orthogonal


gerv:
There's no point producing a patch for this without some agreement, hence an issue rather than a pull request :-)

The initial 3-way choice, while having the desirability of 1-click licensing choice, means that anyone concerned about patents will end up with Apache 2.0, whereas the GPLv3 also has patent language (and, the FSF would say, GPLv2 has an implied patent licence). In fact, the GPLv3's patent language is significantly stronger than that of the Apache licence, because redistribution means a licence is granted; you don't have to modify, and the patented code doesn't have to be part of your modification. So the patent protection for developers and users is a lot larger.

The best solution would be to say (as we do at Mozilla) that patent language is a necessary requirement for a modern open source licence, and in 2013 no-one should recommend a licence without it. That would simplify the choice down to Apache 2.0 or GPLv3+.

However, if that's a step too far for Github, could we instead have the following workflow?
1. A: "I want it simple and permissive" vs. B: "I want others to share their improvements"
2. A: "I'm not worried about patents" vs. B: "I want to try and avoid patent lawsuits"

A. A. would be MIT (or 2-clause BSD, or ISC, or whatever)
A. B. would be Apache 2.0
B. A. would be GPLv2+
B. B. would be GPLv3+

Gerv

ghost:
FWIW, +1 to both options proposed, any would be better than the current status IMHO. The homepage intends to cover 99% cases (according to goals) and it's implying that if you care about patents, you have AL 2.0.
Entirely agree that permissive/copyleft and patents are orthogonal concerns, and the homepage doesn't fit that.

A 3rd option: (or 2.5 :smile:)
- homepage with two choices, "I want it simple and permissive" vs "I want others to share their improvements".
- first choice paragraph presents MIT or AL 2.0, and states that AL 2.0 is also permissive but it contains patent language.
- second choice paragraph presents GPL v2+ or v3+, and adds explicit patent protection for v3 short presentation.

(equivalent with 2nd, but a smaller step)

haacked:
What if we simply added a clause to the GPL description that it also includes patent language?

gerv:
That would be _slightly_ better, but it would still preserve the idea that you have to choose between "I care about patents" and "I care about sharing improvements", whereas the two concerns are orthogonal.

[...]

PR #83, opened 2013-08-29
GPLv3 short description

ghost:
Fix the second phrase on GPL on the homepage.

This PR proposes to replace GPLv3 phrase with an enumeration of differences to v2. Simple enumeration is more objective, accurate, and informative.
Any one of its items is or is not of potential interest to licensors. Patents helps with #48 but it doesn't solve it entirely; license compatibility is already of interest in filed issues; 'tivoization' is known to be reason to choose and reason to strongly reject GPLv3. For who doesn't know the latter, search engines are lucky with an unmistakable term.
It has the same wordcount with the text it replaces. :smile:


haacked:
We're going to be grouping licenses on the license list page. See #111 for more details. Would love to get your feedback on that. The result of that is we won't have a standalone description for GPL v3 on the page.

PR #88, opened 2013-09-14
Upgrade gh-pages gem

afeld:
Also, specify Ruby version in a .ruby-version file and the Gemfile so that it can be picked up by RVM/rbenv.


benbalter:
Thanks! :sparkling_heart:

PR #89, opened 2013-09-14
Add FAQ re: non-code projects

afeld:
While there still seems to be outstanding disagreement over whether Creative Commons licenses should be included on the site/license-chooser or not (#86), just wanted to get in a stopgap.  Thanks to @tekkub for the suggestion.

screen shot 2013-09-14 at 12 55 13 am

/cc @ashleygwilliams @Haacked @jmburges


tekkub:
Oh, I forgot I needed to open up this PR.  :bear: :heart:

benbalter:
I'm all for the idea, but based on the screenshot above, don't know that it accomplishes our goal of demystifying the process of the user, at least not as much as we can.

Would love @cobyism's thoughts.

That box at the bottom originally just had the top line, and is quickly becoming a :christmas_tree:. Can we reimagine it a bit for its new purpose? Perhaps a much smaller row of three boxes? Not-code, not-licensed, long-tail?

afeld:
Had to rearrange and modify the text a bit, but I think it works.  We might consider icons for the bottom row if we go with this?

screen shot 2013-09-16 at 3 20 15 pm

benbalter:
A few ideas:
1. Perhaps change to "my content isn't code", "none of these work for me", and "I don't want to choose a license" for consistency.
2. Can we make the second set smaller / deemphasize them in some way?
3. Crazy idea, what if we make "what if none of these work for me?" a line by itself in between the top three and bottom three, and then rename the middle option to "I want more choices"?

afeld:
You're killin' me @benbalter :-)

screen shot 2013-09-16 at 3 42 39 pm

benbalter:
That's :sparkles: as far as I'm concerned. You're killing it @afeld. 

@Haacked, @cobyism?

benbalter:
Tightened up the copy a bit for parallelism:

screen shot 2013-09-16 at 3 47 14 pm

haacked:
Ha! I was going to jump in as well but @benbalter read my mind. I like that much better. I'm almost inclined to move all that to another page and change the blue text underneath to "none of the above" and make it link to that page.

But I think I could live with this as well. I need to see it in full context first though.

benbalter:
@haacked I'm with you on that. I'd say let's start with this and then think about moving it to a second page once we've lived with it for a bit?

I'd say :shipit:

haacked:
Consider it :ship:'d

jmburges:
Wooo! This looks great! 

Next thing that needs to happen is integrating the license chooser into the new repo stuff.  It could be as easy to as two drop downs, one for source and one for content.

Also maybe integrating a license drop down for content only into the page generator settings section?

afeld:
@jmburges Feel free to open a new Issue with a proposal/design (or even a PR) for a chooser - curious to hear more about what you're thinking!

jmburges:
will do!

ghost:
This is great! Thanks! :thumbsup:

PR #90, opened 2013-09-14
Include list of resources

### LINKED ISSUE #13: Add links to additional resources
Not sure where it would be appropriate on the site to add these, but I think it would be helpful to add links to the Wikipedia pages for the [Bern Convention](http://en.wikipedia.org/wiki/Bern_Convention) (which makes copyright automatic) and [Work For Hire](http://en.wikipedia.org/wiki/Work_for_hire) (the situation where code you write for your job is actually copyrighted to your employer, and not you).

afeld:
Closes #13.

screen shot 2013-09-14 at 2 13 24 am


benbalter:
:+1: on the idea. I'd move the changes to markdown, and would love to let this sit for a bit in case the community wants to weigh in on what's included.

afeld:
The problem is that I needed to add a class to the `<ul>` to key off of in the CSS, but I don't believe there's a way to do this in markdown.  Would love to hear of a better way!

benbalter:
Makes sense. Append `{: .bullets}` to the line immediately following the last `li` [like this](https://raw.github.com/benbalter/open-sourcing-government/gh-pages/_posts/01-intro/0001-01-01-title.md)

afeld:
@benbalter halp?

screen shot 2013-09-16 at 2 00 11 pm

benbalter:
My bad. Thought we were already using Kramdown (which supports markdown extra). https://github.com/github/choosealicense.com/pull/91's going to be a temporary blocker, but then this should be good to :ship: as is.

haacked:
:thumbsup:

Issue #85, opened 2013-09-08
License text is too light on Linux


anshin:
![2013-09-08-1378651547](https://f.cloud.github.com/assets/1342599/1103241/8fb14974-1895-11e3-9edb-22a7a40fd651.png)

It's not impossible to read, but it's fairly unpleasant. I might suggest embedding a fixed-width font.

benbalter:
@cobyism any thoughts here?

/cc #37, #49

tekkub:
It looks like this is the font stack being used for license text:

```
font-family: Consolas, Monaco, Courier, monospace;
```

Linux has a long history of rendering Courier like :poop:.  On github.com we're using this for the code browser, `<pre>` blocks inside readmes and comments:

```
font-family: Consolas, "Liberation Mono", Courier, monospace;
```

It looks like Liberation Mono has the same metrics as Courier New, and probably has a pretty good install base on linux.  It seems codestyle.org is down (gone?) so I can't confirm install percentages.  It looks like this is a free font too, so we could provide it as a `@font-face`.

haacked:
thumbs up to "Liberation Mono". Anyone want to submit a PR with that?

anshin:
@tekkub codestyle.org is returning a [410](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#410) error. I had no idea it existed, so I'm pretty bummed about that.

I'm in favor of Liberation, and it is open according to the [homepage](https://fedorahosted.org/liberation-fonts/).

haacked:
@anshin would you like to submit a PR with Liberation?

anshin:
I would. I'll have to do some research, and I'll try to submit a PR by tomorrow afternoon.

cobyism:
Yes, there could be improvements to be made here in font rendering, but really I see this issue boiling down to contrast. I still reckon actually giving the license text a separate, lighter background (thus making the text appear darker and more well-defined) like I suggested in #49 is what is needed here:

![](https://f.cloud.github.com/assets/296432/807137/ff6753d4-ee48-11e2-9097-01d0ae43272a.png)

haacked:
@jasonlong any thoughts?

@cobyism I thought skeuomorphism is dead and flat designs are all the rage! :trollface: 

Personally, if it makes it more readable, I'm all for it. If there's a way to do it so @jasonlong is happy with it, that also gets my vote. :smile:

anshin:
[...]

PR #93, opened 2013-09-25
put license text in lighter container for better contrast

jasonlong:
This addresses the color contrast part (but not the font stack part) of #85.

I've put the license text in a lighter container as @cobyism mentioned. I needed to add some padding to this container, so I've widened the overall page container to account for this.

I also unboxed the sidebar "How to apply this license" callout. since it looked a bit cluttered with the new license text box.

screen shot 2013-09-25 at 8 57 22 am

/cc @Haacked @anshin

### DIFF
--- a/_includes/sidebar.html
+++ b/_includes/sidebar.html
@@ -1,7 +1,7 @@
 <div class='sidebar'>
 
   <a href="#" data-clipboard-target="license-text" class="js-clipboard-button button">Copy license text to clipboard</a>
-    <div class='how-to-apply callout'>
+    <div class='how-to-apply'>
       <h5>How to apply this license</h5>
       <p>
         {{ page.how }}
@@ -42,5 +42,5 @@
       </ul>
     </div>
     {% endif %}
-    
+
   </div><!-- /sidebar -->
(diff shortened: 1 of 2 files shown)


haacked:
:thumbsup: from me. @cobyism?

_includes/sidebar.html
```diff
@@ -1,7 +1,7 @@
 <div class='sidebar'>
 
   <a href="#" data-clipboard-target="license-text" class="js-clipboard-button button">Copy license text to clipboard</a>
-    <div class='how-to-apply callout'>
+    <div class='how-to-apply'>
```
cobyism, line 4: @jasonlong Does anything else use the `.callout` class? If not we could :skull: that from the CSS too :grin:

cobyism:
:sparkles: Looks great to me! I say :shipit: and see how it goes :smiley:

haacked:
:thumbsup: Thanks!

_includes/sidebar.html
jasonlong, line 4: Good catch @cobyism. I thought there was another place it was used, but apparently not. I'll send a separate pull for the removal.

PR #94, opened 2013-09-26
Remove unused callout class

jasonlong:
As noticed by @cobyism, the `.callout` class is no longer used and this removes it.

/cc @Haacked


cobyism:
:sparkling_heart: :ship:

PR #95, opened 2013-09-26
Added zlib license

nicklockwood:
The zlib license is another concise, permissive license that is similar to MIT and BSD. I appreciate that the aim of the site is not to catalog all such licenses, and you may not consider the zlib license to be sufficiently differentiated from MIT, in which case feel free to reject this pull request.

The reason I feel that it's worth including is that the short bullet points are more readable than the MIT "wall of text", and it also mentions explicitly that attribution in binary forms is "appreciated, but not required", which is a nice and friendly way to word it, and sets it apart from the stern legalise of other licenses IMHO.

### DIFF
--- a/licenses/zlib.txt
+++ b/licenses/zlib.txt
@@ -1,0 +1,45 @@
+---
+layout: license
+title: ZLib License
+permalink: /licenses/zlib/
+
+featured: false
+
+description:  A permissive license, similar to MIT and BSD, but which states the rules around attribution and misrepresentation more explicitly.
+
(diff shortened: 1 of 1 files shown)


haacked:
Thanks for the PR!

> I appreciate that the aim of the site is not to catalog all such licenses, and you may not consider the zlib license to be sufficiently differentiated from MIT, in which case feel free to reject this pull request.

Thanks for your understanding in advance, but this is exactly how I feel. :smile:

PR #99, opened 2013-10-22
Switch to use the plaintext formatting for Apache License

IQAndreas:
It's very clear that the site had copied over the HTML version of the Apache License.

See [lines 122 to 142](https://github.com/github/choosealicense.com/blob/4e2fd06748ba8cc61c13b604b5bd947242bcbd63/licenses/apache.txt#L122) as an example where the unordered list bullets not copied in along with the rest of the text, throwing off the formatting and bunching together text.

Since the site is displaying the plaintext version of the license, it makes more sense to use the nicely formatted [plaintext version provided by Apache](http://www.apache.org/licenses/LICENSE-2.0.txt).

### DIFF
--- a/licenses/apache.txt
+++ b/licenses/apache.txt
@@ -30,194 +30,205 @@ forbidden:
 
 ---
 
-Apache License
-Version 2.0, January 2004
-http://www.apache.org/licenses/
+                                 Apache License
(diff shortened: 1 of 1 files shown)


haacked:
Thank you!

thumbs-up-chuck-norris

PR #101, opened 2013-10-23
Add license sources where missing

IQAndreas:
Not all of the licenses have a link to their original sources, they have been added where needed. I found the original sources for all of them except `No licence` which obviously didn't need one.

One commit also prevents the `Source` link from appearing if none is defined in the frontmatter (right now it just shows `<a href="">Source</a>`, which isn't bad, it's just not all that helpful).

### DIFF
--- a/_includes/sidebar.html
+++ b/_includes/sidebar.html
@@ -12,9 +12,11 @@
       </p>
       {% endif %}
     </div>
+    {% if page.source %}
     <div class="source">
-      <a href="{{ page.source}}">Source</a>
+      <a href="{{ page.source }}">Source</a>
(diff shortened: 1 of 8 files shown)


_includes/sidebar.html
```diff
@@ -12,9 +12,11 @@
       </p>
       {% endif %}
     </div>
+    {% if page.source %}
     <div class="source">
-      <a href="{{ page.source}}">Source</a>
+      <a href="{{ page.source }}">Source</a>
     </div>
+    {% endif %}
```
afeld, line 19: Good catch!

PR #102, opened 2013-10-23
Fix 'using' display in sidebar

afeld:
Not any licenses currently using it, but there was an unnecessary loop that was causing the list to not appear.


haacked:
:shipit: /cc @benbalter

benbalter:
@afeld if you tested and it works locally, :shipit:

PR #104, opened 2013-10-23
Set up HTML::Proofer

afeld:
FYI this won't pass until #102 is merged, so wait until then.  Did the merge in a throwaway branch just to be sure:

screen shot 2013-10-23 at 12 20 03 am

/cc @gjtorikian

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,0 +1,6 @@
+require "html/proofer"
+
+task :test do
+  sh "bundle exec jekyll build"
+  HTML::Proofer.new("./_site").run
+end
--- a/Rakefile
+++ b/Rakefile
@@ -1,7 +1,7 @@
 require "html/proofer"
 
 task :test do
-  sh "bundle exec jekyll build"
+  sh "bundle exec jekyll build --trace"
   # ignore href="#" for the "Copy to clipboard" button
   HTML::Proofer.new("./_site", :href_ignore => ["#"]).run
 end
--- a/Rakefile
+++ b/Rakefile
@@ -2,5 +2,6 @@ require "html/proofer"
 
 task :test do
   sh "bundle exec jekyll build"
-  HTML::Proofer.new("./_site").run
+  # ignore href="#" for the "Copy to clipboard" button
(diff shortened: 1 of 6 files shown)


haacked:
:shipit: once #102 is merged of course. :smile:

Rakefile
```diff
@@ -0,0 +1,7 @@
+require "html/proofer"
+
+task :test do
+  sh "bundle exec jekyll build --trace"
+  # ignore href="#" for the "Copy to clipboard" button
+  HTML::Proofer.new("./_site", :href_ignore => ["#"]).run
```
benbalter, line 6: Where do we link to `#`?

afeld, line 6: https://github.com/github/choosealicense.com/blob/1b5debd6bf4465ae9b1a807b8c11d78586166043/_includes/sidebar.html#L3

gjtorikian, line 6: I think blanket ignoring `#` is dangerous. There have been times when that's been used as a placeholder (when it shouldn't have). 

Why not use a `<button>` instead of an `<a>`?

afeld, line 6: Another alternative I've heard is injecting that "copy to clipboard" button via JS, so that, if they have JS disabled (for some reason), there isn't a link/button that doesn't do anything.

afeld, line 6: ...which would get around the HTML::Proofer.

benbalter, line 6: https://github.com/gjtorikian/html-proofer/issues/7 may solve this... just need to find time to implement.

gjtorikian, line 6: No excuses now! https://github.com/gjtorikian/html-proofer/pull/12

Really though, `:href_ignore => ["#"]` is dangerous.

afeld, line 6: @gjtorikian You want to do a PR making those changes?

afeld, line 6: (or I can)

benbalter, line 6: @afeld any chance we could add `html-proofer-ignore` to those URLs and drop the `href_ignore`? essentially defeats the entire point of HTML Proofer.

PR #105, opened 2013-10-23
Build gh-pages branch

afeld:
As noted [here](http://stackoverflow.com/a/17119658/358804), the `gh-pages` branch isn't built unless it's explicitly enabled. This also adds the build status badge to the README.


haacked:
Ah, thanks! :shipit:

PR #111, opened 2013-10-28
Group licenses together

haacked:
Proposed fix for #29. This groups licenses into "families". The main link links to the most used or latest version of a license.

Note that as part of this change, I shortened the license names in the list...

license-list

...but kept the full name in the license detail page.

license

Would love to get some feedback on this approach: /cc @benbalter, @aden, anyone else?

Also, @jasonlong, I wonder if there's a better way we could present the licenses that belong to a family? Listing them inline in the text could do for now, but if you have ideas to present it better, I'd love to hear them!


haacked:
This would close #8

talniv:
looks good to me, especially because it is neutral rather than suggestive
of a preference for any family member and it doesn't make the decision
about which license is a member of which family any harder.

I prefer the "latest" rather than the "most used" category for grouping
since I'm not sure how to make popularity scale estimation. Either way, I
think it's important to add an explanation somewhere about why a family
name links to a particular member and not another. Not sure where.

haacked:
> I prefer the "latest" rather than the "most used" category for grouping since I'm not sure how to make popularity scale estimation.

We used GitHub data a while back.

>  Either way, I think it's important to add an explanation somewhere about why a family name links to a particular member and not another.

Why? I think we don't have to explain every possible detail.

talniv:
Two gripes I have with popular on GH scale (neither major):
1. still not wall-to-wall popularity scale
2. new version is supposed to be an improved version and this is a small "competitive block" because it leads you to choose one of the oldies. 

Words of explanation kind of important because I think it's good to be clear about the "service provider's" neutrality in this case. Again, a small thing.

jasonlong:
Yeah, it seems like we're having some growing pains (design-wise) now that we're getting some good content in place. I think a lot of these problems would go away if we were to move the license name and description up and out of the table so that the name wouldn't be restricted in their horizontal space. Then we'd have some nice space for variants.

Something like this?
cal

talniv:
:thought_balloon: maybe everyone can agree to link to the license text just from the **real** license titles and then take away the existing _View full license_ link and the proposed family name link? I like that because this way we are not making any decisions that may indicate to users that the maintainers have a preference and/or pushing users towards past (possibly mistaken) popular license choices and actively restricting (possibly beneficial) legal innovation. No real drama but still...

(I may change my mind if there is ever a point where we have a real basis for having a preference and if we ever think that we want to lead the users to that preference. But I think that if this happens there would be good reasons to be more forthright about it rather than do it implicitly like having such links.)

haacked:
Thanks @jasonlong! I like where that's going.

@talniv that doesn't address the fact that the dots for _permitted_, _prohibited_, and _forbidden_ might only apply to one of the choices.

@jasonlong what if we did something like this (forgive the Haacked-up design). So these links are effectively "tabs". When you click the tab, we update the  _permitted_, _prohibited_, and _forbidden_ dots and the "View full license" link goes to the selected license.

license-detail-gpl-v2
license-detail-gpl-v3

licenses/agpl.txt
```diff
@@ -1,5 +1,7 @@
 ---
 title: Affero GPL
+family-title: GPL
```
benbalter: What was the motivation for `family-title` here? Why not just `family`?

licenses/bsd-3-clause.txt
```diff
@@ -1,6 +1,8 @@
 ---
 layout: license
 title: BSD (3-Clause) License
+family-title: BSD
+hide-from-license-list: true
```
benbalter, line 6: Why not simply `hide: true` or `list: false`? Brevity is slightly more Jekly-y, IMHO.

licenses/gpl-v2.txt
```diff
@@ -1,9 +1,13 @@
 ---
 title: GPL v2
+family-title: GPL
+hide-from-license-list: false
 layout: license
 permalink: /licenses/gpl-v2/
 source: http://www.gnu.org/licenses/gpl-2.0.txt
 
+description: GPL is the most widely used free software license and has a strong copyleft requirement. When distributing derived works, the source code of the work must be made available under the same license. There are multiple variants of the GPL each with different requirements, <a href="gpl-v2/">GPL v2</a>, <a href="gpl-v3/">GPL v3</a>, and <a href="agpl/">Affero GPL</a>.
```
benbalter: `_data` may make this easier in 1.3.0 when it :ship:s, as may moving to `_posts` (with the same permalinks) which would allow us to use categories.

Given the existing constraints, you could loop through all posts and do an `if` on the family... since it's static, we don't really care about making it super efficient.

benbalter:
Pending some of the tweaks above (and passing tests), :+1: to shipping this as is, but would love to go back and improve on the architecture a bit, especially with some of the possibilities 1.3.0 provides.

Specifically, would love to pull the families out as an object themselves (e.g., not hide the family description in one license which would prevent us from having a license-specific description for the "primary" license. Could theoretically do it in `_config.yml` now, but think `_data` is cleaner.

benbalter:
> 1. new version is supposed to be an improved version and this is a small "competitive block" because it leads you to choose one of the oldies.

:+1: to release date as determinative of order / primary link. I don't want Choosealicense.com to be in the business of encouraging holy wars or picking winners or loosers. GPLv2? GPLv3? Not our job to choose. If the people that make GPL, theoretically the foremost GPL experts release an updated version, we need to assume it's because they think it's better. When GPLv4 comes out... I don't want to have to rehash the debate. Newest should be :crown:

benbalter:
> @talniv that doesn't address the fact that the dots for permitted, prohibited, and forbidden might only apply to one of the choices.

I'd say dots should point to latest, with links to other versions. Think of it like a software release. The features page tells you the features of the latest version, but you still have links to prior releases.

licenses/agpl.txt
haacked: Because it's shown instead of `title` if present. But `family` is probably sufficient.

licenses/bsd-3-clause.txt
haacked, line 6: I'm a fan of brevity as well, though I admit that C# (my primary language) is known for `LongAndExplicitIdentifiers`.

In this particular case, I worried that `hide` wouldn't be clear about where it's hiding from. I think `list: false` is probably more clear, but that probably means I need to remember to add that field to _every_ license, not just the ones we're hiding from the list, right? I assume it's a Jekyl convention that absence of a field implies a false value?

licenses/gpl-v2.txt
haacked: What is `_data`.

> Given the existing constraints, you could loop through all posts and do an if on the family... since it's static, we don't really care about making it super efficient.

Yeah, I thought of that. I was being lazy. :) But that seems like a cleaner long term solution.

jasonlong:
@Haacked I like the tabbed idea. Would the description be different for each one or just the table?

I can try to put something together in the next day or two here.

haacked:
@jasonlong Just the table. Though we could experiment with both.

haacked:
@benbalter If @jasonlong has time to try it out, I'd like to experiment with the tab approach. What do you think?

benbalter:
@Haacked I'd actually make them tabs :trollface:, but think the functionality is spot on. :smile:

haacked:
@benbalter heh, we'll let the designer come up with something great. But I think we all agree on the general functionality. :thumbsup: 

@jasonlong no rush, but would love to see what you come up with. :)

jasonlong:
Word. This was actually on my list for today.

jasonlong:
I've got a branch where I've started down a path that looks like this. Before going much further, I wanted to run it by you guys. I think it was sort of hard to scan the colored bullet points horizontally, so I'm trying them vertically here. The "View full license" text would change along with the table and it might be a nice touch to change the `href` on the license family (e.g. "GPL") to point to the selected license variant.

Thoughts?
license-families

haacked:
@jasonlong I love this! I actually think the removal of the tables makes it cleaner. Perhaps make the columns slightly closer together? They seem a bit wide. But other than that, I :heart: this direction.

PR #115, opened 2013-10-30
Add private-use to every license

haacked:
There is no license that prohibits private use.

This closes #25 

Per @benbalter's comment:

> It looks like AGPL doesn't have a blurb, a simple, "This is GPL, but distribution also includes hosted services" or similar should be enough to disambiguate IMHO, but I'd move that to a different issue.

We should do this as a separate PR.


jpmckinney:
Looks good

haacked:
Thanks!

PR #117, opened 2013-10-30
Updates

XhmikosR:


haacked:
Hi there! Thanks for contributing. It's helpful if you provide a brief description of your changes in the pull request. For example, you might describe this as updating libraries to more recent libraries and whitespace cleanup with no real content changes (assuming that's the case).

XhmikosR:
Hi.

I would only include the updates, but the other patches conflict. 

As for the description that's why we have the commits tab, don't we?

I don't like words, I prefer actions. :)

haacked:
This PR appeals to my obsessive compulsive nature!

haacked:
Thank you for this contribution!

thumbs-up-funny-asian-guy

haacked:
> As for the description that's why we have the commits tab, don't we?

Commit tabs are for details. Still nice to have a high-level summary.

PR #118, opened 2013-10-31
Load jquery from Google's CDN.

XhmikosR:
If it's not available then use the local fallback.


benbalter:
What's your reasoning here? The site's already behind a CDN, so theoretically you'd save a DNS lookup as is.

XhmikosR:
Well I wasn't aware that you are already using a CDN.

PR #120, opened 2013-10-31
Javascript cleanup

XhmikosR:
1. Get rid of coffeescript dependency
2. Include `annotations.js` in `app.js`
3. javascripts/app.js: Minor cleanup
4. app.js: Fix JSHint warnings by checking the prototype's properties

Only two JSHint issues left, I'll let someone else deal with those

```
app.js: line 90, col 28, 'Choosealicense' is already defined.
app.js: line 152, col 30, Did you mean to return a conditional instead of an assignment?
```


benbalter:
> Get rid of coffeescript dependency

[GitHub style dictates a preference for CoffeeScript](https://github.com/styleguide/javascript)

XhmikosR:
Which seems useless for such a small file but I don't make the rules. I'll revert that change and update the PR.

XhmikosR:
Nah, screw it. Not worth going through all the trouble to rebase my branch with coffeescript.

PR #121, opened 2013-10-31
Minor improvements

XhmikosR:
1. Use a 16x16 favicon.ico
2. Get rid of unicode symbol from selector's name
3. Clean up .gitignore and .gitattributes
4. Ignore more files in _config.yml


XhmikosR:
I wonder, why is `CNAME` excluded from `_site`?

benbalter:
@XhmikosR great to see you contributing to another project!

The unicode selector is there by design, and we've received no indication of anyone having trouble with it.

Again, as I mentioned in https://github.com/github/government.github.com/pull/149,  pull requests with many unrelated changes across multiple files with the title "minor improvements" do little to facilitate an in-depth discussion.

Closing here, but if you are able to submit these improvements from [individual feature branches](http://www.youtube.com/watch?v=-CpllxXPsvw&list=PLg7s6cbtAD15HBa4C41jhT9YQs5bZumTE) would be **a lot** easier to review.

Thanks yet again for some great proposed improvements, and looking forward to getting these merged soon.

XhmikosR:
1. Makes no sense to use the unicode symbol as a selector at all.
2. Like I have explained to you already, there is a reason for this; I start making changes without knowing the outcome beforehand. So I cannot split them most of the time. All this time you took to write the above comment, you could review the PR and merge it or not. :P

XhmikosR:
...instead of shutting down any work I've done, important or not.

tekkub:
@XhmikosR, you can still break the commits up into logical branches after the fact.  Check out `git rebase`, it's a wonderful command.

The reason @benbalter is asking you to do this is because he wants to accept some of the changes here, but not all of them.  If you make a PR for each of the things you singled out in your summary, it's pretty likely the result would be that (3) and (4) would be merged immediately, there would be discussion on (2) but it would be turned down unless you can justify why a unicode ID is bad, and (1) would likely evolve into adding favicons for all the popular resolutions these days.

When you submit everything in one PR, the chances of the entire pull being turned down due to one small thing are much greater.  It's not a personal attack, don't think your contributions are unwelcome, it's simply that this PR contained commits that were turned down, so the whole PR must be turned down as one can't pick and choose bits of a PR to merge.  It's all-or-nothing.

Break this up into 4 distinct PRs and resubmit, you'll be happy with the results, trust me.

XhmikosR:
I might give it a go tomorrow again if I have some time.

I hate breaking the changes into new branches when I know for sure they will conflict.

PR #123, opened 2013-11-01
Clean up ignore files

XhmikosR:


_config.yml
```diff
@@ -62,8 +62,14 @@ rules:
...
 
 exclude:
+  - app.coffee
+  - CONTRIBUTING.md
   - CNAME
-  - Gemfile*
+  - Gemfile
+  - Gemfile.lock
+  - LICENSE.md
+  - licenses.json
```
tekkub: `licenses.json` is actually intended to be generated and served up as a sort of API.  It shouldn't be excluded.

XhmikosR: Is it used atm? Because I see it isn't.

tekkub: Yup: http://choosealicense.com/licenses.json

XhmikosR: OK. How about CNAME? Why is that excluded?

tekkub: CNAME is used by github's server to set things up, it doesn't need to be served to users.

XhmikosR: OK. So this patch is ok I guess; I updated the PR earlier.

PR #126, opened 2013-11-05
Add CC0 as a license

JoshData:
This commit merely adds a license page for CC0 as /licenses/cc0.

Per the discussions at https://github.com/project-open-data/project-open-data.github.io/pull/135 and https://github.com/cfpb/qu/pull/94, there are no substantive differences between CC0 and Unlicense. Practically speaking they are interchangeable. But CC0 has a strong community and is being strongly pushed within government, so it at least deserves a mention.

PR #127, opened 2013-11-05
Add "I'm a government agency" to the homepage

JoshData:
Now that government agencies are starting to see the light of open government, it would be great to have this site help them.

In the United States, the federal government cannot apply typical open source licenses because it does not hold copyright over its creations. However there are substantive reasons for them to apply CC0. See http://theunitedstates.io/licensing/. And elsewhere in the world where governments can claim copyright, let's encourage them not to use it. Government data is not open (IMO) if any copyright restrictions apply (http://www.opengovdata.org/home/8principles).

The first commit in this pull req adds "I'm a government agency" to the homepage and links to the CC0 license, which I suggested adding a moment ago in #126. Since government agencies are already starting to use CC0, and not the Unlicense, I suggest that the recommendation on this site to be to do what other agencies are already doing, to create the least friction. And it conforms to the guidance at http://theunitedstates.io/licensing/ (which, full disclosure, I am a co-author of).

The second commit recognizes that I broke the beautiful layout of the site in the first commit and offers a way to make it pretty again.

Thanks to The Noun Project for the icon: http://thenounproject.com/noun/federal-government/#icon-No15212


JoshData:
CC @benbalter

benbalter:
It looks like the build's not passing because `/licenses/cc0` doesn't exist.

Beyond that, although the logic is fully sound, as much as it pains me, this may be solving for an edge case, as far as the majority of users are concerned, but am open for thoughts.

haacked:
> this may be solving for an edge case, as far as the majority of users are concerned, but am open for thoughts.

Yeah, not sure it needs to be featured on the home page. What if we had a /govt page and a link to it from the about page? And then we start spreading the word to govt institutions by sending them the link to choosealicense.com/govt

JoshData:
It's an important edge case, and one that I think we'd all like to see more of. But if this pull req is reduced to just a link, then I'd suggest a link to http://theunitedstates.io/licensing/ directly rather than creating another page. We cover the groundwork there. (also cc @konklone)

haacked:
That works for me. We already have a list of other resources in the about page. Would you like to submit a PR with how you would like it presented in the about page? Ideally a link and a brief descriptive blurb after it.

JoshData:
I can. I don't want to (cause I want to see it on the homepage!), but I can....

afeld:
Yeah, it's a tough call, but I (unfortunately) agree that it feels like a niche.  Closing for now.

PR #132, opened 2013-11-16
CC0 fallback license is non sublicensable

ghost:
CC0 fallback license grant is non sublicensable.
(cf. https://creativecommons.org/publicdomain/zero/1.0/legalcode)

> Affirmer hereby grants to each affected person a royalty-free, non transferable, non sublicensable, non exclusive, irrevocable and unconditional license to exercise Affirmer's Copyright and Related Rights in the Work

### DIFF
--- a/licenses/cc0.txt
+++ b/licenses/cc0.txt
@@ -17,11 +17,11 @@ permitted:
   - commercial-use
   - modifications
   - distribution
-  - sublicense
   - private-use
 
 forbidden:
   - no-liability
+  - sublicense
 
 ---
 CC0 1.0 Universal


JoshData:
Thanks..... I think.

I had just copied the list of attributes from the Unlicense. I admit I don't understand (or haven't put in the time to research) what permitting/forbidding sublicensing means here, but I'm surprised if CC0 and Unlicense differ in this regard. (See https://github.com/github/choosealicense.com/blob/gh-pages/licenses/unlicense.txt)

Or, another angle- The language you quoted is for the public license fallback, as you noted. This almost never applies. It would be unlikely to apply in the U.S., for instance, where we all seem to presume a public domain dedication is legally possible. Since it's rare for the paragraph to be relevant, it seems like not a good starting point to say whether CC0 is or isn't sublicensable.

konklone:
@JoshData's points are both good, and the next step should be to also mark the Unlicense as non-sublicensable, since it moves material into the public domain where it is recognized.
1. The CC0's sublicensability clause rarely applies, because it would only apply in jurisdictions where there's no public domain, and the license has to fall back to an extremely permissive (but technically copyright-enforced) license.
2. But separately, **the public domain itself is not sublicensable**. You can't go and copyright a Mark Twain novel that's in the public domain. The Unlicense is less rigorous than the CC0, and doesn't deal with anything but jursidictions where the public domain exists, and so it doesn't need to add an explicit sub-licensability clause, because that's already the situation.

This does make more obvious the differences between a public domain dedication and a license scheme, I get why the CC keeps them as separate as they do.

I've created #146 to mark the Unlicense as non-sublicensable.

/cc @tvol @punkish @bendiken

konklone:
Just filed #163 to remove sub-licensability as an attribute. If it's consensus, that will make this PR moot.

PR #135, opened 2013-11-27
LPPL 1.3c

maieul:
The LaTeX Public Licence is to be used for LaTeX project.

### DIFF
--- a/licenses/lppl-v1.3c.txt
+++ b/licenses/lppl-v1.3c.txt
@@ -1,0 +1,447 @@
+---
+title: The LaTeX project public license
+category: Other
+tab-slug: lppl-v1.3c
+hide-from-license-list: false
+layout: license
+permalink: /licenses/lppl-v1.3c/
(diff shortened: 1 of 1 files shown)


haacked:
Per our [about page](http://choosealicense.com/about/) we're not aiming to be a comprehensive list of all possible licenses. Do you feel people who work on LaTeX projects are unaware of the LaTeX license and don't know how to apply it manually?

maieul:
I don't know. I was making the fork / pull request in reply on a ask to propose this.

If it's only for a helpdesk page, forget. If the choice in menu when creating a new repository is used after (for project page, for example), I think it could be nice.

benbalter:
Closing for now, per @Haacked's comment that the goal of the site isn't to be comprehensive. If the LaTex license becomes adopted by other projects or there is confusion surrounding its use, we can revisit then.

/cc @arfon because Math!

PR #137, opened 2013-11-29
Install all dependencies using Bower

afeld:
Admittedly this adds a lot of junk to the repo, so am also considering using `.gitignore` to exclude the files we're not using.  Let me know if you feel strongly one way or another.


afeld:
Should also note that I'm not installing Modernizr w/ Bower, since we're using a custom build.

haacked:
I have no thoughts on this. What is a bower? :) /cc @benbalter

benbalter:
All for Bower (@haacked Bower is an npm-style package manager for static assets), but as an idiosyncrasy, and to enforce pretty URLs for end-users, I'd prefer to set the path to something like `/vendor` via `.bowerrc`. Take a look at Render for an example.

I'm fine with not .gitignoring the various build components, but feel empowered to go for it if you've got a free cycle.

benbalter:
> Should also note that I'm not installing Modernizr w/ Bower, since we're using a custom build.

Maintainability > being a special snowflake. (assume we just didn't use the components we didn't need?) Feel free to bower if that's the case.

haacked:
Ok, I'll let @benbalter give the squirrel in that case. I'm cool with it. :smile:

css/application.css
```diff
@@ -2,9 +2,6 @@
 layout: nil
 ---
 
-{% include css/normalize.min.css %}
-{% include css/jquery.qtip.min.css %}
```
afeld, line 6: Ran into problems trying to reconcile `_includes` w/ Bower putting everything under one directory, so needed to undo this.  Sorry @XhmikosR!

XhmikosR, line 6: No problem, it's 3 http requests instead of 1 so it's your loss :P

afeld, line 6: Consequence accepted :wink:

XhmikosR:
Are you sure the gain justifies the changes? I mean, it's so much bloat for just a few files...

afeld:
@XhmikosR Well, this was intended to make our lives easier, via
- specification of dependencies all in one place
- easy upgrading of libs

but with the various hassles it is causing, I'm now going to agree that it's not worth it and close this out.  I'll keep the branch around for a bit in case we feel reinvigorated about it.  Thanks for the feedback!

XhmikosR:
No problem, just sharing my thoughts :)

BTW, we should make what I did for the css with the js files too.

DoNotSpamPls:
`keep the branch around for a bit in case we feel reinvigorated about it`

4 (almost 5) years later 🤔

PR #138, opened 2013-11-30
Sort rules by label

### LINKED ISSUE #113: standardize order for required/permitted/restricted buckets
screen shot 2013-10-28 at 5 08 55 pm
screen shot 2013-10-28 at 5 08 14 pm
screen shot 2013-10-28 at 5 08 43 pm

Just examples for cases where we have them in different order. 

Is there a reason for this? I like that these spell out the gist of the license and i think there is real value in the visual clarity that can be found in an ordered list that we are losing when we don't keep the same orde.

afeld:
Spent _waaaay_ too much time on this :smile: :gun: Fixes #113.

/cc @cobyism @Haacked @talniv


_config.yml
```diff
@@ -5,65 +6,65 @@ markdown: kramdown
...
-    library-usage:
-      description: The library may be used within a non-open-source application.
-      label: Library usage
-    rename:
-      description: You must change the name of the software if you modify it.
-      label: Rename
+  - description: Include a copy of the license and copyright notice with the code.
+    label: License and copyright notice
+    tag: include-copyright
+  - description: Indicate significant changes made to the code.
```
haacked, line 12: I don't understand this change. It used to be nested under `document-changes` but now it's not. Was that intended?

afeld, line 12: Yeah, you can only use `sort` on arrays of hashes (or normal arrays) in liquid templates, so I needed to change the structure here to allow that to happen.

haacked:
I don't have time to test it, but Travis CI is happy with it and if you've tested it I'll trust it. :shipit:

PR #142, opened 2013-12-08
Update jquery qtip to v2.2.0.

XhmikosR:


_includes/header.html
```diff
@@ -2,7 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="utf-8">
-    <meta content="IE=edge" http-equiv="X-UA-Compatible">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
```
haacked, line 5: This change seems unnecessary.

css/application.css
```diff
@@ -451,6 +455,8 @@ strong {
   font-size: 90%;
 }
 
+{% include css/jquery.qtip.min.css %}
```
haacked, line 458: Why this change? Just curious.

XhmikosR, line 458: With this, we include jquery.qtip.min.css in application.css when jekyll is run. Thus we have one less http request since we only need to include application.css.

_includes/header.html
XhmikosR, line 5: Yeah, slipped into. I can revert it make it in a separate commit or just go along with it since it's purely cosmetic :P

haacked:
thumbs-up-computer-kid-gif

PR #143, opened 2013-12-08
Add normalize.css v2.1.3 minified with clean-css.

XhmikosR:


haacked:
Could you provide some background for this proposed change?

XhmikosR:
The reason I made this change is because normalize.css is a lot more feature complete than what you had there which I assume was some kind of reset stuff.

BTW, I was thinking maybe it would be simpler to add a `ul { list-style-type: none; }` instead of adding `list-style-type: none;` to most individual `ul`. What do you think of this?

EDIT: I went with this approach since it results in less code.

haacked:
I like less code. I want to make sure our designer sees this before I merge it. /cc @jasonlong

jasonlong:
I'm surprised I didn't use normalize in the first place. This works for me. :thumbsup:

haacked:
Great! @XhmikosR can you merge upstream master into your branch (or rebase it against upstream master if you're so inclined).

XhmikosR:
Done.

PR #146, opened 2013-12-09
Mark Unlicense as non-sublicensable

konklone:
Per the discussion in #132, material released under the Unlicense should be marked as non-sublicensable, since the public domain is non-sublicensable.

### DIFF
--- a/licenses/unlicense.txt
+++ b/licenses/unlicense.txt
@@ -19,11 +19,11 @@ permitted:
   - commercial-use
   - modifications
   - distribution
-  - sublicense
   - private-use
 
 forbidden:
   - no-liability
+  - sublicense
 
 ---
 This is free and unencumbered software released into the public domain.


haacked:
Looks good to me. /cc @bhuga

bhuga:
I'm not sure this is really right. You can copy/paste public domain code into licensed software and release it under a license. You would not really be "sublicensing" in the strictest sense of the term, but I'm not sure saying that "sublicensing is forbidden" does anyone any favors, since it's still "sublicensable" in the practical sense.

The exact use of this word is beyond my legal abilities and I would defer to a lawyer. My instinct would be to leave it as it is until someone could say for sure, but if you guys merge this I would not be freaked out.

konklone:
I agree with that interpretation too! You're free to integrate public domain material into a licensed product. I don't know whether that literally means "sub-licensing".

This is to address the underlying point that there's not a practical difference between the Unlicense and CC0 when it comes to sub-licensability. I wouldn't be freaked out if either this one got merged, or it got closed and #132 got reverted instead.

/cc @tvol @punkish @bendiken

haacked:
Yes, it seems to me that CC0 and Unlicense are the same in this regard, but the legal nuance is way beyond my legal grasp. /cc @talniv for thoughts?

benbalter:
Yes, it's technically not-sublicenseable, in the sense that there's not a license for a subsequent use to be the sublicense of, but a bullet of "can-use-this-thing-in-something-else-under-the-same-or-more-restrictive-terms-or-if-unlicense-or-cc0-possibly-even-less-restrictive-terms" while 100% correct, doesn't meet the project's goals of seeking to simplify all the nuances of applying copyright law to a project.

While a legal scholar would dock points on an exam, I'd say for the purposes of our target audience, this much fuzziness with the terms, which would have an asterisks for "if this were a license, but it's not, so really it's freely usable which is functionally the same", is acceptable to leave as is.

haacked:
> is acceptable to leave as is...

Which one? Right now CC0 and Unlicense are not consistent. Should we make them consistent? And if so, with what?

talniv:
I'm with @benbalter's pragmatism: Putting both CC0 and Unlicense under the sub-licensable bucket is more reflective of what they actually permit users to do rather than the reverse, even though 

> a legal scholar would dock points on an exam

benbalter:
> Which one? Right now CC0 and Unlicense are not consistent. Should we make them consistent? And if so, with what?

Got too excited with my witty bullet point. Yes, it should be consistent, and I'd argue it should fall in favor of sublicensable.

Other idea: change the bullet to "licensable" which would solve both use cases.

konklone:
I agree that if it has to fall one way, they should both be described as sub-licensable. The downside of people possibly being confused when they see the license say "non-sublicensable" somewhere in it after they saw it marked as sublicensable on choosealicense.com seems rare, and not a big deal when it happens. The possibility of people seeing them marked as non-sublicensable concluding that they can't mingle Unlicense or CC0 content in their GPL or proprietary project seems (maybe also rare, but) less rare, and a much bigger deal when it does happen.

bhuga:
Do we have to show yes/no? Can we just not have the field, since it's not applicable?

konklone:
Brilliant.

haacked:
@bhuga send me a PR?

konklone:
Done in #163. Closing this PR, as it's superseded.

PR #148, opened 2013-12-09
CSS cleanup

XhmikosR:
Second attempt... I split the changes so that they can be reviewed more easily.

About the unicode character. I know it doesn't break anything, but on the other hand it doesn't provide any gain.

About the shorthand properties, they exist for this specific reason; to combine all properties without bloating the css.

Anyway, I'm open to feedback, as usual.

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -8,7 +8,7 @@
           This site is licensed under the <a href="http://creativecommons.org/licenses/by/3.0/">
           Creative Commons Attribution 3.0 Unported License</a>.
         </p>
-        <div id="with-♥">
+        <div class="with-love">
           Demystified with &lt;3 by <a href="http://github.com">GitHub, Inc.</a>
         </div>
       </footer>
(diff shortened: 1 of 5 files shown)


haacked:
I like these changes! My only concern is that you got rid of the fallback font families. Is that safe? For example, could that negatively affect mobile devices that might not have support for font embedding? I'm guessing by now, it should be just fine.

If you tested these changes, then I'm all for them. /cc @jasonlong

XhmikosR:
@Haacked: fonts are inherited which means you do not need to specify them in every selector...

XhmikosR:
@Haacked: I went with a slightly better approach; I grouped together common header properties.

haacked:
:cool:

XhmikosR:
Thanks!

Now the only real issue is the Firefox copied to clipboard text not showing up...

XhmikosR:
@Haacked: BTW, do you still still support IE 6 and 7? If not we could/should remove the star hacks used in the css.

haacked:
I'll look at the stats tomorrow. I have no idea what kind of traffic we get from those browsers.

PR #149, opened 2013-12-11
GPLv2 description could do with comma about its variants

randomecho:
>  There are multiple variants of the GPL each with different requirements

Without a comma, this passage has no break where it would be used to add the hint of nuance between GPL versions.

It may read better with a comma before you get to the last half of the sentence.

>  There are multiple variants of the GPL, each with different requirements


haacked:
Thanks! :thumbsup:

PR #151, opened 2013-12-13
Combine javascript files.

XhmikosR:
This results in -4 http requests for the main page.


XhmikosR:
While looking into this, I think none of the javascript files are actually needed for the main page.

This means we can add ZeroClipboard.min.js too in the package and include application.js in the license pages only.

XhmikosR:
So let me know about this and I'll squash the two patches into one before merging.

haacked:
I believe you're right! Just test it well. :smile: But I like this plan.

XhmikosR:
Seems to work fine here unless I'm missing something trivial :)

I'll squash and update the PR in a few minutes, still trying to figure out the copy issue with FF.

XhmikosR:
Done.

XhmikosR:
Here is what we have so far  http://gtmetrix.com/compare/9Pdgy6bx/nRKfm0Vp

Not bad I'd say :)

PR #152, opened 2013-12-13
Work around spaces being added when copying to clipboard.

XhmikosR:
Since the license text is wrapped in `<pre>` with `white-space: pre-wrap` empty lines make the #license-text bigger than it actually is.

Also, get rid of the empty newline at the end of the license text files since jekyll adds trailing spaces when it parses those.

PR Note: this might conflict with the other PRs so just ping me to rebase if/when needed.


haacked:
Thanks! 
thumbs-up-daft-punk

PR #155, opened 2013-12-14
Home sprite

XhmikosR:
Notes:
- I went with the `span` approach for the sprite classes. It seems to work fine in all browsers I could test, but if anyone has a better idea let me know.


jasonlong:
As long as the icons render in the same position as they do now, I'm :thumbsup: with it. I'm not sure I buy that `img` and `js` are more standard than `images` and `javascripts`, but I don't feel that strongly about it.

XhmikosR:
@jasonlong: do you have the original images? Or perhaps you could resize them yourself in a better way than I did. I used Snagit for this.

XhmikosR:
This should be ready to merge.

Should I remove or move the unused images to a new folder or something?

XhmikosR:
Note that I cannot test the 2x images even though I tried to set up the Chome dev tools...

haacked:
Just get rid of the unused images.

XhmikosR:
Done. I also changed the sprite's direction to vertical since it results in smaller image.

haacked:
Nice! :sparkles:

jasonlong:
Sorry, I've been off the grid most of the weekend. Looks like this got taken care of. :metal:

XhmikosR:
No worries I think I got it sorted.  If someone could test the 2x images that would be awesome since I couldn't test those myself.

jasonlong:
The 2x stuff looks :ok_hand: on retina.

XhmikosR:
Perfect, thanks!

PR #156, opened 2013-12-14
Missing question mark on question in intro

randomecho:
Sentence for "list of licenses common for specific communities" reads a little easier as:

> list of licenses common to specific communities

Was also reading it as: 

> list of common licenses for specific communities
- Markdown with uppercase first letter, as a type
- Removed "I counted" as the repo voice otherwise doesn't come from a single person


haacked:
I'll miss the little "(I counted)" joke, but I agree, this is cleaner and better. :thumbsup:

randomecho:
Well, maybe you could put it back with "(we counted)" since corporations are people too.

PR #157, opened 2013-12-15
Remove modernizr.js and use html5shiv.js directly.

XhmikosR:
modernizr.js is basically unused apart from the shiv part so this makes more sense. Plus non IE < 9 browsers don't download the extra file anymore.


haacked:
Did you test this on IE?

XhmikosR:
Yeah in compatibility mode 7-10. Basically the html5 elements are of interest to us and that's what the shim is for.

There's still an issue in IE8 that I have fixed in another patch. The issue is present in master too.

haacked:
Sounds good to me. Thanks!

PR #158, opened 2013-12-15
licenses.html: Add missing page title.

XhmikosR:


XhmikosR:
Fixed another issue I noticed when the variables aren't set.

Maybe we should change the check to check for variations inside the class and include the `.variations-tab` only then too...

licenses.html
```diff
@@ -2,10 +2,9 @@
 layout: default
 permalink: licenses/
 class: license-types
+title: Licenses
 ---
 
-<h1>Licenses</h1>
```
haacked, line 7: So is this `h1` rendered by the layout or something else then?

XhmikosR, line 7: Yeah it's in one of the layouts.

PR #161, opened 2013-12-16
patch-2

XhmikosR:
Tested things in compatibility mode for IE 7-10, they work better than before for the home page, and equally for the rest of the pages.


css/application.css
```diff
@@ -6,7 +6,7 @@ layout: nil
 {% include css/jquery.qtip.min.css %}
 
 body {
-  background: #f5f1ec url(../img/bg.jpg);
+  background-image: url(../img/bg.jpg);
```
randomecho, line 9: Wouldn't you want to have the background colour fallback in cases where the image doesn't load?

XhmikosR, line 9: Sure, but I don't see any reason this can happen.

XhmikosR:
@Haacked: Thanks for the quick merge. Hopefully I didn't break anything with all these changes :)

Here is the report so far, after last days' changes  http://gtmetrix.com/compare/7VtSmLeO/nRKfm0Vp

haacked:
Nice!

At this point, I'm trusting you a lot. I assume you're testing and being careful. I just take a quick look and merge. :thumbsup: I just don't have the time to review and test thoroughly.

bugs-bunny

Phil

PR #163, opened 2013-12-17
Remove sub-licensability public domain licenses/waivers

konklone:
Per discussion in #132 and #146, this removes sublicensability as an attribute for the two public domain dedications contained in the project.


haacked:
gandalf-head-nodding

Thanks!

PR #162, opened 2013-12-17
application.css: Minor cleanup.

XhmikosR:
Move the common `color` property along with the other common properties and remove inherited properties.

PR #164, opened 2013-12-18
Attempt to sort licenses in list

afeld:
Not quite sure why it is different in prod than dev (though I remember coming across issues w/ `sort` inconsistency before), but this should do the trick.  Fixes https://github.com/github/choosealicense.com/pull/162#issuecomment-30769453.  See also: https://github.com/jekyll/jekyll/issues/1802#issuecomment-30823393

/cc @XhmikosR

### DIFF
--- a/licenses.html
+++ b/licenses.html
@@ -7,7 +7,9 @@ title: Licenses
 
 <h2>Featured Licenses</h2>
 
-{% for page in site.pages %}
+{% assign sorted_pages = site.pages | sort: 'path' %}
+
+{% for page in sorted_pages %}
   {% if page.layout == "license" %}
     {% if page.featured %}
       {% include license-overview.html %}
@@ -21,7 +23,7 @@ title: Licenses
   community. For example, Perl developers often choose the Artistic License.
 </p>
 
-{% for page in site.pages %}
+{% for page in sorted_pages %}
   {% if page.layout == "license" %}
     {% if page.featured != true and page.hide-from-license-list != true %}
       {% include license-overview.html %}


XhmikosR:
Nice workaround @afeld! Just out of curiosity, how are currently sorted?

I tried the patch and works the same as before on my dev environment; it sorts the licenses by name which is the path too. So hopefully it will fix the issue for the gh-pages too.

afeld:
> how are currently sorted?

I think it's due to how the files are being pulled in in [this method](https://github.com/jekyll/jekyll/blob/12a55b86ac273f79daf225d60289e3540ad928f0/lib/jekyll/site.rb#L151-L178).  Possibly due to [inconsistencies btw OSes](http://stackoverflow.com/a/5529966/358804)?

XhmikosR:
I thought about something being OS related but all pages in /licenses are lowercase already. Oh well, as long as it works right now.

Feel free to merge this since it seems to work fine here.

afeld:
Works on http://choosealicense.com/licenses/ :smiley:

XhmikosR:
Perfect!

afeld:
P.S. funny that the previous sort order of the "Featured licenses" happened to match the order on the front page.

old home page

benbalter:
So to clarify, this sorts the non-featured license alphabetically?

benbalter:
>  funny that the previous sort order of the "Featured licenses" happened to match the order on the front page.

That was intentional. Is it possible to unsort the featured licenses? Really think the first license on the page should be MIT, not Apache. See generally, http://www.youtube.com/watch?v=-bAAlPXB2-c

XhmikosR:
@benbalter: it might have worked for github pages but on Windows I got the result you see in the screenshots...

You could do

``` diff
 licenses.html | 6 +++---
 1 file changed, 3 insertions(+), 3 deletions(-)

diff --git a/licenses.html b/licenses.html
index 0445a54..6ad7bd0 100644
--- a/licenses.html
+++ b/licenses.html
@@ -7,9 +7,7 @@ title: Licenses

 <h2>Featured Licenses</h2>

-{% assign sorted_pages = site.pages | sort: 'path' %}
-
-{% for page in sorted_pages %}
+{% for page in site.pages %}
   {% if page.layout == "license" %}
     {% if page.featured %}
       {% include license-overview.html %}
@@ -23,6 +21,8 @@ title: Licenses
   community. For example, Perl developers often choose the Artistic License.
 </p>

+{% assign sorted_pages = site.pages | sort: 'path' %}
+
 {% for page in sorted_pages %}
   {% if page.layout == "license" %}
     {% if page.featured != true and page.hide-from-license-list != true %}

```

but the order in the featured licenses won't be consistent again.

afeld:
@benbalter Sorry for not waiting to hear from you!  @XhmikosR's quick-fix would work, but would be better to make the sort order explicit, rather than just rely on the fluke.

XhmikosR:
It should be possible to make the featured licenses' order explicit since they are only 3 at this point but I'm not sure it's worth the hassle...

I have no opinion on the subject as long as the output is the same regardless of the OS :)

XhmikosR:
BTW, can someone give me a tip on how to make github pages actually ignore the CNAME and publish the pages? Apart from me deleting the file. This is a feature I'd like to see because when there's a CNAME file the fork can only be used to develop locally.

afeld:
@XhmikosR The Pages URL will work even if you have a CNAME file (e.g. http://afeld.github.io/hackerhours.org/)... you just need to push a new commit on the `gh-pages` on your fork to trigger the page build.

XhmikosR:
@afeld But when I push to the gh-pages branch then I get the error that CNAME is already taken...

An example I just pushed to test this is http://xhmikosr.github.io/government.github.com/

afeld:
Ah ok.  In that case, I'd do the changes in a feature branch, then merge those into your `gh-pages` branch (which has `CNAME` removed) to test.

XhmikosR:
I tried it and only the html files are loaded. All other resources like CSS and JS return 404. So it still fails it just sort of works because there's index.html in the gh-pages branch. But I don't get any error in the repo settings.

afeld:
Ha, that's because all of those asset URLs are absolute (`/js/...`), but your site is being served from a sub-path (`/government.github.com/`).  Options now are to a) set up your own custom domain so that the site is served from `/`, or to do some Jekyll/Liquid hackery to make all those URLs relative.  Wish I had a better solution for ya :confused:

XhmikosR:
Ah too bad :(  I wish there was an easier way to be able to test things in real life rather than my limited 1280x900 virtual machine...

afeld:
Page order inconsistency addressed here: https://github.com/jekyll/jekyll/pull/1848

afeld:
PR to enable Jekyll to sort by custom properties: https://github.com/jekyll/jekyll/pull/1849

PR #166, opened 2013-12-18
Don't ignore all '#' URLs when running HTML::Proofer

afeld:
Explicitly ignore the clipboard button from proofing, which was why the `:href_ignore` was there in the first place.  Per https://github.com/github/choosealicense.com/pull/104/files#r7311111.

/cc @gjtorikian @benbalter


benbalter:
:+1:

PR #173, opened 2013-12-28
Fix apostrophes

doktorbro:


about.md
```diff
@@ -2,7 +2,7 @@
 title: About
 layout: default
 permalink: /about/
-title: What's this about?
+title: What’s this about?
```
haacked, line 5: Interesting. I'm surprised this wasn't converted using smart quotes. Seems like a bug in Jekyll. What do you think @benbalter?

haacked:
Thanks! We love the attention to detail. :thumbsup:

about.md
doktorbro, line 5: No, it’s not a bug. You must explicitly pass the content to a markdown renderer, if you want rendering outside of `{{content}}`. Use the liquid filter `markdownify` here https://github.com/github/choosealicense.com/blob/gh-pages/_includes/header.html#L6

haacked, line 5: Ah! Mind sending a PR that does that?

PR #174, opened 2013-12-28
Add link to the Open Source Software Collaboration Counseling

doktorbro:
The Open Source Software Collaboration Counseling exists since 2010. The OSSCC tries to serve the same purpose as CAL:

> To simplify license selection, we provide the OSSCC license interoperability guide together with a commented list of preferred OSS licenses.
> 
> The OSSCC also tries to provide a platform to support collaboration between existing currently separate OSS entities and to politically support OSS.
> 
> The OSSCC is currently a group of volunteering people located in Germany and Norway but we are looking for more contributors and staff members worldwide.

### DIFF
--- a/about.md
+++ b/about.md
@@ -19,6 +19,7 @@ This site is not a comprehensive directory of open source licenses.  We think th
 * [Comparison of free and open-source software licenses](http://en.wikipedia.org/wiki/Comparison_of_free_and_open-source_software_licenses) on Wikipedia
 * [License differentiator](http://www.oss-watch.ac.uk/apps/licdiff/) from [OSS Watch](http://www.oss-watch.ac.uk/)
 * [TLDRLegal](http://www.tldrlegal.com/)
+* [The Open Source Software Collaboration Counseling](http://www.osscc.net/)
 {: .bullets}
 
 ### Tools


afeld:
:thumbsup: from me. @Haacked @benbalter?

benbalter:
Would hate to judge a book by its cover, but in pursuit of simplifying resources by pointing visitors at only the best, and not being a comprehensive list of all potential licensing resources available, may make sense to pass on this one. Don't know that we want `about.md` to be a Yahoo-directory style laundry list of links.

Take a look at [osscc.net](http://osscc.net). The first thing you'll notice is a meta-refresh redirect to `/en/index.html`. The second thing you'll notice is a wall of unstyled text (that's not a bug, there's literally no CSS file), and if you open up the source, you'll all tags are uppercase, including `<I>`s.

All in all, with a presentation like that, regardless of content, not sure if we'd be helping users or just confusing them.

doktorbro:
@benbalter Yes, they don’t know how to write HTML and CSS. But the website is fast, structured, and clean. What do you think about the content?

BTW, I could show you more than five technical mistakes on your [own website](http://ben.balter.com/). There is always somebody who thinks different about publishing or is more skilled.

tekkub:
Ignoring presentation (wall of text!), [this table](http://www.osscc.net/en/licenses.html#compatibility) looks to be the meat of that site.  Mmmmm meat.

They actually present some nice info there that we do not, namely "Patent protection" (we only cover "patent grant") and linking/merging against another license (is this "sublicensing"? I'm not exactly clear on what that one is precisely).

benbalter:
> Ignoring presentation (wall of text!), this table looks to be the meat of that site. Mmmmm meat.

Perhaps [wikipedia's chart](http://en.wikipedia.org/wiki/Comparison_of_free_and_open-source_software_licenses) is a better resource then?

tekkub:
Oh, and for what it's worth...

> regardless of content, not sure if we'd be helping users or just confusing them.

I didn't feel like that site helped me make a decision, in fact the impression I got was that they didn't like the GPL family of licenses because they're not compatible with other licenses, they don't like BSD and MIT because they don't mention patents at all, and they greatly prefer Apache and CDDL because of the "patent protection" bits.  It just left me with the feeling that patent protection was really **really** important.

tekkub:
> Perhaps wikipedia's chart is a better resource then?

That chart is nicer, and more direct.  I'd be in favor of us adding "cross-license linking" and "release under another license" to our data points for each license.

doktorbro:
I like the chart too.

I’m happy that the OSSCC brings some aspects of licensing to your attention. The only considerable resource found on the OSSCC is the free book _[Open Source Licensing. Software Freedom and Intellectual Property Law](http://www.rosenlaw.com/oslbook.htm)_ by a real lawyer Lawrence Rosen. But it’s in PDF format, so I think it will confuse people to death.

Issue #97, opened 2013-10-05
Mix of software and content?


afeld:
An example:

I teach a JS class and have all my materials [in a repo](https://github.com/afeld/advanced_js) - both written materials, assignments, etc., and functional code that makes the slide deck itself work.  I want to "properly" open source the whole thing.

What is the best way to handle these cases?  Creative Commons [discourages](http://wiki.creativecommons.org/Frequently_Asked_Questions#Can_I_apply_a_Creative_Commons_license_to_software.3F) use of its licenses for software, but the language of something like the MIT license sound very code-specific.  Do I need to include one license code, one for non-code?  Where is that line drawn?  Are there any licenses that can cover all of it?  While I'm sure it's impossible to give a general, definitive answer, some guidance in the FAQ might be helpful.

/cc @hoolio

haacked:
> Do I need to include one license code, one for non-code?

That's pretty much what I would do. It could be as simple as having an MIT license in the root with a brief blurb that it applies to the source code.

For the content, you'd typically have a footer that mentions it's creative commons. See the footer on haacked.com for an example.

haacked:
> some guidance in the FAQ might be helpful.

Want to submit a PR?

afeld:
The outstanding question is: can one license be used for both?  e.g. with code samples, which do those count as?  For MIT license, for example, do you _need_ to specify a different license for the content?

haacked:
> Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files

Hmm, seems like MIT is primarily concerned with software. I suppose you could consider your training materials to be part of the software. It seems like you're stretching it though. MIT wasn't intended for content, it was intended for software.

> do you need to specify a different license for the content?

You don't _need_ to do anything. As with most things, the real question is what a court would decide. So what is the scenario here that would require a court decision?

I can think of one scenario.

My guess is that the risk isn't to you at all, but to people using your content. The risk is you (the copyright holder) might come along later and say they don't have rights to copy your content, just the code. I don't know what a court would decide in that case. Perhaps they might rule in your favor because existing conventions for content is to apply a copyright notice on the content itself or use a content-friendly license like CC.

So if you wanted to use MIT only, perhaps add a note to the top that says it applies to source code and content. That would mitigate the risk to people using your content. If you later claimed the license didn't apply to the content, they could point to this note you made as evidence of your true intentions.

[...]

Issue #172, opened 2013-12-20
Add GUST Font License


PanderMusubi:
Please add GUST Font License (GFL)

https://www.tug.org/fonts/licenses/gfl.html

doktorbro:
:-1:

The OFL is the only open-source license specifically designed for fonts. #130

PanderMusubi:
OK, GFL will probably not be used by new projects. If OFL is going to be added, this one can be closed.

PR #175, opened 2013-12-29
Add page around licensing content

afeld:
Follow-up to https://github.com/github/choosealicense.com/issues/97.

screen shot 2013-12-29 at 1 46 59 am

screen shot 2013-12-29 at 1 47 16 am

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -5,7 +5,7 @@
         </nav>
         <p>
           If you have questions or issues, it is always best to consult a legal professional.<br>
-          This site is licensed under the <a href="http://creativecommons.org/licenses/by/3.0/">
+          The content of this site is licensed under the <a href="http://creativecommons.org/licenses/by/3.0/">
           Creative Commons Attribution 3.0 Unported License</a>.
         </p>
         <div class="with-love">
--- a/index.html
+++ b/index.html
@@ -56,9 +56,9 @@ description: A site to provide non-judgmental guidance on choosing a license for
 </h2>
 <ul class="triptych see-more clearfix">
   <li class="first-li">
-    <h3>My content isn’t code.</h3>
+    <h3>My project isn’t code.</h3>
     <p>
-      <a href="content">There are licenses for that</a>.
+      <a href="non-code">There are licenses for that</a>.
     </p>
   </li>
   <li>
--- a/index.html
+++ b/index.html
@@ -58,7 +58,7 @@ description: A site to provide non-judgmental guidance on choosing a license for
   <li class="first-li">
     <h3>My content isn’t code.</h3>
     <p>
-      Check out <a href="http://creativecommons.org/choose/">Creative Commons</a>.
+      <a href="content">There are licenses for that</a>.
     </p>
   </li>
   <li>
--- a/non-code.md
+++ b/non-code.md
@@ -1,9 +1,9 @@
 ---
-title: Licensing Content
+title: Non-Code
 layout: default
-permalink: /content/
+permalink: /non-code/
 ---
 
-While the focus of this site is open source software, we recommend checking out [Creative Commons](http://creativecommons.org/choose/) for other types of projects.
+While the focus of this site is open source software, we recommend checking out [Creative Commons](http://creativecommons.org/choose/) for most other types of projects.
 
 If your project contains a mix of software and content, you can include multiple licenses, as long as you are explicit about which license applies to what segments of the project.  See [the license disclaimer for this site](https://github.com/github/choosealicense.com#license) as an example.
--- a/about.md
+++ b/about.md
@@ -1,5 +1,4 @@
 ---
-title: About
 layout: default
 permalink: /about/
 title: What’s this about?
--- a/content.md
+++ b/content.md
@@ -1,0 +1,9 @@
+---
+title: Licensing Content
+layout: default
+permalink: /content/
+---
+
+While the focus of this site is open source software, we recommend checking out [Creative Commons](http://creativecommons.org/choose/) for other types of projects.
+
+If your project contains a mix of software and content, you can include multiple licenses, as long as you are explicit about which license applies to what segments of the project.  See [the license disclaimer for this site](https://github.com/github/choosealicense.com#license) as an example.
--- a/terms-of-service.md
+++ b/terms-of-service.md
@@ -2,7 +2,6 @@
 title: Terms of Service
 layout: default
 permalink: /terms-of-service/
-
 ---
 
 ### 1. Introduction


afeld:
This could also be a good place for mentioning things like font licenses.

/cc @davelab6 https://github.com/github/choosealicense.com/pull/130 @PanderMusubi https://github.com/github/choosealicense.com/issues/172

afeld:
...so maybe it should have a broader name like `non-code`?

davelab6:
I suggest s/My content isn't code/My work isn't code/ 

I agree `s/content/non-code/`

afeld:
How about "My **project** isn't code"?

davelab6:
+1 to project

haacked:
I like "project" or "work" :) :shipit:

davelab6:
I think project is better than work because it covers the minority of people who are hobbyists :)

haacked:
I think of "work" as in my "body of work". But I can live with "project". :smile:

afeld:
Updated!

afeld:
Bump!

/cc @benbalter @kfogel https://github.com/benbalter/benbalter.github.com/pull/98

_includes/footer.html
```diff
@@ -5,7 +5,7 @@
         </nav>
         <p>
           If you have questions or issues, it is always best to consult a legal professional.<br>
-          This site is licensed under the <a href="http://creativecommons.org/licenses/by/3.0/">
+          The content of this site is licensed under the <a href="http://creativecommons.org/licenses/by/3.0/">
```
benbalter, line 8: This site's content?

index.html
```diff
@@ -56,9 +56,9 @@
 </h2>
 <ul class="triptych see-more clearfix">
   <li class="first-li">
-    <h3>My content isn’t code.</h3>
+    <h3>My project isn’t code.</h3>
```
benbalter, line 59: What was wrong with project?

non-code.md
```diff
@@ -0,0 +1,9 @@
+---
+title: Non-Code
+layout: default
+permalink: /non-code/
+---
+
+While the focus of this site is open source software, we recommend checking out [Creative Commons](http://creativecommons.org/choose/) for most other types of projects.
```
benbalter, line 7: This would be our first license "recommendation".

Also, why does the site's focus matter? If this is beyond our focus, why even include it in the site?

non-code.md
```diff
@@ -0,0 +1,9 @@
...
+
+If your project contains a mix of software and content, you can include multiple licenses, as long as you are explicit about which license applies to what segments of the project.  See [the license disclaimer for this site](https://github.com/github/choosealicense.com#license) as an example.
```
benbalter, line 9: I feel like this adds a whole lot of complexity (thus not demystifying) for what is a big edge case for the site's purpose and inches us closer towards being comprehensive, which is not what we want to be.

Perhaps maybe link directly to CC from the homepage like we do now, but start an FAQ, if this is absolutely essential to our purpose?

benbalter:
Can you elaborate a bit on the core use case for this? Most open source licenses can cover code and documentation. I get Jekyll sites, but think that's a minority of our target audience, no?

non-code.md
kfogel, line 7: Ben Balter notifications@github.com writes:

> > +While the focus of this site is open source software, we recommend checking out [Creative Commons](http://creativecommons.org/choose/) for most other types of projects.
> 
> This would be our first license "recommendation".

It's not a license recommendation yet, FWIW.

Creative Commons is an organization, and they offer various licenses.  Of those licenses, the three that are basically compatible with open source are the Attribution (CC-BY) license, the Attribution-ShareAlike (CC-BY-SA), and their Public Domain Dedication (CC0).

All their other licenses have restrictions on derivative works or on commercial use.  Whenever you see "-ND" or "-NC" on a CC licence, it means "no derivatives" or "non-commercial use only", respectively, and is thus incompatible with open source.

This is a distinction CC itself observes, by the way.  They put a seal saying "Approved for Free Cultural Works" on the three Free licenses: CC-BY, CC-BY-SA, and CC0.  You can see an example here:

  http://creativecommons.org/licenses/by-sa/4.0/

For more of an overview, see

  http://creativecommons.org/licenses/  

And for the question of whether to use CC0 when releasing software, see:

  http://opensource.org/faq#cc-zero

There.  I'll bet that was more information than you ever wanted on this topic! :-)

Best,
-K

_includes/footer.html
davelab6, line 8: But it could be the images and other 'non-content' parts of the site's theme...

index.html
davelab6, line 59: +1

davelab6:
I think the rationale is that many 'open source' projects aren't software. And they need a license.

PanderMusubi:
Prjoects such as fonts with an OFL license are an example. In this section a user can choose OFL.

benbalter:
@afeld still interested in seeing this happen?

_includes/footer.html
stuartpb, line 8: I would say that "content" encapsulates images and theming assets, but it doesn't include ops things like the way the the content is provided (the server and what not) which I would normally consider under the umbrella of "this site". My $0.02.

index.html
afeld, line 59: @benbalter With "content", you mean? "Project" feels more all-encompassing, but I don't feel strongly about it.

non-code.md
afeld, line 7: What about "there are other licenses more appropriate for projects that aren't code, e.g. Creative Commons or the Open Font License"?

non-code.md
afeld, line 9: > a big edge case for the site's purpose 

If the site to demystify licenses has a mixed license, it's probably worth mentioning, don't you think? Totally open to any ideas about how to make it less mysterious without going into a lot more detail.

> start an FAQ

What if we moved the "What if none of these work for me?" section of the homepage into an FAQ page, and added this question?

afeld:
@benbalter Sorry for leaving this stale for so long – it was starred in my old inbox. I still think the information would be useful, though it may make more sense on an FAQ page.

> Can you elaborate a bit on the core use case for this?

Fonts are one non-coding use case, though very niche. The bigger thing I had in mind was text files checked into repositories, e.g. Jekyll sites.

index.html
parkr, line 59: "Content" might imply a subset of the work product of a project

afeld, line 59: @parkr So do you favor one over another?

parkr, line 59: I'd probably say `content` and specify that pieces of a project can be individually licensed.

bkeepers:
@benbalter what are your thoughts on this now?

benbalter:
I like the idea of having a page dedicated to non-code content and definitely see the use case, both in the context of GitHub, and the broader world of documentation, Jekyll sites, data, etc. I'm struggling with the implementation in two ways:
1. If the decision tree is basically is this a font? Use OFL, otherwise, use CC, and here are your choices, I'm having trouble seeing what value we can provide to readers beyond https://creativecommons.org/choose/. It may be enough to link out to there directly, but then we're pushing folks to non-open source, non-software licenses. Would adding CC-BY (we already have CC-0 and OFL) be enough?
2. I don't have a good sense of how we could do this without exposing a rabbit hole of complexity around licensing code + content in the same repo, what it means to be open source, etc. 

I'm all for this PR (thanks again @afeld for raising the issue and making the case well), I'm just worried that we'll do more harm than good and invite a :christmas_tree: of edge cases. What if we just pointed visitors to CC-0 and CC-BY? Would that solve your intended use case?

afeld:
> Use OFL, otherwise, use CC

I was keeping this separate from [the OFL discussion](https://github.com/github/choosealicense.com/pull/130), but this page seems like an appropriate place to mention it. I unfortunately can't push to this branch anymore, but I'd add something to the effect of "there are licenses that are applicable to certain types of content (e.g. the Open Font License), but in general, use CC".

> I'm having trouble seeing what value we can provide to readers beyond https://creativecommons.org/choose/

Well, I suppose the main change here is that it explains how to handle code+non-code content in repositories.

> https://creativecommons.org/choose/ ... but then we're pushing folks to non-open source...licenses

Not sure I follow...? All of the the CC licenses are all "open".

> a rabbit hole of complexity around licensing code + content in the same repo

This applies to a fair number of repositories, e.g. where there is a documentation site checked in with the code. I am more knowledgeable than your average :bear: about open source licensing, and I'm _still_ lost about The Right Way to do handle those cases. The sentence in this PR is my best effort at general guidance.

benbalter:
> Not sure I follow...? All of the the CC licenses are all "open".

I was going off of what @kfogel said [above](https://github.com/github/choosealicense.com/pull/175#discussion_r9098828):

> All their other licenses have restrictions on derivative works or on commercial use.  Whenever you see "-ND" or "-NC" on a CC licence, it means "no derivatives" or "non-commercial use only", respectively, and is thus incompatible with open source.

afeld:
Ah. Well, I think that's a tangential discussion, since there's already a link to https://creativecommons.org/choose/ on the homepage. Wasn't my intention to state any opinion on those types of licenses.

kfogel:
@afeld, FWIW, most people don't realize that some of the CC licenses offered at https://creativecommons.org/choose/ are incompatible with open source -- if the referrer doesn't say explicitly "Choose CC-BY, CC-SA, or CC0" and link to those individually, people are quite likely to choose a -ND or -NC license for the docs or other content in an otherwise open source project.  This isn't CC's fault, really.  They've kind of gone out of their way to make the distinction between the Free and non-Free CC licenses, insofar as they can, in other places.  But by the time someone lands at the CC license chooser, if that distinction isn't already on the person's mind, the chooser page isn't in a position to do much about it, since its choice-flow is sort of orthogonal to the free vs non-free question.

afeld:
@kfogel I hear you – mind if we defer that conversation to after/if this PR is merged?

Moved this PR to #307 so that I could fix merge conflicts.

kfogel:
@afeld -- oh, np.  I just wanted to mention the problems with merely pointing to the CC license chooser in this context.  Do things in whatever order is easiset for you, of course.

PR #176, opened 2014-01-03
Make sure titles are in smarty pants

doktorbro:
Ref.: https://github.com/github/choosealicense.com/pull/173#discussion_r8601554


haacked:
Hmm, the build didn't seem to like that. Not sure if it's related to your change or not.

benbalter:
Believe it was due to an unrelated Gem bump. Tested locally and tests pass.

@penibelst thanks for the pull request!

PR #177, opened 2014-01-09
Cleaning up description of GNU GPL in the index.html and changing the note in the gplv2 and gplv3 license pages

joshuagay:
I made some minor changes to how the GNU GPL is described in the index.html page as well a change to the note on the GPLv2 and GPLv3 license pages stating that all source code files should state their copyright and licensing terms, which I think could make sense for all other licensing pages as well.

### DIFF
--- a/index.html
+++ b/index.html
@@ -42,7 +42,11 @@ description: A site to provide non-judgmental guidance on choosing a license for
       <h3>I care about sharing improvements.</h3>
     </a>
     <p>
-      The GPL (<a href="licenses/gpl-v2">V2</a> or <a href="licenses/gpl-v3">V3</a>) is a copyleft license that requires anyone who distributes your code or a derivative work to make the source available under the same terms. V3 is similar to V2, but further restricts use in hardware that forbids software alterations.
+      The GNU GPL (<a href="licenses/gpl-v2">v2</a>
+      or <a href="licenses/gpl-v3">v3</a>) is a copyleft license that
+      protects free code from being turned into proprietary
+      software. Version 3 is international and protects better against
+      DRM, lockdown and patents; it's compatible with Apache 2.0.
     </p>
     <p>
       <strong>Linux</strong>, <strong>Git</strong>, and <strong>WordPress</strong> use the GPL.
--- a/index.html
+++ b/index.html
@@ -42,11 +42,7 @@ description: A site to provide non-judgmental guidance on choosing a license for
       <h3>I care about sharing improvements.</h3>
     </a>
     <p>
-      The GNU GPL (<a href="licenses/gpl-v2">v2</a>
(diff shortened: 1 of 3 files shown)


index.html
```diff
@@ -42,7 +42,7 @@
       <h3>I care about sharing improvements.</h3>
     </a>
     <p>
-      The GPL (<a href="licenses/gpl-v2">V2</a> or <a href="licenses/gpl-v3">V3</a>) is a copyleft license that requires anyone who distributes your code or a derivative work to make the source available under the same terms. V3 is similar to V2, but further restricts use in hardware that forbids software alterations.
+      The GNU GPL (<a href="licenses/gpl-v2">v2</a> or <a href="licenses/gpl-v3">v3</a>) is a copyleft license that protects free code from being turned into proprietary software. Version 3 is international and protects better against DRM, lockdown and patents; it's compatible with Apache 2.0.
```
benbalter, line 45: What are you trying to solve for here?

> a copyleft license that protects free code from being turned into proprietary software

This gives the user less context then they currently get. It's describing the philosophy, not the practical implications for me as a consumer or a publisher. 

> protects better against DRM

DRM is jargon. The target audience should be a non-technical user.

> it's compatible with Apache 2.0.

Why is that vital to include on the home page? What's license compatibility? What's Apache 2.0?

licenses/gpl-v2.txt
```diff
@@ -13,7 +13,7 @@ featured: true
 
 how: Create a text file (typically named LICENSE or LICENSE.txt) in the root of your source code and copy the text of the license into the file.
 
-note: The Free Software Foundation recommends taking the additional step of adding a boilerplate notice to the top of each file. The boilerplate can be found at the end of the license.
+note: As with any free license, you should put near the top of each source file a copyright notice and a license notice for the license you are using.  Without this, the code can get disconnected from its license when someone copies it into another program or another file; then people will not know how they can use that code.
```
benbalter, line 16: If it's applicable to all licenses, why is it metadata on GPL? Why not create a best-practices page, add it to about, or hard code it to all licenses?

index.html
wwahammy, line 45: I think the first sentence is actually clearer in the pull request than the original. The original says it requires you to distribute the code under the same terms but doesn't describe the terms. There's probably a better phrasing than "protects free code from being turned into proprietary software" but at least it says something.

josinalvo:
I found this explanation still a bit technical. Care to look at #260 ?

bkeepers:
> This gives the user less context then they currently get.

I agree with @benbalter here. I'm not sure this helps clarify anything.

@joshuagay I'm going to close this pull requests since it's stale, but I'm happy to reopen it and continue the conversation here if you are interested.

PR #180, opened 2014-01-23
Selectivizr

XhmikosR:


haacked:
What? Seems odd to add a JS library just for this one change.

XhmikosR:
This library was already in before I removed it back then because it didn't work.

But since it now works it makes sense to use it again so that css3 features work in older browsers.

haacked:
Ah, ok.

PR #181, opened 2014-01-23
Markdownify

XhmikosR:
Minor tweaks for better HTML output.

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -3,8 +3,8 @@
   <head>
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
-    <title>{% if page.title %}{{ page.title | markdownify | strip_html }} - {% endif %}{{ site.title}}</title>
-    {% if page.description %}<meta name="description" content="{{ page.description | strip_html }}">{% endif %}
+    <title>{% if page.title %}{{ page.title | markdownify | strip_html | strip_newlines }} - {% endif %}{{ site.title}}</title>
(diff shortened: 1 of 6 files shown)


XhmikosR:
BTW I'm not sure if we want to be consistent with the angle brackets. I mean, only one license is using them (agpl), while 4 others use square brackets.

Let me know what you prefer and I can take care of this.

haacked:
All the GPL licenses have angle brackets in them. what's the result of this change to those licenses?

XhmikosR:
They are shown as they are, properly after they are converted to use the html entities.

XhmikosR:
For example for GPLv3

``` html
You should have received a copy of the GNU General Public License
along with this program.  If not, see &lt;http://www.gnu.org/licenses/&gt;.
```

XhmikosR:
Only downside, if we can call it that, is that due to Jekyll's issue with whitespace, when one copies the license text to clipboard they end up with useless trailing spaces.

haacked:
Ewww. People hate that. There's no hack for that?

XhmikosR:
I'll check again but it's time this is fixed in Jekyll...

XhmikosR:
With `strip_newlines` it beats the point since we end up with a one line text. `newline_to_br` results in useless empty lines. `replace` doesn't accept regex as far as I can tell.

Perhaps it's time we add a plugin to minify the HTML?

XhmikosR:
Potential candidates:

https://gist.github.com/mytharcher/2758691
https://github.com/imathis/jekyll-minify-html

XhmikosR:
@Haacked: Please check #184 for a quick workaround.

PR #182, opened 2014-01-23
Generate sitemap.xml and add robots.txt.

XhmikosR:
Only issue so far is that the generated sitemap.xml includes application.css and javascript.js.

I assume that's because those are parsed by Jekyll since they include other files. So I guess we'll have to exclude those from being included. Any ideas are welcome.

PS. I really wish Jekyll would stop outputting superfluous whitespaces after using its tags like `{% endfor %}`...

### DIFF
--- a/sitemap.xml
+++ b/sitemap.xml
@@ -11,12 +11,14 @@ layout: nil
     <priority>1.0</priority>
   </url>
   {% for page in site.pages %}
+  {% if page.layout != "nil" %}
   <url>
     <loc>{{ site.url }}{{ page.url }}</loc>
     <lastmod>{{ site.time | date_to_xmlschema }}</lastmod>
     <changefreq>weekly</changefreq>
     <priority>0.7</priority>
   </url>
+  {% endif %}
   {% endfor %}
   {% for post in site.posts %}
   <url>
--- a/sitemap.xml
+++ b/sitemap.xml
@@ -1,0 +1,29 @@
+---
+layout: nil
+---
+<?xml version="1.0" encoding="UTF-8"?>
+<?xml-stylesheet type="text/xsl" href="/sitemap.xsl"?>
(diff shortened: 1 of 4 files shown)


haacked:
Maybe exclude by file extension for now?

XhmikosR:
Not sure how to do that but I can try again tomorrow.

XhmikosR:
I fixed the remaining issues, I hope it's ok.

sitemap.xml
```diff
@@ -0,0 +1,31 @@
...
+<?xml-stylesheet type="text/xsl" href="/sitemap.xsl"?>
+<urlset xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.sitemaps.org/schemas/sitemap/0.9 http://www.sitemaps.org/schemas/sitemap/0.9/sitemap.xsd" xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
+  <url>
+    <loc>{{ site.url }}/</loc>
+    <lastmod>{{ site.time | date_to_xmlschema }}</lastmod>
+    <changefreq>daily</changefreq>
+    <priority>1.0</priority>
+  </url>
+  {% for page in site.pages %}
+  {% if page.layout != "nil" %}
```
haacked, line 14: Good idea!

haacked, line 14: If the layout is not set, couldn't this just be:

`{% if page.layout %}`

haacked, line 14: Ah, never mind. I noticed we explicitly set the layouts to nil. I assume you tested this well. :)

XhmikosR, line 14: I tested it and it works fine here. I found this the quickest and simplest solution at the time.

XhmikosR:
Here is how it looks like  https://gist.github.com/XhmikosR/8601648

haacked:
Thanks a lot!

XhmikosR:
If you have set up a webmasters tools account for the page, you should submit the sitemap to google :)

PR #184, opened 2014-01-24
Revert 13c9eb92fb1b01d441e68570979a2da349914d9c.

XhmikosR:
Due to Jekyll outputting useless whitespace, we need to work around this ourselves.

### DIFF
--- a/_layouts/license.html
+++ b/_layouts/license.html
@@ -2,7 +2,7 @@
 
       <div class="clearfix">
         <div class="license-body">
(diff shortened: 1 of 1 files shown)


XhmikosR:
I replace the angle brackets with their HTML entities.

EDIT: Actually don't merge this, the quotes aren't parsed like before now since `markdownify` isn't applied.

XhmikosR:
OK I verified everything. It's like before today's PR's. We should really use `markdownify` though and minify HTML ourselves.

PR #185, opened 2014-01-27
Add meta tag for google webmaster tools

haacked:


XhmikosR:
Can you move this `meta` after `title`? It doesn't change anything but it makes sense to keep them together.

haacked:
I just removed it. I don't think we need it after I've verified the site.

XhmikosR:
Not sure about it, but if you read it somewhere I guess it's ok.

PR #186, opened 2014-01-27
Remove sitemap.xsl

haacked:
This doesn't exist.


XhmikosR:
Bad copy paste... Thanks!

XhmikosR:
And here is why :)   https://github.com/github/government.github.com/blob/gh-pages/sitemap.xml#L5

PR #190, opened 2014-01-31
Version third party libraries with Bower

benbalter:
\+ bump a few while we're in there.


XhmikosR:
I honestly don't see how this is better. You add 16k lines to the repo when it could be 100 tops...

Just require bower install instead.

bower.json
```diff
@@ -0,0 +1,18 @@
...
+  "homepage": "choosealicense.com",
+  "authors": [
+    "GitHub Inc."
+  ],
+  "description": "A site to provide non-judgmental guidance on choosing a license for your open source project",
+  "main": "index.html",
+  "license": "MIT",
+  "dependencies": {
+    "zeroclipboard": "~1.3.2",
+    "jquery": "~2.1.0",
```
XhmikosR, line 13: This simply breaks any older IE support.

benbalter, line 13: What version should it be? Glad to accept a PR to downgrade it.

XhmikosR, line 13: See my replies in government.github,com.

XhmikosR, line 13: Also, you didn't add the CSS files in bower. I still disagree with the bower approach because the repo is bloated but if you are going with this do it for the CSS files too ;)

PR #191, opened 2014-01-31
Downgrade jquery and use the CSS from Bower.

XhmikosR:


XhmikosR:
I will update the PR later or tomorrow to use ZeroClipboard.swf from Bower too otherwise problems will occur.

benbalter:
>  I will update the PR later or tomorrow to use ZeroClipboard.swf from Bower too otherwise problems will occur.

Any thoughts on how to do that while still keeping the bowered components in `_includes` so that they can be concatenated?

XhmikosR:
@benbalter: I guess you took care of this so I'm closing the PR.

PR #193, opened 2014-02-16
Update the year of copyright to 2014

shinnn:
If `2013 - 2014` is preferred, I'll fix this PR.


haacked:
Thanks!

tekkub:
If you went with `{{ site.time | date: "%Y" }}` it will always be up to date as of the last deployment.

tekkub:
Oh derp, this is for the license for this repo, not the value displayed on the site.

afeld:
Ha, this a good meta-issue... when including licenses, I have no idea if I'm supposed to include the year range that the project has been active, or just the current year.

/cc @talniv

PR #197, opened 2014-04-22
Missing "patent grant" feature in AGPL

kappa7194:
The AGPL license is missing the "patent grant" feature in the "permitted" category.

Unless I'm mistaken [section 11](http://www.gnu.org/licenses/agpl-3.0.html#section11) of the license is about granting patents licenses (it's the same text found in [GPLv3's section 11](http://www.gnu.org/licenses/gpl-3.0.html#section11)).

### DIFF
--- a/licenses/agpl.txt
+++ b/licenses/agpl.txt
@@ -20,6 +20,7 @@ permitted:
   - commercial-use
   - modifications
   - distribution
+  - patent-grant
   - private-use
 
 forbidden:


ax3l:
just came across that and it confused me, too. Is @kappa7194 right and that is just an error?
**edit:** furthermore there is no quick explanation of the "affero" feature. This could be in-cooperated in the required or forbidden section.

benbalter:
Great catch @kappa7194! 

@ax3l You're absolutely right. Any interest in taking that up in a separate pull request?

ax3l:
@benbalter sounds good, I would be up for that.

shall I either add it to the tooltip text of "Disclose Source" or add a new field in the "required" column stating something like "SaaS = Distribution" and adding a note in the tooltip again?

I think the latter would be more explicit when clicking through the different GPL versions.

PR #202, opened 2014-06-05
Fixes #201

### LINKED ISSUE #201: agpl.txt should be agpl-v3.txt
Hi.

AGPL and AGPLv3 are two different licenses.
AGPLv3 is GNU license, AGPL (v1) is not.
Your agpl.txt contains the AGPLv3 license, so it should be specified with a '-v3' (like you did with gpl-v2, gpl-v3).

As stated in http://www.gnu.org/licenses/ :

```
GNU Affero General Public License (AGPL)
GNU AGPLv3 (The Affero General Public License version 1 is not a GNU license, but it was designed to serve a purpose much like the GNU AGPL's.)
```

alejandrosantana:
Fix for issue #201
- Changed agpl.txt to agpl-v3.txt to reflect the correct license it contains (GNU AGPL-v3).
- Added agpl-v1.txt and agpl-v2 (non GNU licenses).

Notes: Double check the headers. I tried to mimmick other files, but I may have done something wrong.


benbalter:
What's the value in adding AGPLv1 and AGPLv2? Is there a resource to learn what's different and which to choose? The goal of the site isn't to be comprehensive, and would love to avoid GPL bloat if possible.

alejandrosantana:
I get the point.
Maybe the correct move is to simply rename agpl.txt to agpl-v3.txt  (as AGPLv3 is intended to replace v1 and v2). This way we can achieve to both reflect the correct license contained, and avoid bloating.
I will remove v1 and v2 from the pull request and keep the change on agpl-v3.txt

benbalter:
@alejandrosantana thanks for updating the pull request. What's the value in adding v3 now? Users can assume it's the latest version. This would break existing links to the AGPL license, and then would break links again, if there's another license version bump?

alejandrosantana:
@benbalter I was just trying to be accurate because in other licenses in this project, it is specified (see: gpl-v2.txt, gpl-v3.txt, lgpl-v2.1.txt, lgpl-v3.txt) and as I am used to see 'version 3' explicitly specified as AGPL-3, AGPLv3 in other projects.
But I get your point in keeping the whole picture simple and don't break anything with new versions.

In the spirit of maintaining the list clear, simple and tidy, may I suggest one tiny thing?
Somewhere in the README file, it could be added something like: 
`"If no license version is specified, it should refer to the last available version."`

That would be explicit, clear and keeping the KISS concept, isn't it? What do you think?

benbalter:
As @nuclearsandwich pointed out over in #63, we can now support redirects. May make sense to make `gpl-v3` the true version (as you originally suggested) and redirect from `gpl` which we can move to the latest version?

alejandrosantana:
Might be. That's up the project maintainers and their vision.
That would be more consistent, creating agpl-v3.txt (as there is a gpl-v3.txt, lgpl-v3.txt), or redirenting all (agpl -> agpl-v3, gpl -> gpl-v3,...).
That's just up to you, either one will work for me (though I prefer explicit versioning). Just trying to give my two cents. :)

benbalter:
@alejandrosantana typo in my comment — my apologies. Meant `agpl`. If you add the following I'd be glad to merge, if it's what you had in mind:

```
redirect_from: /licenses/agpl/
```

That will allow linking to `licenses/agpl/` to get the latest version but we can explicitly version `agpl-v3`, etc. to make the version number clear. Make sense?

alejandrosantana:
It looks good to me.
I added the line you mention, but please, review the change (triling slash is OK?), because I am not familiar with this specifics.

benbalter:
Thanks @alejandrosantana for the pull request, and for the tweaks. Really glad to have this addition.

alejandrosantana:
Just my two cents. Thanks to you for your work!

nuclearsandwich:
s/nuclearsandwich/tekkub/ I think you'll find. Thanks for the complement

tekkub:
> s/nuclearsandwich/tekkub/ I think you'll find. Thanks for the complement

I don't think it was meant to be one.  He basically called you a fat smelly tanuki.

nuclearsandwich:
@tekkub I've been called worse.

PR #204, opened 2014-06-16
AGPL: Highlight the Affero Feature

ax3l:
This is a follow up to #197

A new point in the "required" section of the AGPL highlights the difference to the plain GPL.
It states that running a network service of the software triggers the disclose-source requirement, e.g. in SaaS use cases.

This is my personal interpretation of the paragraph 13 AGPL, IANAL.

Comments are welcome!

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -17,8 +17,8 @@ rules:
     label: Disclose Source
     tag: disclose-source
   - description: Users who interact with the software via network are given the right to receive a copy of the corresponding source code.
-    label: SaaS is Distribution
-    tag: saas-disclose
+    label: Network Use is Distribution
+    tag: network-use-disclose
   - description: The library may be used within a non-open-source application.
     label: Library usage
     tag: library-usage
--- a/_config.yml
+++ b/_config.yml
@@ -16,6 +16,9 @@ rules:
   - description: Source code must be made available when distributing the software. In the case of LGPL, the source for the library (and not the entire program) must be made available.
     label: Disclose Source
     tag: disclose-source
+  - description: Users who interact with the software via network are given the right to receive a copy of the corresponding source code.
+    label: SaaS is Distribution
+    tag: saas-disclose
   - description: The library may be used within a non-open-source application.
     label: Library usage
     tag: library-usage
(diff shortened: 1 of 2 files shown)


benbalter:
:+1: from me, but will let it sit for a bit for others to weigh in. Thanks @ax3l!

kappa7194:
Personally I'd change the label: "SaaS is Distribution" seems a permission ("Distribution" is present under "Permitted").

Moreover the license states "interacting with it remotely", so it applies to any kind of interaction, it's not limited to SaaS (see [Why the Affero GPL](http://www.gnu.org/licenses/why-affero-gpl.html)). Think of a modified version of Wordpress for a personal blog or MediaWiki for a project wiki (a localized application) in contrast to Microsoft Office 365 or Adobe Creative Cloud (a true SaaS).

Maybe something like "Disclose source for remote access"?

ax3l:
Thanks @kappa7194 for the feedback.
You are probably right, SaaS is just one aspect that it covers, I was trying to look for something handy to sum it up in little words.

Maybe "Disclose source for remote access" is a bit long.
We could squeeze it a bit to "Public Service is Distribution" ?

**edit:** oh well done, I just reduced it by 3 chars -.o

benbalter:
"Remote use is distribution"?

ax3l:
uh, even better :+1: 
I was focusing more on the provider side but one could also take it that way around.

kappa7194:
> "Remote use is distribution"

+1

parkr:
> "Remote use is distribution"?

This is actually confusing to me because of the ambiguity of "remote". "Network use is distribution" or something which implies distribution of the product in production over the Web is what I'd prefer.

ax3l:
actually the license states "... remotely through a computer network ..." so both are not too bad :baby_chick:

ax3l:
_rebased_ against the recent changes in `gh-pages` to allow for an automated merge again.

benbalter:
Awesome. Thanks for clarifying this @ax3l! :tada: :sparkler:

ax3l:
Thanks to you guys! @kappa7194 @parkr @benbalter :rocket:

PR #207, opened 2014-06-17
Use jsonify to build license json

benbalter:
Rather than manually building the JSON string for `/licenses.json`, rely on Jekyll's `jsonify` filter.

This also expands the information made available via the API to the full license text and all metadata and ensures the JSON if properly escaped, valid JSON in all cases.

Note: until Jekyll 2.0 drops, we still have to build the for loop ourself and manually place the trailing comma, as the `where` filter is not yet implemented to let us filter by layout. :soon:


parkr:
:+1:

Issue #51, opened 2013-07-17
The wording implies that choosing a non-copyleft license means not caring about sharing and improvements


akdjka:
The way you present non-copyleft licenses is very, very incorrect.
I'm yet to see an open source project that wouldn't care about improvements. They all want contributions, but permissive ones choose not to try to force others to contribute (for various reasons).

cobyism:
The three summaries on the landing page are there to highlight three common _deciding factors_ that 99% of people choosing a license are likely to be _primarily_ concerned with—they’re not meant to be exhaustive descriptions.

Can you be more specific about what you believe should be changed? What does the site say that is incorrect?

akdjka:
The exact part?
"I care about sharing improvements."

There is nothing strictly incorrect in it, copyleft is about sharing. But non-copyleft open source is about sharing exactly as well, so caring about sharing should not be factor in choosing a particular open source license. And presenting a non-difference as a primary difference for 99% of people to base their decisions on is very flawed IMHO.

tekkub:
Simply adding the word "enforced" or something similar (and perhaps less aggressive sounding?) might be all that's needed here.  The real deciding factor here seems to not be _encouraging_ sharing to happen, but rather _mandating_ that it does.

Heck, maybe something like...

> I want to ensure others share their improvements.

cobyism:
Would something like:

> I care about _reciprocal_ sharing.

…be a more correct alternative?

akdjka:
Well, IMHO 'reciprocal sharing' is a difference between open source and shared source, not copyleft and non-copyleft, so it's not any better.
'enforced sharing' is correct and better than anything that I've had on my mind.

haacked:
Enforce sounds too strong. Makes me think this guy will show up at your doorstep and make you share.

![](http://1.bp.blogspot.com/_pU9jlgWtrsk/SxW40UwwsfI/AAAAAAAAAY4/H4sMN9VCh7w/s1600/twist.jpg)

How about...

> I want to require that others share improvements

/cc @hoolio, @benbalter

artob:
Perhaps 'mandate' is the word you're looking for, as per @tekkub's comment?

Mind you, there's not much point using the GPL if one isn't going to enforce its mandates. And the end result of that would indeed be that, yes, some men with guns will show up at your door if that's what it takes. So, 'enforce' is not incorrect.

hoolio:
> I want to require that others share improvements

[...]

PR #214, opened 2014-07-29
tweaked summary of the GPL

lsmith77:
see https://twitter.com/mikl/status/494123034836746241

### DIFF
--- a/index.html
+++ b/index.html
@@ -40,7 +40,7 @@ permalink: /
   <li class="copyleft">
     <a href="licenses/gpl-v2">
       <span class="triptych-sprite circular"></span>
-      <h3>I care about sharing improvements.</h3>
+      <h3>I care about legally enforcing sharing improvements.</h3>
     </a>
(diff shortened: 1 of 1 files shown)


benbalter:
@lsmith77 thanks for the pull request (and context)!

See https://github.com/github/choosealicense.com/issues/51 for some background on the wording.

cwebber:
Hm, I certainly appreciate the goals of #51 and the thinking that lead to this pull request, though I really don't like this wording.  "Legally enforcing" is kind of meta, all licenses should be legally enforceable, and enforcing something legally is not necessarily the part people want out of copyleft, that's a vector towards achieving this result. The phrasing "I want to require that others share improvements" is better I believe.

bkeepers:
Thanks for the suggestion. Since this discussion hasn't been active for a while, and there was a lot of discussion around this in #51, I'm going to close this for now. We're happy to continue the conversation though.

PR #218, opened 2014-09-09
Remove untenable assertion from goals

ghost:
Whether in civil rights, human rights, land rights, "intellectual property" rights (as in this case) or any other rights, the act of deciding who ought to be granted which rights in which circumstances is not only inherently political, it is the very essence of politics. Therefore, it is preposterous for choosealicense.com, if it is to be meaningful, to aim to be "Politics Free".

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,7 @@ licenses.
 
 # Immediate Goals
 
+* Non-partisan. Our goal is to help you find a license that meets *your* goals.
 * Well designed, but that goes without saying.
 * The homepage should have just enough to help 99% of folks make a decision.
 * For the 1%, the site will contain a list of licenses common to specific
--- a/README.md
+++ b/README.md
@@ -13,7 +13,6 @@ licenses.
 
 # Immediate Goals
 
-* Politics Free - Let's just not get into it.
 * Well designed, but that goes without saying.
 * The homepage should have just enough to help 99% of folks make a decision.
 * For the 1%, the site will contain a list of licenses common to specific


tekkub:
"Politics" may have been a poor choice of word, but I think that the spirit of that bullet point should be preserved.  I have always interpreted that spirit to be "we don't want to start a :fire: :facepunch: over what license is the best, or the internal conflicts (i.e. politics) of any group of people involved with any of these licenses."

In short, this isn't the place for people to get on a soapbox about any license.

ghost:
Better? :)

haacked:
I like it! @benbalter?

afeld:
:+1:

benbalter:
How is does "Our goal is to help you find a license that meets _your_ goals." interplay with "Not comprehensive.", which is a lower priority. If the license that meets my needs isn't listed, this would suggest that we should now add it.

The original pull request took issue with the word political. The distinction's semantic, but if that's the issue at hand, let's just swap `non-partisan` for `not political`. Why is the 2nd half of the line insufficient as is?

tekkub:
I think this version communicates our intentions much better, I like it.

> If the license that meets my needs isn't listed, this would suggest that we should now add it.

We link to some external sites with more comprehensive lists on our about page.  I think "I don't have a license that covers your needs, but here are some other resources that might" qualifies as "helping".

benbalter:
I'm not as strongly opinionated on this one, so @tekkub and @Haacked I'll defer to your judgement.

haacked:
selfie-0

ghost:
Thanks! :+1: all round :)

PR #224, opened 2014-10-01
Update gpl-3.0.txt

nomce:
Changed the sublicense part from forbidden to permitted, according to Wikipedia.

### DIFF
--- a/_licenses/gpl-3.0.txt
+++ b/_licenses/gpl-3.0.txt
@@ -21,6 +21,7 @@ required:
 
 permitted:
   - commercial-use
+  - sublicense
   - modifications
   - distribution
   - patent-grant
@@ -28,7 +29,6 @@ permitted:
 
 forbidden:
   - no-liability
-  - no-sublicense
 
 ---
                     GNU GENERAL PUBLIC LICENSE


benbalter:
> according to Wikipedia.

Link?

nomce:
http://en.wikipedia.org/wiki/Comparison_of_free_and_open-source_software_licenses

Sorry, I forgot the link. The third table (section "Article comparison"), column Subliscensing: GPL v2 - no, GPL v3 - yes.

nomce:
So, will you pull this request, please?

benbalter:
Just because something is on wikipedia, that doesn't make it true (especially when a legal interpretation without a cite).

Take a look at [line 209](https://github.com/github/choosealicense.com/blob/gh-pages/_licenses/gpl-3.0.txt#L209) of the license:

> Sublicensing is not allowed; section 10 makes it unnecessary.

benbalter:
Closing as § 10 states:

> 1. Automatic Licensing of Downstream Recipients.
> 
> Each time you convey a covered work, the recipient automatically
> receives a license from the original licensors, to run, modify and
> propagate that work, subject to this License.  You are not responsible
> for enforcing compliance by third parties with this License.

Please reopen if I'm missing something or there's reason to believe otherwise.

benbalter:
Actually I like @bkeepers solution over in https://github.com/github/choosealicense.com/pull/303.

PR #231, opened 2014-12-18
Mention public domain in "no license"

mindspillage:
Some people who choose this option may really intend PD; mention that option and what to do


bhuga:
Long parentheses can be difficult to follow. I would start a new paragraph with "This might not be what you intend:"

> Disallowing use of your code might not be what you intend by "no license". If your goal is to release your code with no restrictions, you must explicitly opt-out of copyright protections. An explicit public domain dedication such as the [Unlicense](/licenses/unlicense) or [CC0](/licenses/cc0) might be what you were looking for.

PR #232, opened 2014-12-18
Use W3C compliant `textContent` instead of `innerText

### LINKED ISSUE #46: "Copy license text to clipboard" does not work on firefox
When clickin the "Copy license text to clipboard", the license is not copied.
Running Firefox 22 on ArchLinux.

I'm constantly prompted to install a propietary plugin (#45), so that may be related.

### LINKED ISSUE #119: Copy to clipboard doesn't show "Copied" with Firefox
Chrome works fine, only FF 25+ I tried has this problem.

**Update**

To clarify, @XhmikosR told me that the license text is actually copied just fine, but the button text is not updated to reflect that.

gjtorikian:
Closes https://github.com/github/choosealicense.com/issues/119, closes https://github.com/github/choosealicense.com/issues/46.

Firefox doesn't have `innerText`, but it does have `textContent`. This change allows the button to say "Copied!" when click on, and, on mouseleave, it goes back to the default text. This was also tested and working in Chrome.

ff

It's a minimal change, but still: /cc @github/choosealicense-com

### DIFF
--- a/assets/js/app.coffee
+++ b/assets/js/app.coffee
@@ -90,11 +90,11 @@ class Choosealicense
 
   # Callback to restore the clipboard button's original text
   clipboardMouseout: (client, args) ->
-    @innerText = $(this).data("clipboard-prompt")
+    @textContent = $(this).data("clipboard-prompt")
 
   # Post-copy user feedback callback
   clipboardComplete: (client, args) ->
-    @innerText = "Copied!"
+    @textContent = "Copied!"
 
   # Initializes pill navigation for license variations
   initLicenseVariationNav: ->


benbalter:
:+1: nice catch.

gjtorikian:
For more reading: https://stackoverflow.com/questions/22990812/why-is-innertext-not-working-in-firefox

> innerText is the "old Internet Explorer" way of doing it.

Which made me laugh. :grinning:

XhmikosR:
@gjtorikian: How about just checking what is present and use that instead?
Like

``` js
var textMethod = document.body.textContent ? 'textContent' : 'innerText';
```

PR #233, opened 2014-12-20
Remove `layout: nil`.

XhmikosR:


parkr:
:+1: You can also use `layout: null` if there are layout defaults in the config yaml.

XhmikosR:
I just went with what is already being used :)

benbalter:
Thanks @XhmikosR!

PR #234, opened 2014-12-20
Use `innerText` if `textContent` isn't available.

XhmikosR:
This should cover all cases.

### DIFF
--- a/assets/js/app.coffee
+++ b/assets/js/app.coffee
@@ -88,16 +88,13 @@ class Choosealicense
     else
       @initAlternativeClipboard()
 
-  # Check which text selection method is available
-  textMethod = if document.body.textContent then "textContent" else "innerText"
-
   # Callback to restore the clipboard button's original text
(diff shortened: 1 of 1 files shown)


XhmikosR:
Alternatively, we could use jQuery's `html()` which should also work in all browsers.

I have pushed both solutions; I'll wait for feedback in order to rebase and remove one of the two.

Note that I have tested this with IE 9+, Chrome and Firefox work fine. IE 8 doesn't work, but doesn't work in current gh-pages either.

/CC @afeld @gjtorikian

XhmikosR:
So, any thoughts?

benbalter:
@XhmikosR is there an issue that this fixes? If not, could you provide some more context? Is there an advantage to one approach over the other?

XhmikosR:
Have a look to `innerText and`textContent` and you will see that there are cases one is missing. The jQuery solution should work everywhere.

Note, that I haven't looked at this patch for months, so no idea what happens now.

PR #236, opened 2014-12-20
Update frontend assets.

XhmikosR:


XhmikosR:
This reminds me, maybe we should switch to html5printshiv instead of the plain one.

XhmikosR:
Bump

XhmikosR:
I guess I should give up.

benbalter:
@XhmikosR it's hard to tell what's going on from the diff. Can you provide a human-readable description? Were there any substantive changes? Also, it looks like at least one fine got un-minified, and some of the paths changed?

XhmikosR:
@benbalter: quite the opposite; everything used is minified. As for the diff, there's no way to do that. Check out the branch locally if you want.

strugee:
> As for the diff, there's no way to do that.

Yes there is. I believe what @benbalter was looking for was something like:

> - Updated JQuery from x.yz to a.bc
> - Added Foolib for some purpose

etc.

PR #237, opened 2014-12-30
Clean up mention of public domain on no license page

bhuga:
Continues the very good idea to add a mention of public domain to the 'no license' page at https://github.com/github/choosealicense.com/pull/231. cc @benbalter for a sanity check.

cc @mindspillage


parkr:
Nice. :+1:

benbalter:
> If you want to share your work with others, you must opt into it.... If you want to share your work with others, please consider choosing an open source license.

Really like this. One suggestion, If I land on the no-license page (e.g., through Google), I believe we'd want to push the user _to_ license their code (that's the whole point of the site), with public domain being a strong alternative. Perhaps add a sentence before public domain, along the lines of if you want to share your work with the least restrictions, consider [using the MIT license](#) (or even linking to the home page).

parkr:
Isn't public domain the least restrictive "licence"?

benbalter:
@parkr Semantically, it's not a license. It's the rejection of copyright, and thus can't be licensed. Practically, "least restrictive" may not be the best default to point uneducated users to, as that restriction that MIT imposes, for example, is the restriction that you can't sue the author if the software has a bug.

bhuga:
> If I land on the no-license page (e.g., through Google), I believe we'd want to push the user to license their code (that's the whole point of the site), with public domain being a strong alternative.

I was thinking that our front page is a pretty strong nudge to license, but I wasn't thinking about google. I'll cram a "permissive license" snippet in there and link it to the front page's wizard-style walkthrough.

PR #240, opened 2015-01-07
Update README.md with new location of rules

stuartpb:
Rules are in `_data/rules.yml` now, not `_config.yml`.


benbalter:
Thanks @stuartpb!

PR #244, opened 2015-01-10
Use curly apostrophes

doktorbro:
Follow up for #173


benbalter:
Nice catch. Just making a note here, these dumb quotes were in an HTML file, not a markdown file, and thus not processed through Jekyll. For most site content, we should try to push the smart/dumb logic to Jekyll's renderer whenever possible. Thanks again @penibelst!

PR #246, opened 2015-01-10
A null value is used to indicate the lack of a value

doktorbro:
This fixes the build warning:

```
Build Warning: Layout 'nil' requested in robots.txt does not exist.
```


doktorbro:
Dupe of #233.

PR #247, opened 2015-01-17
Applying the LGPL v3 correctly

real-or-random:
The LGPL v3 does not include the boilerplate notice. One should use a modified variant of the boilerplate notice at the end of GPL v3, see https://www.gnu.org/licenses/gpl-howto.en.html:

> When using the Lesser GPL, insert the word “Lesser” before “General” in all three places. When using the GNU AGPL, insert the word “Affero” before “General” in all three places.


benbalter:
I'm :+1:. Anyone else have any strong thoughts?

bkeepers:
I'm :+1:

PR #210, opened 2014-06-24
Added the WTFPL licence

pingiun:
I think the WTFPL is a good addition to the licenses because it's the most open license out there. It's a good option if you want to go full open source.

PR #250, opened 2015-02-07
Added WTFPL

PoroCYon:


randomecho:
#3 #15 and #210 might be relevant here as to why WTFPL might/will not be included, if the stance is still the same.

bkeepers:
Thanks for the pull request. Our stance is still the same as explained in https://github.com/github/choosealicense.com/pull/210.

PR #251, opened 2015-02-26
Add BSD-3 and AGPL descriptions

benbalter:
For some reason the BSD-3 and AGPL were missing descriptions. This adds simple descriptions for each, based on the primary license from that category + a short description of what distinguishes them.

### DIFF
--- a/_licenses/agpl-3.0.txt
+++ b/_licenses/agpl-3.0.txt
@@ -8,7 +8,7 @@ permalink: /licenses/agpl-3.0/
 source: http://www.gnu.org/licenses/agpl-3.0.txt
 redirect_from: /licenses/agpl/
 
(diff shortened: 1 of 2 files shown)


benbalter:
> I'm not sure all would agree that the differences between the BSD family and MIT are minute. Is this sentence essential?

@talniv That was pulled from the existing BSD description. Can you elaborate on why you think the differences aren't minute? In practice, at least in terms of what choosealicense.com tracks, they seem very simliar, no?

benbalter:
> How about starting with "the GPL family of licenses..."?... I think it would be good to say "The AGPL family of licenses..." instead of just the AGPL in the third sentence 

@talniv done in 79368c3.

benbalter:
Going to merge this as is, as it's blocking a downstream project, but more than happy to go back and continue to improve these are any other license descriptions going forward.

PR #252, opened 2015-03-07
Add additional tests

benbalter:
This instructs HTML Proofer to validate the generated HTML and adds some additional tests to validate the license metadata. Specifically:
- Each license has a title
- Each license has a description
- Each license has the required, permitted, and forbidden groups
- Each rule within those groups is a valid rule listed in `_data/rules.yml`

~~It's properly failing now, blocked by #251.~~


benbalter:
Via 530e15d771595ad821c7c4637b065fb11f58f979, it now validates that all filenames (ids) are valid SPDX IDs, and that all titles are the proper SPDX title.

PR #253, opened 2015-03-07
Add license nicknames for license list

benbalter:
This fixes a problem introduced in https://github.com/github/choosealicense.com/pull/252 whereby the official license names on the license list become a bit too much for the design.

This introduces the concept of the license "nickname" (the common name, what was listed before for many licenses) that we can use on the license list to keep things simple for those browsing, but can still use the official SPDX compliant name where appropriate.
### Before

screen shot 2015-03-07 at 2 16 02 pm
### After

screen shot 2015-03-07 at 2 16 07 pm

### DIFF
--- a/_includes/license-overview.html
+++ b/_includes/license-overview.html
@@ -16,7 +16,13 @@
       <ul class="nav-pills js-nav-pills">
         {% for variation in site.licenses %}
           {% if variation.category and variation.category == license.category %}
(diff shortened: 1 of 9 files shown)


benbalter:
Merging because the formatting's a bit off right now on the live site, but more than glad to discuss further or revert if necessary.

parkr:
:+1:

bkeepers:
:+1:

PR #254, opened 2015-03-08
Use Sass

benbalter:
This pull request moves things over to Jekyll's native Sass functionality to allow us to concatenate and minify the stylesheets. Specifically:
- Symlink from `_sass` to vendored versions of normalize and jquery qtip
- Rename `application.css` to `application.scss`
- `@import` the vendored stylesheets (and remove the direct calls from the header)
- Set type to compressed
- Make H1's 35px instead of 40px to accommodate larger titles

Tested on a fork (to confirm the symlink worked) and everything appeared as expected. There should be no visual changes here.


bkeepers:
:+1:

> Symlink from `_sass` to vendored versions of normalize and jquery qtip

I haven't seen that before in a jekyll site. Is that a common technique?

benbalter:
> I haven't seen that before in a jekyll site. Is that a common technique?

No, but it worked. It's the only way to concat vendored CSS that I could think of (when we moved to bower, we unconcat'd what used to be includes).

parkr:
> >  I haven't seen that before in a jekyll site. Is that a common technique?
> 
> No, but it worked. It's the only way to concat vendored CSS that I could think of (when we moved to bower, we unconcat'd what used to be includes).

@gjtorikian wrote `include_relative` to concat, but you get the benefit of minification and the sass cache with this method.

gjtorikian:
Well, we only use `include_relative` to concat JS; for example, in an _application.coffee_ file:

```
---
---

{% include_relative coffeescripts/platform_switcher.coffee %}
{% include_relative coffeescripts/image_handler.coffee %}
{% include_relative coffeescripts/helper.coffee %}
{% include_relative coffeescripts/jquery.lunr.search.coffee %}
{% include_relative coffeescripts/dropdown.coffee %}
```

We then minify the final file with [jekyll-jsminify](https://github.com/gjtorikian/jekyll-jsminify).

We have no vendored CSS in the Help site (which is extremely :disappointed:). The symlink trick here is neat.

_config.yml
```diff
@@ -34,7 +34,11 @@ exclude:
...
   - assets/vendor/zeroclipboard/test
-  
+
 gems:
   - jekyll-sitemap
   - jekyll-redirect-from
+
+sass:
+    sass_dir: _sass
+    style: :compressed
```
gjtorikian, line 44: Might as well have `script/server` refer to a dev config file that reads:

``` yaml
sass:
  http_path: ''
  style: :expanded
```

Keep the SASS expanded in dev, compressed in prod.

PR #256, opened 2015-03-08
Fix GPL v3.0 nickname

### LINKED ISSUE #255: GNU GPLv3 and GNU GPLv2 under the same name
I was looking to my favourite version of the GNU GPL and I found myself a bit confused when saw that. :sweat: 

I seems that an error has been introduced in the titles. Hope that my issue helped.

capture

bkeepers:
Closes https://github.com/github/choosealicense.com/issues/255

/cc https://github.com/github/choosealicense.com/pull/253 @benbalter @MagicKiwi

### DIFF
--- a/_licenses/gpl-3.0.txt
+++ b/_licenses/gpl-3.0.txt
@@ -1,6 +1,6 @@
 ---
 title: GNU General Public License v3.0
-nickname: GNU GPL v2.0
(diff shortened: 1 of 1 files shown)


benbalter:
That was my mistake. Great catch. Thanks @bkeepers. :+1:

PR #260, opened 2015-03-27
making GPL text easier to understand

josinalvo:
I found the GPL text a bit dry. Does one need to know what copyleft is to choose it ?

Tryed to right a neutral text (not pro, not against) that makes it clearer

### DIFF
--- a/index.html
+++ b/index.html
@@ -43,7 +43,7 @@ permalink: /
       <h3>I care about sharing improvements.</h3>
     </a>
     <p>
-      The GPL (<a href="licenses/gpl-2.0/">V2</a> or <a href="licenses/gpl-3.0/">V3</a>) is a copyleft license that requires anyone who distributes your code or a derivative work to make the source available under the same terms. V3 is similar to V2, but further restricts use in hardware that forbids software alterations.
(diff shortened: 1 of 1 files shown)


benbalter:
See https://github.com/github/choosealicense.com/issues/51 and https://github.com/github/choosealicense.com/pull/177 for some additional context. It seems to many people, the GPL means many things, and it'd be great to harmonize the parallel efforts.

josinalvo:
I want people to "pay it forward"

The GPL (<a href="licenses/gpl-2.0/">V2</a> or <a href="licenses/gpl-3.0/">V3</a>) is a copyleft license: You allow A to see and alter the code, but if A gives (or sells) it to B, A must allow B to see and alter as well. V3 is similar to V2, but if A puts the code in a machine, then gives that machine to B, the machine must allow B to run altered versions of the code.

josinalvo:
How about that ?

I changed the title to help with #51 concerns
(and, for some reason, changed the whole damn text ...)

bkeepers:
Thanks for the suggestion. Since this discussion hasn't been active for a while, I'm going to close this for now. We're happy to continue the conversation though.

Issue #109, opened 2013-10-25
BSD (3-Clause) License license should populate all placeholders


izuzak:
Creating a repository with the BSD (3-Clause) License doesn't define values for all placeholders in the license. 

Specifically, the `{organization}` placeholder is not replaced with any value, and this has to be done manually after creating the repository:

https://github.com/github/choosealicense.com/blob/gh-pages/licenses/bsd-3-clause.txt#L38

A better approach might be to replace `{organization}` with `{fullname}`.

cc @Haacked @benbalter

bhuga:
These interpolate correctly on .com again. If we want to update the licenses it's another thing.

haacked:
@bhuga huh? I didn't understand your comment. Could you clarify?

bhuga:
We had a bug for a while were almost nothing would get interpolated. Creating a new repo with a BSD-3 now adds a fullname: https://github.com/bhuga/testlicense/blob/master/LICENSE#L1

haacked:
@bhuga ah! Thanks for fixing that. But this bug is specifically about the `{organization}` field in the BSD-3.

> Neither the name of the {organization} nor the names of it

You can see it a few lines into your test license. The suggestion is to change that to `fullname` instead of `organization`. OR, perhaps we change the code on the .com side to fill it in. Which I'm starting to think is preferred because I'd rather not mess with the license texts.

bhuga:
Ah, so we just missed that one in https://github.com/github/choosealicense.com/pull/55! That's my bad; sorry, I misunderstood.

Should be as simple as updating it to `[fullname]` and re-importing to .com.

haacked:
So what I'm wondering is, can we make it work by keeping it [organization] here? Or is that just dumb? I'm just a tiny bit wary of changing the text on this site to something it's not in the original.

bhuga:
We don't replace `[organization]`, just `[fullname]` of the repository owner (because which organization would it be? What if there is none?). So it would read a bit weird for personal ownership; I suppose that whole line would have to be removed, and, well...yuck?

Not sure we can solve this one cleanly and easily.

haacked:
Well I guess 999 times out of 1000, the person applying this license is probably also the "organization". So maybe we just also replace "[organization]" with the "[fullname]". And the user can correct it.

[...]

PR #262, opened 2015-04-14
Update  bsd-3-clause.txt

lord63:
Replace the not exist {organization} option with [project].

The discuss can be found in #109
{organiztion} was changed to [project] in #208

By the way, I think we can use this way in the [opensource.org](http://opensource.org/licenses/BSD-3-Clause):

> 1. Neither the name of the copyright holder nor the names of its contributors may be used to endorse or promote products derived from this software without specific prior written permission.

Or just keep the original.

Any ideas? :smiley:


benbalter:
Great catch @lord63. Thanks@

PR #263, opened 2015-04-26
Updated year of copyright to 2015

AchoArnold:


Jitrixis:
Not agree with this pull, the year after the copyright notice is the year of the first publication. It shouldn't be updated. https://en.wikipedia.org/wiki/Copyright#Copyright_notices_in_the_United_States

**_Propose corrected pull : #264**_

bkeepers:
#264 was merged.

PR #266, opened 2015-05-29
MS Open Source licenses

matt40k:
Add the the Microsoft Open Source licenses which are heavily used on
Codeplex

### DIFF
--- a/_licenses/ms-pl.txt
+++ b/_licenses/ms-pl.txt
@@ -1,12 +1,12 @@
 ---
-layout: license
 title: Microsoft Public License
 hidden: true  
-permalink: /licenses/ms-pl/
 
 source: http://opensource.org/licenses/ms-pl
 
 description: "Microsoft Open Source"
+
+how: Create a text file (typically named LICENSE or LICENSE.txt) in the root of your source code and copy the text of the license into the file.
 
 required:
   - include-copyright
--- a/_licenses/ms-pl.txt
+++ b/_licenses/ms-pl.txt
@@ -1,13 +1,12 @@
 ---
-title: Microsoft Public License
-nickname: Ms-PL
-category: Microsoft
 layout: license
-permalink: /licenses/ms-pl/
-source: http://opensource.org/licenses/ms-pl
-redirect_from: /licenses/ms/
-description: "Microsoft Open Source"
+title: Microsoft Public License
 hidden: true  
+permalink: /licenses/ms-pl/
+
+source: http://opensource.org/licenses/ms-pl
+
+description: "Microsoft Open Source"
 
 required:
   - include-copyright
--- a/_licenses/ms-pl.txt
+++ b/_licenses/ms-pl.txt
@@ -7,8 +7,7 @@ permalink: /licenses/ms-pl/
 source: http://opensource.org/licenses/ms-pl
 redirect_from: /licenses/ms/
 description: "Microsoft Open Source"
-hide-from-license-list: true
-tab-slug: ms-pl
+hidden: true  
 
 required:
   - include-copyright
--- a/_licenses/ms-pl.txt
+++ b/_licenses/ms-pl.txt
@@ -1,0 +1,77 @@
+---
+title: Microsoft Public
+nickname: Ms-PL
+category: Microsoft
+layout: license
+permalink: /licenses/ms-pl/
+source: http://opensource.org/licenses/ms-pl
+redirect_from: /licenses/ms/
+description: "Microsoft Open Source"
+hide-from-license-list: true
+tab-slug: ms-pl
+
+required:
+  - include-copyright
+
+permitted:
+  - commercial-use
+  - modifications
+  - distribution
+  - patent-grant
+  - private-use
+
+forbidden:
+  - no-liability
+  - trademark-use
+---
+
+                Microsoft Public License (MS-PL)
+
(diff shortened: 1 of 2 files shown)


benbalter:
@matt40k Can you talk a bit more about why you think the license should be added? Who uses it and why someone would prefer an MS license over one of the existing licenses listed?

wind39:
I think many projects on Codeplex are migrating to Github. Even [Microsoft's .NET open source projects](https://github.com/dotnet) are on Github too, under MIT License.

I don't see the need of MS License here.

matt40k:
@benbalter - the primary reason would to keep the migration process simple - you don't have to re-license. 

@wind39 - Those projects weren't ported from CodePlex however, when they made the decision to release the code, part of that decision, to choose GitHub over their own (CodePlex), would have been selecting a different license. If you look at the MS SQL Server samples which is hosted on Codeplex (https://msftdbprodsamples.codeplex.com/license) it is licensed under Ms-PL.

I personally found the MS licenses simpler to understand.

strugee:
> @benbalter - the primary reason would to keep the migration process simple - you don't have to re-license.

You can use whatever license you want on GitHub, regardless of whether or not it's in ChooseALicense. The only thing on GitHub.com that ChooseALicense affects is the "license" dropdown when you create a repo. And all that does is automatically commit a `LICENSE` file into your repo. You don't have to relicense to migrate to GitHub.

bkeepers:
It sounds like there is not a compelling reason to add this right now. We're happy to reconsider if it begins to receive more adoption or other circumstances change.

matt40k:
I think it would be silly if GitHub allowed any license, after all, someone could post some code up, then have a truly horrid license that said if anyone actually uses it, they are required to summit their first born child for sacrifice or something equally ridiculous. At the moment it actually looks like this could be possible based on the T&C so long as GitHub is legally allowed to distribute it. Anyway, I digress off the point. Happy its at least its got an answer, even if it isn't positive outcome

strugee:
> I think it would be silly if GitHub allowed any license, after all, someone could post some code up, then have a truly horrid license that said if anyone actually uses it, they are required to summit their first born child for sacrifice or something equally ridiculous.

The problem with this kind of thinking is that we all have different opinions on what is "ridiculous". E.g. some people might think that GitHub should ban copyleft licenses and some would literally leave GitHub if they did that. It's not GitHub's place to decide licenses; only the copyright holder can do that.

Also, think of the pain of bugging GitHub every time someone wrote a new license.

bkeepers:
>  It's not GitHub's place to decide licenses; only the copyright holder can do that.

:+1: 

The code you put on GitHub is yours, not ours. There are many legitimate reasons for having a repository public and not have it open source. We think the OSI has done a great job of maintaining the open source definition, and defer to them in terms of which licenses qualify as "open source".

matt40k:
@strugee - You seem to be confusing approved with meeting requirements. GitHub MUST have some requirements, it MUST have rights to distribute software. It must have something about meeting complying with US laws. It's like the roman army, they didn't care what gods you prayed to, if any, so long as they didn't get in the way of your duty. GitHub doesn't care what license you use, just so long as it doesn't interfere with their biz.

ChooseALicense.com has Travis step to check that the license is registered with SPDX License list - https://spdx.org/licenses/ - All though this isn't GitHub, it does show at least some criteria required to get a license added, the only other criteria I can see is that it must be considered "cool". I can't really see much difference between BSD and MIT, still I am not a lawyer.

I could accept if this particular request got rejected as its from commercial company that looks like self promotion, ie Microsoft created as they want people to believe there is Microsoft and nothing else - ala Steve Ballmer days (although you could argue that about MIT, however the key would be commercial).

Anyway, it'll be nice if you could at least say, hey, we only have the popular licenses listed here as defined by the licenses already in use in the wild - using something like openhub.net to gather stats. 

Ideally however, you could just as easy automatically build the spdx list and have them list under "Other" then leave the current list as the popular group - this will at least promote the discussion of licenses - which after all, isn't that why this project exists?

strugee:
> Anyway, it'll be nice if you could at least say, hey, we only have the popular licenses listed here as defined by the licenses already in use in the wild - using something like openhub.net to gather stats.

https://github.com/github/choosealicense.com/#immediate-goals

> this will at least promote the discussion of licenses - which after all, isn't that why this project exists?

No, not really. This project exists in response to the fact that a vast number of GitHub repositories have no license, because their authors don't realize that just throwing the code up on GitHub isn't legally enough. This project is curated, not comprehensive, because if it were comprehensive people who have never encountered the idea of a license before would get overwhelmed. That's why your pull got rejected.

benbalter:
@matt40k I'd be happy to merge this given a few changes:
1. Per https://github.com/github/choosealicense.com/issues/273, please add `hidden: true` to the YAML front matter.
2. Remove the `tab-slug` and `hide-from-license-list` keys, they are only used for license families
3. Where did the indentation come from? Is this the "canonical" indentation?

matt40k:
1. \- Pass, was so long ago. Happy to remove.

Am I being thick, but is the Travis CI build failing because of another license? Or am I missing something?

278- ./_site/licenses/ofl/index.html
279  \*  htmlParseEntityRef: no name

benbalter:
Thanks @matt40k! :v: :tada: :necktie:

PR #267, opened 2015-07-05
Fix broken links in CONTRIBUTING.md

jdennes:
There were a couple of broken links which were hard-coded to use the master branch, which doesn't exist. This makes those links branch-agnostic. /cc @benbalter


benbalter:
Nice catch. Thanks @jdennes!

Issue #273, opened 2015-08-05
Proposal: "Hidden" licenses


benbalter:
I'd like to propose that we introduce a new "class" of licenses to choosealicense.com which I'm calling "hidden licenses". Hidden licenses would act just like any other license, except they would not appear on http://choosealicense.com/licenses/.

We get **a lot** of requests to add licenses representing the extremely long tail of license choices out their in the open source community. See https://github.com/github/choosealicense.com/pull/248, https://github.com/github/choosealicense.com/pull/216, https://github.com/github/choosealicense.com/pull/200, https://github.com/github/choosealicense.com/pull/130, https://github.com/github/choosealicense.com/pull/86, https://github.com/github/choosealicense.com/pull/250, https://github.com/github/choosealicense.com/pull/266, https://github.com/github/choosealicense.com/pull/210, https://github.com/github/choosealicense.com/pull/195, https://github.com/github/choosealicense.com/pull/135, https://github.com/github/choosealicense.com/pull/95, https://github.com/github/choosealicense.com/pull/15, and https://github.com/github/choosealicense.com/pull/17 to name a few.

Traditionally, we've been hesitant to add licenses beyond the dozen or so most popular licenses. [The goal of choosealiense.com](https://github.com/github/choosealicense.com#immediate-goals) is not to be comprehensive. Instead, it's designed to _simplify_ the process of choosing a license. That's why the homepage, the primary route, has just three licenses, with blurbs that may not 100% reflect every nuance of the licenses' differences. 

Adding more licenses to the mental decision tree a user makes complicates the process and is antithetical to the site's goals. Decision paralysis is a real thing. see https://github.com/github/choosealicense.com/issues/229#issuecomment-105289767 for an example of this in action.

That said, I think there's a case for a third class of licenses (beyond featured and not-featured) that have received a certain base-level of notoriety such that a typical developer is likely to come across a licensed projects in their day-to-day interactions, but aren't widely used enough to warrant being presented as an option.

There's a few reasons for this:
- If a developer comes across a licensed project they can search for or be linked to the license page on choosealicense.com to better understand what they can and can't do under the license.
- It can provide a softer onramp for new licenses. Rather than spending time in discussions that almost always seem to devolve into the relative merits of a given license (which is not what the decision turns on), licenses with a certain numeric threshold (say 10k projects on GitHub via code search), can be added as a hidden license where the description and attributes can begin to be curated.
[...]

PR #274, opened 2015-08-06
Hidden licenses

benbalter:
This pull request implements the "hidden license" functionality as described in https://github.com/github/choosealicense.com/issues/273. Licenses with `hidden: true` in their YAML front matter will not appear on the `/licenses` list.

In order to keep the front matter keys sane, `hide-from-license-list`, which really was used to described license variants, is now `variant`, with e.g., `GPLv3` being a "variant" of the GPL family of licenses.

Licenses default to `hidden: false` via `_config.yml` unless otherwise specified.

Fixes https://github.com/github/choosealicense.com/issues/273.

### DIFF
--- a/licenses.html
+++ b/licenses.html
@@ -17,7 +17,7 @@ title: Licenses
   Some communities tend to have specific licenses preferred by the
   community. For example, Perl developers often choose the Artistic License.
 </p>
-{% assign licenses = site.licenses | where:"featured",false | where:"hide-from-license-list",false %}
+{% assign licenses = site.licenses | where:"featured",false | where:"hidden",false | where:"variant",false %}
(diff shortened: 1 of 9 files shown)


licenses.html
```diff
@@ -17,7 +17,7 @@
   Some communities tend to have specific licenses preferred by the
   community. For example, Perl developers often choose the Artistic License.
 </p>
-{% assign licenses = site.licenses | where:"featured",false | where:"hide-from-license-list",false %}
+{% assign licenses = site.licenses | where:"featured",false | where:"hidden",false | where:"variant",false %}
```
mislav, line 20: TIL `where` filter in Liquid :smile: :+1:

PR #286, opened 2015-09-01
Remove recommendation to add email address to ISC

### LINKED ISSUE #285: ISC license incorrectly recommends email
Last week I noticed that the ISC license contains the following line, which was added in f89201c7f4e01e5addf2f63302cfc61132ce962d (which was merged in #50):

```
Copyright (c) [year], [fullname] <[email]>
```

The value of the `how` key also contains: "...[email] is optional but recommended.", which I've been unable to substantiate anywhere. Seeing as how an email address is never recommended in any other license (although it's explicitly required by both the OFL 1.1 and the WTFPL), is there any reason not to remove this? I'd be more than happy to put together a pull request.

Copies of the canonical ISC license are available through the [ISC](http://www.isc.org/downloads/software-support-policy/isc-license/) and the [OSI](http://opensource.org/licenses/isc-license.txt).

christianbundy:
Remove non-canonical license and usage text from f89201c

Resolves #285

### DIFF
--- a/_licenses/isc.txt
+++ b/_licenses/isc.txt
@@ -7,7 +7,7 @@ tab-slug: isc
 
 description: A permissive license lets people do anything with your code with proper attribution and without warranty. The ISC license is functionally equivalent to the <a href="/licenses/bsd">BSD 2-Clause</a> and <a href="/licenses/mit">MIT</a> licenses, removing some language that is no longer necessary.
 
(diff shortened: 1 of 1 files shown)


bkeepers:
This makes sense to me after reading the discussion in #285.

@benbalter thoughts?

benbalter:
Agree. Thanks @christianbundy! :game_die: :baby_chick: :ocean:

PR #294, opened 2015-09-18
Remove background

bkeepers:
The textured background makes the site feel a little dated to me. I'm not familiar with the history of the design, but was curious what everyone would think about dropping the background. Thoughts?

choosing_an_oss_license_doesn t_need_to_be_scary_-_choosealicense_com

licenses_-_choosealicense_com

apache_license_2_0_-_choosealicense_com

/cc @github/choosealicense-com


benbalter:
I want to say it might have been @cobyism's doing?

jasonlong:
Yep, that was my doing and I definitely agree @bkeepers. :thumbsup: from me.

bkeepers:
:cool:, thanks @jasonlong.  Anyone else have thoughts?

benbalter:
:shipit:

PR #295, opened 2015-09-20
Remove flash detection

### LINKED ISSUE #235: Switching tabs in licenses doesn't work without Flash player

bkeepers:
Fixes #235

This removes the attempt to detect if flash is available that was added in #26.

Instead, it attempts to initialize ZeroClipboard and also sets up the fallback. ZeroClipboard will be a no-op if flash is not available, in which case the fallback will already be configured. If flash is available, it will put a flash object over the button and the fallback will never get called.

I tested this with both flash enabled and disabled on IE 10 on Windows, and the latest Safari, Chrome and Firefox on the mac.

/cc @jonrohan @XhmikosR

### DIFF
--- a/assets/js/app.coffee
+++ b/assets/js/app.coffee
@@ -2,14 +2,6 @@
 ---
 
 class Choosealicense
-
-  # Checks if Flash is available in the client.
-  flashAvailable: ->
-    if ActiveXObject?
-      !!(new ActiveXObject("ShockwaveFlash.ShockwaveFlash"))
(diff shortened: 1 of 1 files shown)


XhmikosR:
Unfortunately I can't try it locally but if you tried it on IE10 without Flash installed and tabs switching worked it should be OK.

benbalter:
> 2 additions and 20 deletions.

Always for that. :+1:

PR #296, opened 2015-09-25
Add AFL 3.0 as a hidden license

benbalter:
AFL 3.0 is OFL 3.0 without the copyleft requirements. Adding to resolve ambiguity between AFL and OFL when using choosealicense as a base of comparison.

/cc @vmg, @bkeepers


vmg:
:+1::+1::+1:

PR #300, opened 2015-10-23
Minor SEO optimizations

benbalter:
I feel gross doing it, but it's a necessary evil. Right now if you google for common keywords (license names, etc.) we're not even in the results. Let's fix that.

Specifically:
- Rename site in `title` tag from `ChooseALicense.com` to `Choose a License` 
- Add open graph (Facebook, Slack, etc.) and Twitter meta title and description tags
- Add canonical URL
- Change homepage tagline from  "Choosing an ~~OSS~~ license doesn’t need to be scary" to "Choosing an **open source** license doesn’t need to be scary"

Otherwise, human-readable output should be the same. Not sure if there's anything else we can/should be doing.

/cc @bkeepers, @github/seo

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1,6 +1,6 @@
 ---
 title: Choose a License
-description: A site to provide non-judgmental guidance on choosing a license for your open source project
+description: Non-judgmental guidance on choosing a license for your open source project
 relative_permalinks: false
 markdown: kramdown
 url: "http://choosealicense.com"
--- a/_config.yml
+++ b/_config.yml
@@ -1,5 +1,5 @@
 ---
-title: ChooseALicense.com
+title: Choose a License
 relative_permalinks: false
 markdown: kramdown
 url: "http://choosealicense.com"
@@ -43,3 +43,7 @@ gems:
 sass:
     sass_dir: _sass
     style: :compressed
+
+twitter:
+  username: "@github"
+  account_id: "13334762"
--- a/_config.yml
+++ b/_config.yml
@@ -47,4 +47,3 @@ sass:
 
 twitter:
   username: "@github"
-  account_id: "13334762"
--- a/_config.yml
+++ b/_config.yml
@@ -1,5 +1,6 @@
 ---
 title: Choose a License
+description: A site to provide non-judgmental guidance on choosing a license for your open source project
 relative_permalinks: false
 markdown: kramdown
 url: "http://choosealicense.com"
--- a/_includes/seo.html
+++ b/_includes/seo.html
@@ -14,11 +14,12 @@
     <title>{{ seo_title }}</title>
     <meta property="og:title" content="{{ seo_title }}"/>
     <meta property="og:site_name" content="{{ site.title }}"/>
+    <title itemprop='name'>{{ site.title }}</title>
 
     <meta name="description" content="{{ seo_description }}">
     <meta property='og:description' content="{{ seo_description }}"/>
 
-    <link rel="canonical" href="{{ site.github.url }}{{ page.url }}"/>
+    <link rel="canonical" href="{{ site.github.url }}{{ page.url }}" itemprop="url"/>
     <meta property='og:url' content='{{ site.github.url }}{{ page.url }}'/>
     <meta property='og:locale' content='en_us'/>
 
--- a/_includes/seo.html
+++ b/_includes/seo.html
@@ -7,7 +7,7 @@
     {% if page.description %}
       {% assign seo_description = page.description | strip_html | strip_newlines %}
     {% else %}
-      {% assign seo_description = site.tagline | strip_html | strip_newlines %}
+      {% assign seo_description = site.description | strip_html | strip_newlines %}
     {% endif %}
 
     <!-- le seo -->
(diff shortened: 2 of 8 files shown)


_includes/seo.html
```diff
@@ -0,0 +1,31 @@
...
+    <link rel="canonical" href="{{ site.github.url }}{{ page.url }}"/>
+    <meta property='og:url' content='{{ site.github.url }}{{ page.url }}'/>
+    <meta property='og:locale' content='en_us'/>
+
+    <meta property="twitter:account_id" content="{{ site.twitter.account_id }}" />
+    <meta property="twitter:card" content="summary" />
+    <meta property="twitter:title" content="{{ seo_title }}" />
+    <meta property="twitter:description" content="{{ seo_description }}" />
+    <meta property="twitter:site" content="{{ site.twitter.username }}" />
+    <meta property="twitter:creator" content="{{ site.twitter.username }}" />
```
leereilly: The `twitter:creator` field isn't necessary for regular summary cards.

_includes/seo.html
```diff
@@ -0,0 +1,31 @@
...
+    <meta property="og:site_name" content="{{ site.title }}"/>
+
+    <meta name="description" content="{{ seo_description }}">
+    <meta property='og:description' content="{{ seo_description }}"/>
+
+    <link rel="canonical" href="{{ site.github.url }}{{ page.url }}"/>
+    <meta property='og:url' content='{{ site.github.url }}{{ page.url }}'/>
+    <meta property='og:locale' content='en_us'/>
+
+    <meta property="twitter:account_id" content="{{ site.twitter.account_id }}" />
```
leereilly: Don't need `twitter:account_id` here.

_includes/seo.html
```diff
@@ -0,0 +1,31 @@
...
+    <meta property="twitter:card" content="summary" />
```
leereilly: The Twitter docs use `name` instead of `property`. It doesn't make any difference to them (or most consumers we'd care most about) as far as I can tell, but 3rd party consumers/validators sometimes complain about it i.e. consider using

``` html
<meta name="twitter:card" content="summary" />
```

instead of 

``` html
<meta property="twitter:card" content="summary" />
```

if you want to be a perfectionist :ok_hand:

benbalter:
Also added [JSON-LD breadcrumbs](https://developers.google.com/structured-data/breadcrumbs) in  2caf704

_config.yml
```diff
@@ -43,3 +44,7 @@ gems:
 sass:
     sass_dir: _sass
     style: :compressed
+
+twitter:
+  username: "@github"
+  account_id: "13334762"
```
leereilly: `account_id` not required here...

leereilly:
Love it, @benbalter! Is there a relevant image we could use for `twitter:image` and `og:image`?

_config.yml
```diff
@@ -1,5 +1,6 @@
 ---
-title: ChooseALicense.com
+title: Choose a License
+description: A site to provide non-judgmental guidance on choosing a license for your open source project
```
leereilly: Maybe shorten this to ~ "Non-judgmental guidance on choosing a license for your open source project"?

benbalter:
> Is there a relevant image we could use

Is it silly to use a screenshot of the home page?

screen shot 2015-10-23 at 6 15 36 pm

benbalter:
Also @leereilly thanks for dropping your :books:. Really appreciated.

leereilly:
> Is it silly to use a screenshot of the home page?

Yeah - I wouldn't advise that. The smaller text would be largely unreadable on most devices and it's probably a little too busy. An image with the text <kbd>Choosing an OSS license doesn’t need to be scary </kbd> would be... OK, but not great. 

I don't have any great ideas for it right now, but maybe you can pitch this as as small project in the github/creative? In the meantime, I'll keep my thinking hat on :+1:

index.html
```diff
@@ -2,8 +2,7 @@
 layout: default
 class: home
 hide_breadcrumbs: true
-title: Choosing an OSS license doesn’t need to be scary
-description: A site to provide non-judgmental guidance on choosing a license for your open source project
+title: Choosing an open source license doesn’t need to be scary
```
leereilly, line 5: By the same token, you might consider changing the `<h1>` on the index page similarly from:

image

to

image

The former definitely _looks_ nicer.

bkeepers:
:+1: nice work.

> Is there a relevant image we could use for `twitter:image` and `og:image`?

Happy stallman? :trollface:

index.html
benbalter, line 5: Changing the title in the YAML front matter updates the H1. Bumped the font size down to 60 from 63 and things seem better for both humans and robots IMHO:

screen shot 2015-10-24 at 1 13 14 pm

benbalter:
Thanks for the :eyes:. Since it's 99% backend changes, going to merge to see what Google sees, and can continue to optimize (e.g., the Twitter image) on a subsequent pass.

benbalter:
Not sure how much folks are Tweeting about licenses, but these are :fire::

screen shot 2015-10-24 at 1 19 13 pm
screen shot 2015-10-24 at 1 19 26 pm

waldyrious:
Since most licenses don't have an associated image/logo, how about standardizing on an image with each license's initials? I think it would be better than a single image for every license page.

Also, it could be good to implement a length limit on the description of each license -- e.g. the BSD/ISC description gets cut off in the Twitter snippet and probably elsewhere.

PR #302, opened 2015-10-31
Use Mozilla's plaint text version of the MPL

benbalter:
Via @richardfontana [over on Twitter](https://twitter.com/richardfontana/status/660472898948206592):

> GitHub license picker version of MPL 2.0 incorrect - compare https://mozorg.cdn.mozilla.net/media/MPL/2.0/index.815ca599c9df.txt https://raw.githubusercontent.com/richardfontana/mpl-test/master/LICENSE cc: @benbalter @tieguy @gerv

You can see a detailed comparison of the version @richardfontana linked in his Tweet to the current version [in this diff](https://github.com/github/choosealicense.com/compare/gh-pages...compare-plaintext-mpl?diff=unified&name=compare-plaintext-mpl&w=1). It doesn't appear there are any substantive changes, but [as he properly pointed out](https://twitter.com/richardfontana/status/660481425871921153):, dropping the conspicuous markings could potentially give rise to a UCC Issue (and cause some issues with license detection).

Looking at the commit history, it seems @haacked [added the license in Nov. 2012](https://github.com/github/choosealicense.com/commit/90da2c12b46d8b3e17a1686d53ed98557fb2b30d) as HTML, which he then [converted to plain text two days later](https://github.com/github/choosealicense.com/commit/25e402bd8919b95b53121f366f4c2e731d429dc8#diff-895566de4a5c612e96507c146c63425bL22).

The important thing to me is that we're using the canonical license text. It appears the version in the site (and thus the license dropdown and API on GitHub.com) is based off [the HTML version](https://www.mozilla.org/en-US/MPL/2.0/), but there's also [a plain text version](https://mozorg.cdn.mozilla.net/media/MPL/2.0/index.815ca599c9df.txt) linked [from Mozilla's MPL page](https://www.mozilla.org/en-US/MPL/) (although it's a bit odd that it's a versioned CDN url). 

This pull requests moves the license text over to Mozilla's "official" plain text version, rather than our bespoke, downgraded-from-HTML, plaintext version, if there's consensus that that is, in fact, the canonical version that we should be using.

### DIFF
--- a/_licenses/mpl-2.0.txt
+++ b/_licenses/mpl-2.0.txt
@@ -1,7 +1,7 @@
 ---
 title: Mozilla Public License 2.0
 redirect_from: /licenses/mozilla/
-source: http://www.mozilla.org/MPL/2.0/
+source: https://www.mozilla.org/media/MPL/2.0/index.txt
 
 description: The Mozilla Public License (MPL 2.0) is maintained by the Mozilla foundation. This license attempts to be a compromise between the permissive BSD license and the reciprocal GPL license.
 
--- a/_licenses/mpl-2.0.txt
+++ b/_licenses/mpl-2.0.txt
@@ -25,365 +25,376 @@ forbidden:
 
 ---
 
-Mozilla Public License, version 2.0
+Mozilla Public License Version 2.0
+==================================
 
 1. Definitions
+--------------
 
 1.1. "Contributor"
-
-     means each individual or legal entity that creates, contributes to the
-     creation of, or owns Covered Software.
+    means each individual or legal entity that creates, contributes to
+    the creation of, or owns Covered Software.
 
 1.2. "Contributor Version"
-
-     means the combination of the Contributions of others (if any) used by a
-     Contributor and that particular Contributor's Contribution.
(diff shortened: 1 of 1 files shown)


gerv:
Hey. Not sure what's going on with those CDN URLs, but the canonical URL to the canonical plain text version is http://www.mozilla.org/MPL/2.0/index.txt . Switching to that would be great :-)

benbalter:
> the canonical URL to the canonical plain text version is http://www.mozilla.org/MPL/2.0/index.txt . Switching to that would be great

Ahh. Couldn't find that. Thanks. The text is identical, but I updated the metadata to point to the plain-text version via 37c1d94.

bkeepers:
:+1: looks good to me.  We should probably check the other licenses in https://github.com/github/choosealicense.com/commit/25e402bd8919b95b53121f366f4c2e731d429dc8 for proper formatting.

benbalter:
Failed tests were an upstream issue, [now resolved](https://github.com/benbalter/jekyll-seo-tag/pull/3) so that tests pass. Will merge here and go back and verify the other OG licenses are their canonical version.

PR #303, opened 2015-11-02
Remove sublicense warning for GPL-3.0

### LINKED ISSUE #10: Sublicensing warning inappropriate for GPLv3
There's a warning about sublicensing being forbidden for GPLv3 & v2 (and Affero). While this is technically correct, I think a warning is inappropriate as the risk to the license user is minimal.  

Clause 10 explains that, instead of you sublicensing the original work, 
    "Each time you convey a covered work, the recipient automatically receives 
    a license from the original licensors, to run, modify and propagate that work, 
    subject to this License."  
Sublicensing is thus redundant, since every recipient has a license to the original work from the copyright holder, plus a license from you to your modifications.

I suggest just removing the warning.

bkeepers:
From Section 10 of the GPL:

> Each time you convey a covered work, the recipient automatically
> receives a license from the original licensors, to run, modify and
> propagate that work, subject to this License.

Sublicensing is technically forbidden, but the copyleft makes the ability to sublicense irrelevant.

Fixes #10 
Closes #224

/cc @webmink @benbalter

### DIFF
--- a/_licenses/gpl-3.0.txt
+++ b/_licenses/gpl-3.0.txt
@@ -27,7 +27,6 @@ permitted:
 
 forbidden:
   - no-liability
-  - no-sublicense
 
 ---
 


benbalter:
Makes sense. Good call.

PR #305, opened 2015-11-02
Rename "Patent Grant" to "Patent Use"

### LINKED ISSUE #168: move "Patent Grant" from "Permitted" list to "Required" list
Hope i'm not resurrecting a resolved issue (I looked and couldn't trace one):

If I'm not misreading the licenses, granting rights to their patents is part of what the contributors are required to do under some of the licenses and therefore belongs under the "Required" list and not the "Permitted" list. (Everybody else is then permitted to _use_ and not to _grant_ those patents)

bkeepers:
Per the discussion in #168, the consumers of the softare are granted the right to use patents. So "Patent Use" makes more sense from a consumer perspective than "Patent Grant".

Fixes #168

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ The license properties (rules) are stored as a bulleted list within the licenses
 * `distribution` - You may distribute this software.
 * `sublicense` - You may grant a sublicense to modify and distribute this software to third parties not included in the license.
 * `private-use` - You may use and modify the software without distributing it.
-* `patent-grant` - This license provides an express grant of patent rights from the contributor to the recipient.
+* `patent-use` - This license provides an express grant of patent rights from the contributor to the recipient.
 
 #### Forbidden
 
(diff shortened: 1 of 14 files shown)


benbalter:
>  makes more sense from a consumer perspective

Realizing I said the exact opposite over in (#168), should the rules be worded from the code publisher or the code consumer's perspective? From [the README](https://github.com/github/choosealicense.com#immediate-goals):

> Our goal is to help you find a license that meets your goals.

I suspect the most common use case is a code publisher coming to the site to choose a license for their code, not a code consumer coming to the site to learn what they can do under a certain license (and TLDR may solve for that better with its long tail of licenses).

Not super strongly opinionated, but we should figure out our target audience (consumers or publishers) and optimize the language accordingly (even if it's been an admittedly squishy concept in the past).

bkeepers:
The target audience is definitely publishers, but the Required/Permitted/Forbidden section is about what _consumers_ can do. Does that make sense?

benbalter:
Yeah, that makes sense. In my mind, "you" in the tooltips is almost the publisher speaking to the consumer, or put another way, is the same tense as it would be / is in the license itself. To do anything else (e.g., "downstream consumers are required to...") makes it too cumbersome. :+1:

PR #307, opened 2015-11-02
Add page around licensing content, Take II

afeld:
Moved from #175.

Tweaked the language to not make an explicit recommendation.


afeld:
Heh, the build seems to fail since the page is not actually live yet.

mlinksva:
This is definitely an improvement over what's currently on the home page. I'm going to incorporate these commits into a new PR further in this direction after #33 (adding CC-BY-4.0 and CC-BY-SA-4.0 to the collection).

PR #308, opened 2015-11-17
Prefer (L)GPL 3.0 to 2.x.

mina86:
The GPL v3 is an improvement over v2 fixing several issues and
increasing compatibility with other free software licenses.  As such
there is no reason to feature v2 prominently rather than v3.  See
http://www.gnu.org/licenses/rms-why-gplv3.en.html for more rationale.


bkeepers:
Thanks for the pull request, and the link to a primary source.

PR #317, opened 2016-01-14
Create goc-1.0

nutellarlz:


strugee:
The description of this license seems to be copied from the GNU GPL's description (GOC is certainly not the most widely-used license for free software) but even if it wasn't, this appears to be a license that you've written. ChooseALicense focuses on quality, not quantity. As such we only list very widely-used licenses.

In fact, I'd like to strongly encourage you to not use GOC at all (and perhaps entirely remove it from the Internet) due to the fact that it contributes to [license proliferation](https://en.wikipedia.org/wiki/License_proliferation), which is a serious problem that causes lots of headaches for everyone.

strugee:
@JTechMe I just read your license and AFAICT it has no practical difference to the GPL, besides being far, far less proven and probably providing more loopholes to violate the spirit of the license.

Again, please consider switching to using the GPL in the interest of stopping license proliferation.

JTechMe:
@strugee

PR #319, opened 2016-01-19
Added license lppl-1-3c.txt

sieversMartin:
This is the standard license for developing LaTeX packages and classes. I kindly ask to add this to the set of licenses. Thank you.


waldyrious:
> I kindly to add this

I believe you accidentally the verb :)

sieversMartin:
You're right.

sieversMartin:
Does anyone have an idea, why the cibuild script still fails? The name "LaTeX Project Public License v1.3c" is correctly set and the Spex data should be available. However the scripts looks for the ID lppl-1-3c instead of lppl-1.3c.

benbalter:
> However the scripts looks for the ID lppl-1-3c instead of lppl-1.3c.

@sieversMartin I believe that's because of the license's filename, which should be its SPDX ID.

sieversMartin:
Well, so I have to contact the LaTeX project and ask them to rename their files? Or do you mean the filename I chose? In this case I'll try that first.

sieversMartin:
I still have a problem: External link http://choosealicense.com/licenses/lppl-1.3c/ failed: 404 No error

Can I solve that by myself?

_licenses/lppl-1.3c.txt
```diff
@@ -0,0 +1,448 @@
...
+required:
+  - include-copyright
+  - document-changes 
+  - disclose-source
+  
+permitted:
+  - commercial-use
+  - modifications
+  - distribution
+  - sublicense
```
mlinksva: I don't see this in the license, suggest dropping this line.

_licenses/lppl-1.3c.txt
```diff
@@ -0,0 +1,448 @@
...
+  - patent-use
```
mlinksva: Nor this.

mlinksva:
@sieversMartin http://choosealicense.com/licenses/lppl-1.3c/ will only work after the PR is merged. I noted a couple things that should be changed before that in line notes.

sieversMartin:
Okay, I removed these two lines.

benbalter:
That test shouldn't be failing, because it should be using the local version it builds, not the live version (and the license shouldn't be listed on `/licenses/` if it's hidden). I'll take a closer look at the test.

mlinksva:
@sieversMartin if you pull newer commits into your branch the tests should then succeed.

sieversMartin:
Thank you. Done.

mlinksva:
@sieversMartin I don't know the LPPL well and its community background/practice almost not at all. I wondered if you could shed any light on [whether](https://github.com/github/choosealicense.com/pull/344#issuecomment-180837227) `disclose-source` is accurate for LPPL-1.3c? I don't see it, but my reading is naive.

wking:
On Sat, Feb 06, 2016 at 11:03:43AM -0800, Mike Linksvayer wrote:

> … whether `disclose-source` is accurate for LPPL-1.3c? I don't see
> it, but my reading is naive.

I'm also naive, but §6.d looks like an effective “Derived Work
distributors must disclose the source of the original Work” to me.
That seems more like a strengthened attribution requirement than a
share-alike requirement, because the licensee doesn't have to release
their modifications (they just have to describe them for §6.b).  But
I'm not sure which source (the Work, a Derived Work, either, or both)
the ‘disclose-source’ label is trying to cover [1](https://github.com/github/choosealicense.com/pull/344#issuecomment-180621772).

sieversMartin:
@mlinksva Another site classifies it as a must: https://tldrlegal.com/license/latex-project-public-license-v1.3c-%28lppl-1.3c%29. To be sure, I could ask the people from the LaTeX project.

mlinksva:
Isn't all that clear to me but I suppose you're right @wking ... 6d doesn't directly require source of the original work, but if that's the form the unmodified original came in, which is probably the usual case for works under the license, then it is effectively as you describe.

I suppose 6b's "complete and accurate log of the changes" could even be taken to require disclosure as source patches, as anything less would be incomplete and likely not 100% accurate. (6a regulates how source of modifications may be disclosed, but doesn't require disclosure).

OK, I'm at least naively convinced it is fair to leave `disclose-source` for LPPL 1.3c. @sieversMartin if you get a chance to ask people from the LaTeX project who understand the history and practice and could confirm or disconfirm, that'd be helpful.

PR #320, opened 2016-01-19
added eupl-1.1.txt licence

EmidioStani:
Added eupl licence accordingly to the rules:

https://github.com/github/choosealicense.com/blob/gh-pages/CONTRIBUTING.md#adding-a-license

### DIFF
--- a/_licenses/eupl-1.1.txt
+++ b/_licenses/eupl-1.1.txt
@@ -7,7 +7,7 @@ family: EUPL
 featured: false
 source: https://joinup.ec.europa.eu/community/eupl/og_page/european-union-public-licence-eupl-v11
 
-description: The “European Union Public Licence” (EUPL) is the first European Free/Open Source Software (F/OSS) licence. It has been created on the initiative of the European Commission. It is now approved by the European Commission in 22 official languages of the European Union.
+description: The “European Union Public Licence” (EUPL) is a copyleft free/open source software license created on the initiative of and approved by the European Commission in 22 official languages of the European Union.
 
 how: Create a text file (typically named COPYING or LICENCE.txt) in the root of your source code and copy the text of the license into the file.
 
--- a/_licenses/eupl-1.1.txt
+++ b/_licenses/eupl-1.1.txt
@@ -1,0 +1,346 @@
+﻿---
+title: European Union Public License 1.1
+nickname: EUPL-1.1
+tab-slug: eupl-v1.1
+redirect_from: /licenses/eupl-v1.1/
+family: EUPL
+featured: false
+source: https://joinup.ec.europa.eu/community/eupl/og_page/european-union-public-licence-eupl-v11
+
+description: The “European Union Public Licence” (EUPL) The EUPL is the first European Free/Open Source Software (F/OSS) licence. It has been created on the initiative of the European Commission. It is now approved by the European Commission in 22 official languages of the European Union.
+
+how: Create a text file (typically named COPYING or LICENCE.txt) in the root of your source code and copy the text of the license into the file.
+
+note: The European Commission recommends taking the additional step of adding a boilerplate notice to the top of each file. The boilerplate can be found at https://joinup.ec.europa.eu/sites/default/files/ckeditor_files/files/EUPL%201_1%20Guidelines%20EN%20Joinup.pdf
+
+required:
+  - include-copyright
+  - disclose-source
+  - document-changes
+  - network-use-disclose
+
+permitted:
+  - commercial-use
+  - modifications
+  - distribution
+  - patent-use
+  - private-use
+  - sublicense combined derivatives under a compatible licence
+
+forbidden:
+  - no-liability
+  - trademark use
+
+---
+
+European Union Public Licence
+V. 1.1
+
+
+EUPL © the European Community 2007
+
+
+This European Union Public Licence (the “EUPL”) applies to the
+Work or Software (as defined below) which is provided under the terms of this
+Licence. Any use of the Work, other than as authorised under this Licence is
+prohibited (to the extent such use is covered by a right of the copyright
+holder of the Work).
+
+The Original Work is provided under the terms of this
+Licence when the Licensor (as defined below) has placed the following notice
+immediately following the copyright notice for the Original Work:
+
+Licensed under the EUPL V.1.1
+
+or has expressed by any other mean his willingness to license under the EUPL.
+
+
+1. Definitions
+
+In this Licence, the
+following terms have the following meaning:
+
+- The Licence: this Licence.
+
+- The Original Work or the Software: the software distributed
+and/or communicated by the Licensor under this Licence, available as Source
+Code and also as Executable Code as the case may be.
+
+- Derivative Works:
+the works or software that could be created by the Licensee, based upon the
+Original Work or modifications thereof. This Licence does not define the
+extent of modification or dependence on the Original Work required in order to
+classify a work as a Derivative Work; this extent is determined by copyright
+law applicable in the country mentioned in Article 15. 
+
+- The Work: the Original Work and/or its Derivative Works.
+
+- The Source Code: the human-readable form of the Work which is the most
+convenient for people to study and modify.
+
+- The Executable Code: any code which has generally been compiled and which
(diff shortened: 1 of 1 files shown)


_licenses/eupl-1.1.txt
```diff
@@ -0,0 +1,346 @@
+﻿---
+title: European Union Public License 1.1
+nickname: EUPL-1.1
+tab-slug: eupl-v1.1
+redirect_from: /licenses/eupl-v1.1/
+family: EUPL
+featured: false
+source: https://joinup.ec.europa.eu/community/eupl/og_page/european-union-public-licence-eupl-v11
+
+description: The “European Union Public Licence” (EUPL) The EUPL is the first European Free/Open Source Software (F/OSS) licence. It has been created on the initiative of the European Commission. It is now approved by the European Commission in 22 official languages of the European Union.
```
mlinksva: "The EUPL" seems like a re-start of the sentence and could be removed. I don't think the "first" statement is true: CeCILL existed since 2004.

_licenses/eupl-1.1.txt
```diff
@@ -0,0 +1,346 @@
...
+  - document-changes
+  - network-use-disclose
+
+permitted:
+  - commercial-use
+  - modifications
+  - distribution
+  - patent-use
+  - private-use
+  - sublicense combined derivatives under a compatible licence
```
mlinksva: sublicense should be removed; properties don't allow for qualification here, and this one is used only for licenses that explicitly permit sublicensing, not ones that allow derivatives under a compatible license

_licenses/eupl-1.1.txt
```diff
@@ -0,0 +1,346 @@
...
+
+forbidden:
+  - no-liability
+  - trademark use
```
mlinksva: trademark-use (missing dash)

benbalter: Huh. That should have errored out the test. I'll take a look.

EmidioStani:
So I should fix all the issues above and re-submit ?

mlinksva:
Just fix the issues above in the branch you made this pull request for, and this pull request will be updated will the fixes.

EmidioStani:
ok i fixed the text, let me know if there is anything else to do.

_licenses/eupl-1.1.txt
```diff
@@ -0,0 +1,345 @@
+﻿---
+title: European Union Public License 1.1
+nickname: EUPL-1.1
+tab-slug: eupl-v1.1
+redirect_from: /licenses/eupl-v1.1/
+family: EUPL
+featured: false
+source: https://joinup.ec.europa.eu/community/eupl/og_page/european-union-public-licence-eupl-v11
+
+description: The “European Union Public Licence” (EUPL) is the first European Free/Open Source Software (F/OSS) licence. It has been created on the initiative of the European Commission. It is now approved by the European Commission in 22 official languages of the European Union.
```
mlinksva: I still don't see how "first European F/OSS license" is true given CeCILL. I suggest something like

description: The “European Union Public Licence” (EUPL) is a copyleft free/open source software license created on the initiative of and approved by the European Commission in 22 official languages of the European Union.

EmidioStani: Hello, thank you for the information. Indeed I meant EU license but the term European is confusing as there were previous licenses in some European Member States before EUPL. I will adapt the text. Thanks a lot for your comments!

EmidioStani:
ok I fixed the first sentence, thanks for let me know any other change

mlinksva:
Looks good. Thanks for submitting this and working through the changes.

EmidioStani:
Hello, juts one thing, how do I reach the license from the chooselicense.com, is there any link to it ?

mlinksva:
There aren't any links to http://choosealicense.com/licenses/eupl-1.1/ on the site other than http://choosealicense.com/sitemap.xml

I wonder if we shouldn't have an 'Even more licenses' section at the bottom of /licenses/ with just a list of otherwise hidden licenses with a sentence as to why they are there / linking to CONTRIBUTING.md?

waldyrious:
> I wonder if we shouldn't have an 'Even more licenses' section at the bottom of /licenses/

:+1:

benbalter:
> I wonder if we shouldn't have an 'Even more licenses' section at the bottom of /licenses/ with just a list of otherwise hidden licenses

I don't know that that would serve our [immediate goals](https://github.com/github/choosealicense.com#immediate-goals):

> Not comprehensive. Seems like an odd goal, but there are a bajillion licenses out there. We're going to have to filter that down to a small list of those that matter.

The primary goal of the site is to disambiguate license choices for code publishers. The more choices we present, the more cognitive overhead we require to complete that task.

waldyrious:
IMHO the fact that the extra licenses wouldn't be in the current expanded list (which is already one step removed from the front page), but tucked away in an even less accessible page, would be sufficient to prevent the cognitive overhead in most cases. Besides, if the extra licenses are included in the "choosealicense.com" project, it's only logical that they should somehow be reachable by navigating the "choosealicense.com" site, even if that takes a few extra steps and an attentive eye.

One of the great assets of this project is having these licenses in a standardized format that clearly summarizes their characteristics, which is lacking in most similar resources out there I know of -- so depriving people from having such a useful presentation for less popular licenses would be a net negative in my view -- for instance, it could make it harder for people to decide whether moving from a current obscure license to a more common one would make sense.

MattiSG:
@EmidioStani Are you 100% sure about the `network-use-disclose` part?

I see such a provision on Choosealicence.com, and in an ambiguous statement in the EU [compatibility table](https://joinup.ec.europa.eu/software/page/eupl/eupl-compatible-open-source-licences):

> Requesting for an exception should be facilitated by the fact the EUPL covers “software as a service” (SaaS) like the AGPL.

However, I don't read any such provision in the license text itself, and am worried that the compatible licenses only mention GPLv2, not AGPL.

It seems to me only the EUPLv1.2 could have such terms, as mentioned in the [draft notification](https://joinup.ec.europa.eu/community/eupl/news/public-consultation-draft-eupl-v-12-has-been-fruitful):

> The Annexe (main modification) increases downstream interoperability with more compatible licences: GPL v3, AGPL v3, LGPL, MPL v2.

mlinksva:
EUPL-1.1 (emphasis added):

> Distribution and/or Communication: any act of selling, giving, lending,
> renting, distributing, communicating, transmitting, or otherwise
> making available, on-line or off-line, copies of the Work **or providing access
> to its essential functionalities** at the disposal of any other natural or legal
> person.

EUPL-1.1's downstream compatibility list doesn't have disclosure or similar license as a condition of providing access to users over a network as a requirement for compatibility. I haven't seen anything written about this decision, but I imagine it was a compromise. Strong copyleft is required by the list.

MattiSG:
Thank you very much for the highlight, @mlinksva. I feel pretty stupid to not have spotted this in the two times I read it. I guess I was looking for something mentioning network access. This phrasing is actually even better.

I'm not sure I understand your second paragraph, though. What do you mean by _“Strong copyleft is required by the list”_?

Finally, on the _Compatibility clause_:

> If the Licensee Distributes and/or Communicates Derivative Works or copies, [it] can be done under
> the terms of this Compatible Licence. […] “Compatible Licence” refers to the licences listed in the appendix attached to this Licence. Should the Licensee’s obligations under the Compatible Licence conflict with his/her obligations under this Licence, the obligations of the Compatible Licence shall prevail.

I see a possible loophole here, since that could mean the EUPL network disclosure clause could be removed by “downgrading” to a GPLv2 license, for example. But I'm not sure this would be a case of “conflicting” obligations.

mlinksva:
I misremembered, should have left off "strong"; only copyleft is required. For the requirements of the EUPL compatibility list, see https://joinup.ec.europa.eu/system/files/doc/Doc3ef5.pdf#page=17

Yes there's a potential loophole (and an even bigger one with the not-so-strong EPL being on the list). My guess is that this is a compromise - at the time the compatibility list was created OSL and AGPLv1 were AFAIK the only licenses that attempted to govern use to provide network services, and they each were very little used. Compatibility with only those two licenses would not have created much flexibility for EUPL licensed projects to be combined with other copylefted projects.

PR #324, opened 2016-01-20
Make licenses hidden by default

benbalter:
The intent here is to make it easier for contributors to propose new licenses be added, without multiple rounds of back-and-forth to correct metadata, or extensive review by maintainers. In other words, let robots do the work, rather than humans.

There should be no change to the site output or metadata. Instead, this PR does three things:
1. Makes `hidden: true` the default for licenses in `_config.yml`
2. Adds `hidden: false` to previously shown licenses, and removes `hidden: true` from hidden licenses 
3. Adds a test to ensure the number of shown licenses does not expand unless we explicitly update the test.

Fixes https://github.com/github/choosealicense.com/issues/323.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -18,7 +18,7 @@ defaults:
     values:
       layout: license
       featured: false
-      hidden: false
+      hidden: true
       variant: false
 
 exclude:
@@ -41,7 +41,7 @@ gems:
   - jekyll-sitemap
   - jekyll-redirect-from
   - jekyll-seo-tag
-  
+
 sass:
     sass_dir: _sass
     style: :compressed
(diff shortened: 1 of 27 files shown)


mlinksva:
I don't think we should merge requests until they have correct metadata, which robots aren't going to correct, so I'm not sure I understand your first paragraph.

Whether explicit property is set on hidden or shown licenses doesn't make any difference to me; if it does to you and you think flipping a good idea, then I agree. :)

Test looks good.

benbalter:
> I don't think we should merge requests until they have correct metadata, which robots aren't going to correct, so I'm not sure I understand your first paragraph.

Ahh. I meant that a contributor could get immediate feedback on whether the license metadata beyond the legal properties (e.g., license name, hidden/not-hidden ,etc.) was at least in line with out contributor guidelines. There have been a few requests to add licenses that I had to say "can you please add `hidden: true` to the license (for less adopted licenses). This way, the contributor doesn't need to take that step, and if they do add `hidden: false`, the test should err out, or if they update the test, we'll easily see that in the diff. We'll obviously still have to review the diff and legal properties, but this will at least offload some of the basic "does it meet our guidelines" checks to Travis.

mlinksva:
:cool:

PR #325, opened 2016-01-20
Make it clear in the CONTRIBUTING instructions that the license file name should be the SPDX ID

benbalter:
Fixes https://github.com/github/choosealicense.com/issues/321.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,6 +22,7 @@ Choosealicense.com is intended to demystify license choices, not present all of 
 If your proposed license meets the above criteria, here's a few other things to keep in mind as you propose the license's addition:
 
 * Licenses live in the `/_licenses` folder.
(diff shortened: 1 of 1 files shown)


mlinksva:
Good addition, though we seem to lowercase the spdx id (which is usually upper but can be mixed case). Perhaps that should be noted.

I see that you raised the issue of case sensitivity on the spdx list and it [appears](https://bugs.linuxfoundation.org/show_bug.cgi?id=1327) they are going to document that identifiers are case insensitive in the next version of their spec so I guess our lowercasing is fine.

benbalter:
Yep. And [there was a request](https://github.com/benbalter/licensee/issues/72), to also output the mixed case SPDX ID as a property, e.g., for purposes of display, which is also on my list of things to tackle in a subsequent PR.

Issue #196, opened 2014-04-15
"No license" does not allow commercial or private use


asb:
Currently the [page for "no license"](http://choosealicense.com/licenses/no-license/) marks "Commercial Use" and "Private Use" as permitted, but this is false. You could argue that the Github TOS allow private use (as it explicitly says that you consent to public projects being forkable) but this doesn't apply elsewhere and it's still not something I'd rely on.

benbalter:
This was previously discussed, but I'm not finding the directly relevant issue (halp?). Two that may provide some context are https://github.com/github/choosealicense.com/pull/115 and https://github.com/github/choosealicense.com/pull/25.

haacked:
Yeah, this is a good point. Want to submit a PR?

benbalter:
The counter argument, is that there's always an implied license for _some_ use. You need to distinguish between unlicensed code knowingly posted publicly, and code sitting on my computer, which I neither post nor license. 

If the New York Times posts a news story on their website, there's an implied license for private use in that I am allowed to read it, print it, email it to a friend, etc. If I were a company, I could probably cut it out of the print edition and put it on my wall (e.g., a restaurant review).

There's some implied license for use for code knowingly posted publicly. If I wrote a blog post with a code snippet in it (or post it to stack exchange, etc.), you'd likely feel comfortable taking that one line and using it in a script you're writing for personal use, even without a formal license. Not sure where the line is.

haacked:
Thanks for the clarification @benbalter.

cubiclesoft:
Without a license, a developer opens themselves up to lawsuits.  There is a reason that licenses include the words "WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT." and then proceeds to include further statements about no liability.  Lawyers wrote and vetted such language because, without it, if the code "causes harm" (usually traceable to financial loss), the developer may be held liable for damages.

The argument about the New York Times is false.  The NYT has a Terms of Service document that clearly spells out what may and may not be done with their content:

https://www.nytimes.com/content/help/rights/terms/terms-of-service.html#b

In essence, they clarify their claim on their copyright materials and their lawyers put that document together.  There is no implied license since their claim on their content is clearly written and communicated.

[...]

PR #326, opened 2016-01-26
consolidate no license info in /no-license, remove from /licenses

mlinksva:
As noted in #196 (which this fixes) the provided metadata for "no license" isn't accurate. My preferred fix is to remove it. The option is available from the home page, there's no need to represent again on /licenses (no license isn't one).

I don't think this should break anything (e.g., "None" seems to be a hardcoded default in the GitHub new repo drop-down rather than being pulled from the choosealicense dataset) and documentation for the in-preview https://developer.github.com/v3/licenses/ doesn't show "no license" in response to list request. It's quite likely that I'm missing something.


benbalter:
The original thinking was the `/no-license` was intended for code publishers, to better understand what rights they grant (or retain), but _not_ licensing their code. The `licenses/no-license` was intended to be able to provide code consumers information on what rights they receive / don't receive, thinking of them less as licenses, and more like "license states". I'm not totally against it, per se, but if we remove the `licenses/no-licenses`, code consumers lose value, in that they no longer have the side bar rights metadata.

In terms of github, no-license is considered a "hidden" license, meaning it doesn't appear in the drop down or on `api.github.com/licenses`, but if a repository has no license (I think) we return the no-license metadata (but I may be wrong). Easy enough to fix upstream or modify the behavior depending on the above discussion.

mlinksva:
Thanks for the background thinking.

The side bar rights metadata isn't accurate so I don't think consumers are benefiting. But, good point on providing useful info to consumers. I added a practical section to /no-license for users. What do you think?

benbalter:
What metadata would you put in the sidebar, to make it accurate? We'll likely want to reimplement the "no-license" license within Licensee (or maybe leave it here, but make it hidden).

mlinksva:
`curl -H "Accept: application/vnd.github.drax-preview+json" https://api.github.com/repos/github/testrepo` obtains `"license": null`. That's the right behavior I think, rather than returning no license metadata. If we haven't detected a license, we just don't know. Returning the no license metadata would make it seem like we did.

The most accurate metadata for no license of the type we're providing for licenses is no metadata. For licenses, the metadata describes what the license permits, and the requirements and prohibitions that condition and scope those permissions. For no license, we have a different situation, where the metadata is trying to describe default copyright law. This is a lot messier. We can accurately say that the copyright holder hasn't given any express permissions, but the existing metadata scheme doesn't have a way of expressing that. What the user is free to do due to an implied license or copyright limitations and exceptions is unclear at best. Requirements just aren't relevant. Some of the prohibitions could be stated with relatively more certainty, but it's hard to not over-describe what is forbidden by default, make it seem what is not listed as forbidden might be allowed (ambiguous at best, as noted previously), or both. If there's a really strong use case for providing this type of metadata for default copyright status, I guess we could provide only forbiddens, but I'd rather just not try to describe default copyright, again, unless there is a really strong reason to do so.

benbalter:
All right.  You're logic makes sense (sorry it took me a while to get there). Once this merges, I'll deal with the impact to Licensee, to ensure we can keep pulling updates into GitHub.

**Edit**: Thanks for taking the time to walk me through your thinking.

mlinksva:
Thanks as always for taking the time to lay out the context and making me articulate my thinking. I'll go ahead and merge now.

PR #327, opened 2016-01-28
remove BOM from eupl-1.1.txt which caused it to not render

mlinksva:
BOM = byte order mark

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -10,7 +10,8 @@ end
 task :test do
   sh "bundle exec jekyll build --trace"
   Rake::Task["spec"].invoke
-  HTML::Proofer.new("./_site", :check_html => true).run
+  HTML::Proofer.new("./_site", :check_html => true,
+                    :href_swap => { %r{http://choosealicense.com} => "" }).run
 end
 
 task :approved_licenses do
--- a/_licenses/eupl-1.1.txt
+++ b/_licenses/eupl-1.1.txt
@@ -1,4 +1,4 @@
-﻿---
+---
 title: European Union Public License 1.1
 nickname: EUPL-1.1
 tab-slug: eupl-v1.1


benbalter:
:+1: I'm not sure why tests are failing. Merging in master _might_ resolve it.

mlinksva:
Actually, I can reproduce failure locally, which demonstrates I didn't run `rake test` before pushing.

Here's what I think's happening:
- _site/licenses/eupl-1.1/index.html contains a href to http://choosealicense.com/licenses/eupl-1.1/
- html-proofer checks all links in _site
- http://choosealicense.com/licenses/eupl-1.1/ (or any newly introduced link, not yet live) of course obtains 404

Braindead fix would be to have test tell html-proofer to ignore choosealicense.com links, with obvious downside of not catching unintentionally broken links within the site.

Sophisticated fix I guess would be to compare links in origin/gh-pages and local branch, and tell html-proofer to ignore choosealicense.ccom links introduced locally.

Or even better, looking at https://github.com/gjtorikian/html-proofer `href_swap` on choosealicense.com links to have those links tested locally rather than against live site. I'll see if I can make that work.

But if my speculation about how things work is obviously off-base, please advise. :smile:

mlinksva:
That seems to do it, and if I introduce a http://choosealicense.com/... link that doesn't exist locally, the test fails, as it should.

benbalter:
~~Tests pass for me locally without the change in 96fc8b1. Theoretically, there shouldn't be any choosealicense.com links locally, since everything uses site.url or site.github.url. Do you know where the link is in the page?~~

benbalter:
Just kidding. I was using your fix. I think I see what's going on.

mlinksva:
`<link rel="canonical" href="http://choosealicense.com/licenses/eupl-1.1/" itemprop="url" />`

benbalter:
Yep. It's via the SEO tag plugin. Looks like your fix is the simplest. :+1:

PR #329, opened 2016-01-28
skipping no-license no longer pertinent as no-license is no-more

mlinksva:
Result of #326


mlinksva:
Merging this now mainly to trigger another pages rebuild...

PR #330, opened 2016-01-29
Test each license file does not begin with byte order mark

mlinksva:
My Ruby is probably extremely bad. For one thing I couldn't get the first 3 bytes read to equal "\xEF\xBB\xBF" (with any operator) and I couldn't discern what the difference was in irb. So I resorted to comparing them as byte arrays, which works fine.

Not sure this test is even worth including, but critique of any sort welcome in any case.

### DIFF
--- a/spec/license_bom_spec.rb
+++ b/spec/license_bom_spec.rb
@@ -1,0 +1,16 @@
+require 'spec_helper'
+
+describe "license file" do
+
+  bom = "\xEF\xBB\xBF".each_byte.to_a
+
+  files = Dir["_licenses/*.txt"]
+
+  files.each do |fn|
+    first3 = File.open(fn) {|f| f.read(3)}
+
+    it "#{fn} should not have a byte order mark" do
+      expect(first3.each_byte.to_a).not_to eq(bom)
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


benbalter:
Pushed up a small change in 41a9f1a. Feel free to revert if you disagree. While there was nothing wrong with your approach, I did three things:
1. Look up the licenses directory by absolute path. I don't know that we could assume the pwd on Travis or on the user's system. People do silly things.
2. Use regex to check if the file starts with `\xEF\xBB\xBF`. I tested by adding a BOM to a licenses and confirming it properly erred.
3. Provide the contributor with instructions on how to remove the BOM. While the test isn't the most Ruby way to do things (looking for a value to be false, rather than the real test), it gives the contributor a clear next step:

```
  2) byte order marks the mit license does not begin with a byte order mark
     Failure/Error: expect(bom).to eql(false), msg
       License file begins with a Byte Order Mark. See http://stackoverflow.com/a/1068700.
     # ./spec/license_bom_spec.rb:9:in `block (4 levels) in <top (required)>'
```

Last, I realized Jekyll can't parse files that begin with a BOM, so added a quick test to ensure the number of parsed licenses match the number of files in the `_licenses` folder, which could be a sign that something's amiss.

mlinksva:
I very much agree, those are good changes, thanks!

PR #333, opened 2016-01-30
Artistic License 2.0 (c) is for license text, not a template

### LINKED ISSUE #310: Should the copyright notice on the artistic license refer to the license itself or the project it licenses?
In https://github.com/github/choosealicense.com/blob/gh-pages/_licenses/artistic-2.0.txt#L29, the 2nd line of the license, replaces [the standard](https://opensource.org/licenses/Artistic-2.0):

> Copyright (c) 2000-2006, The Perl Foundation.

with:

>  Copyright (c) [year] [fullname]

(Which gets mail-merged when programmatically added to a project)

In every other license I know of, a copyright notice on the first or second line is intended to refer to the project as a whole, not the license. When copyright notices for the license itself are included, e.g., https://github.com/github/choosealicense.com/blob/gh-pages/_licenses/afl-3.0.txt#L69, it's explicit that it's for the license, not the project.

@mlinksva Do you happen to know? If not, do you know someone from the perl foundation which we can ask?

mlinksva:
fix #310 introduced in 0161a3d75d5f72fe842886b79c07db6ceb773ae2

### DIFF
--- a/_licenses/artistic-2.0.txt
+++ b/_licenses/artistic-2.0.txt
@@ -27,7 +27,7 @@ hidden: false
 
                The Artistic License 2.0
 
-           Copyright (c) [year] [fullname]
+           Copyright (c) 2000-2006, The Perl Foundation.
 
      Everyone is permitted to copy and distribute verbatim copies
       of this license document, but changing it is not allowed.


bkeepers:
:+1:

mlinksva:
I'm going to go ahead and merge this as it's correct and what we have now isn't. I don't know by what scheme we're doing plain text formatting here, so it's possible leading spaces should be removed from the line.

PR #334, opened 2016-01-30
Remove confusing and inaccurate [no-]sublicensing property

mlinksva:
LGPLv3 doesn't allow sublicensing, justg as LGPLv2 doesn't as
former is additional permissions for GPLv3 and these don't include
sublicensing

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,6 @@ The license properties (rules) are stored as a bulleted list within the licenses
 * `commercial-use` - This software and derivatives may be used for commercial purposes.
 * `modifications` - This software may be modified.
 * `distribution` - You may distribute this software.
-* `sublicense` - You may grant a sublicense to modify and distribute this software to third parties not included in the license.
 * `private-use` - You may use and modify the software without distributing it.
 * `patent-use` - This license provides an express grant of patent rights from the contributor to the recipient.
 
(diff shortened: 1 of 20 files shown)


bkeepers:
This was removed in #303 due to discussion in #10. My understanding is that sublicensing is technically forbidden, but the copyleft makes the ability to sublicense irrelevant.

/cc @mebmink

mlinksva:
Well #303 was an incomplete implementation as there's inconsistency across copyleft licenses represented at choosealicense.com presently.

#10 is muddled as discussion of sublicensing usually is. It's definitely not about conveying the work. It's the ability of a licensee to license the rights it has gained through the license to its licensees. That's not needed for open source licenses as you're always getting a direct license from each contributor.

I wouldn't mind much just removing sublicensing / no-sublicensing properties altogether from the site as it is a confusing concept and not one people choosing a license need to understand.

But if we're going to keep the property it's pretty silly to present it inaccurately.

mlinksva:
(Thanks for the pointer to past discussion BTW; I should've searched first.)

mlinksva:
Noticed additional inaccuracies. Convinced myself best to remove description of sublicensing from the site. Even if made accurate would be confusing and not helpful in choosing a license. Noise at best, more realistically anti-information.

benbalter:
@mlinksva defer to your judgement here.

PR #336, opened 2016-01-31
Update example projects

bkeepers:
This updates the example projects listed on the homepage for each license.
- add a 3rd example (.NET Core) to MIT for consistency
- replace SVN & NuGet with more more modern and well-known examples
  (Swift, Docker) for Apache
- Add a GPLv3 project to examples

There are not currently guidelines for how we choose the projects. We (@amateurhuman and I) tried to choose projects that:
- are more likely to be known by people in most ecosystems (sorry NuGet)
- represent different types of projects, like programming languages, libraries, and utilities

### DIFF
--- a/index.html
+++ b/index.html
@@ -42,7 +42,7 @@ permalink: /
       <h3>I care about sharing improvements.</h3>
     </a>
     <p>
-      The <a href="licenses/gpl-3.0/">GNU GPL</a> is a copyleft license that lets people do anything they want with your code as long as they give the same freedoms to anyone   else to which give your code or something based on it.
+      The GPL (<a href="licenses/gpl-2.0/">V2</a> or <a href="licenses/gpl-3.0/">V3</a>) is a copyleft license that requires anyone who distributes your code or a derivative work to make the source available under the same terms. V3 is similar to V2, but further restricts use in hardware that forbids software alterations.
     </p>
     <p>
       <strong>WordPress</strong>, <strong>Linux</strong>, and <strong>Bash</strong> use the GPL.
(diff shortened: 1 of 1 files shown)


mlinksva:
Seems an improvement to me. I'd suggest Android rather than Docker for Apache, just because former is a mass consumer brand.

I was eventually going to suggest removing the examples altogether. I don't see how they help a newcomer understand the licenses. Only people who follow projects closely have any idea of the reasons why projects use particular licenses, and often the reasons are highly idiosyncratic anyway.

benbalter:
Nice. :+1:

index.html
```diff
@@ -42,10 +42,10 @@
       <h3>I care about sharing improvements.</h3>
     </a>
     <p>
-      The GPL (<a href="licenses/gpl-2.0/">V2</a> or <a href="licenses/gpl-3.0/">V3</a>) is a copyleft license that requires anyone who distributes your code or a derivative work to make the source available under the same terms. V3 is similar to V2, but further restricts use in hardware that forbids software alterations.
+      The <a href="licenses/gpl-3.0/">GNU GPL</a> is a copyleft license that lets people do anything they want with your code as long as they give the same freedoms to anyone   else to which give your code or something based on it.
```
waldyrious: There's a triple space in "anyone   else"

bkeepers: Ooops, I didn't mean to commit this change yet. I'll revert and open another PR with suggested wording changes for GPL.

amateurhuman:
> I was eventually going to suggest removing the examples altogether. I don't see how they help a newcomer understand the licenses.

I think they're helpful for newcomers to be able to relate to their community. If I'm working on an open source developer tool, I might want to consider using the same license as something like Docker since that may be what my community is comfortable with. I'd say Android developers likely want to release code under the same license as Android (some do it for complex technical reasons around license compatibility, others do it because they want to fit with the community). Either way, I think they guide users in a helpful way.

PR #343, opened 2016-02-03
Remove 'library-usage'

wking:
The last meaningful change to this tag was c4c48d49 (Change nonstatic
to library usage, 2013-07-10), but I'm not sure where that discussion
happened.  In any case, that commit changed some "must" wording to
"may" wording, which seems like it should move the label from required
to permitted.

The 'usage' -> 'use' change is just for consistency with the other
labels about using a package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,6 @@ The license properties (rules) are stored as a bulleted list within the licenses
 * `document-changes` - Indicate significant changes made to the code.
 * `disclose-source` - Source code must be made available when distributing the software. In the case of LGPL and OSL 3.0, the source for the library (and not the entire program) must be made available.
 * `network-use-disclose` - Users who interact with the software via network are given the right to receive a copy of the corresponding source code.
-* `library-usage` - The library may be used within a non-open-source application.
 * `rename` - You must change the name of the software if you modify it.
 
 #### Permitted
--- a/_data/rules.yml
+++ b/_data/rules.yml
@@ -11,9 +11,6 @@ required:
 - description: Users who interact with the software via network are given the right to receive a copy of the corresponding source code.
   label: Network Use is Distribution
   tag: network-use-disclose
-- description: The library may be used within a non-open-source application.
-  label: Library usage
-  tag: library-usage
 - description: You must change the name of the software if you modify it.
   label: Rename
   tag: rename
--- a/_licenses/lgpl-2.1.html
+++ b/_licenses/lgpl-2.1.html
@@ -15,7 +15,6 @@ note: The Free Software Foundation recommends taking the additional step of addi
 
 required:
   - include-copyright
-  - library-usage
   - disclose-source
 
 permitted:
--- a/_licenses/lgpl-3.0.html
+++ b/_licenses/lgpl-3.0.html
@@ -14,7 +14,6 @@ note: The Free Software Foundation recommends taking the additional step of addi
 
 required:
   - include-copyright
-  - library-usage
   - disclose-source
 
 permitted:


mlinksva:
This is an improvement over what we've got, but stated this way it also applies to all non-hidden licenses on the site except for GPL and AGPL. One difference between LGPL and those licenses is already noted in the 'disclose source' tooltip.

Curiously the aspect that 'library-use/age' is trying to get at isn't reflected in any property -- modifications must be under the same license. Seems a property indicating that ought be added to all copyleft licenses, with library caveat also noted in a tooltip.

But for this change, I'd suggest removing library-use/age entirely rather than propagating to most of the licenses on the site.

wking:
On Wed, Feb 03, 2016 at 03:37:44PM -0800, Mike Linksvayer wrote:

> Curiously the aspect that 'library-use/age' is trying to get at
> isn't reflected in any property -- modifications must be under the
> same license. Seems a property indicating that ought be added to all
> copyleft licenses, with library caveat also noted in a tooltip.
> 
> But for this change, I'd suggest removing library-use/age entirely
> rather than propagating to most of the licenses on the site.

Good points.  Relaxing disclose-source for linking software is hard to
fit into this framework.  Maybe we need a
disclose-source-linking-exception label?

mlinksva:
Disclose-source relaxation for LGPL is already noted in tooltip.

I suspect we need a totally different property noting requirement for modifications to be under same license. `share-alike` would be one possibility. This could also be relaxed for LGPL via tooltip.

Getting relaxation for LGPL out of tooltip requires either an odd property (like the one this PR initially modifies) or properties that can have non-binary values, which might be a good thing to do longer term but is a bigger change.

wking:
On Wed, Feb 03, 2016 at 04:00:42PM -0800, Mike Linksvayer wrote:

> Disclose-source relaxation for LGPL is already noted in tooltip.

It's currently just a blanket tooltip, so it's extra noise for
copyleft licenses without a linking exception.  The OSL license is
also currently unlinked [1,2], so removing the OSL reference from
the tooltip would be nice.

> I suspect we need a totally different property noting requirement
> for modifications to be under same license. `share-alike` would be
> one possibility.

Sounds good to me, but that seems orthogonal to the library-usage
change.

> Getting relaxation for LGPL out of tooltip requires either an odd
> property (like the one this PR initially modifies)…

disclose-source-linking-exception (replacing disclose-source) for the
LGPL and OSL seems like it would let us drop the blanket tooltip
caveat in favor of more specific requirements for those licenses,
without requiring the large changes [3](https://github.com/github/choosealicense.com/pull/343#issuecomment-179532710) to other permissive-linking
licenses.  Can you take another stab at explaining why
disclose-source-linking-exception is a bad idea?

Rosen might take issue with the term “linking” in the label [4](http://rosenlaw.com/OSL3.0-explained.htm#_Toc187293087).  I
like his reasoning, but I hear “linking exception” mentioned
frequently and am not sure what generic collective-work term could be
used instead.

> … or properties that can have non-binary values, which might be a
> good thing to do longer term but is a bigger change.

One of the benefits of this site is that it glosses over a lot of
these technical distinctions.  Having sub-label knobs sounds like it
would distract from the high-level overview, unless you only envision
it controlling tooltip information.

wking:
On Wed, Feb 03, 2016 at 04:25:08PM -0800, W. Trevor King wrote:

> > Getting relaxation for LGPL out of tooltip requires either an odd
> > property (like the one this PR initially modifies)…
> 
> disclose-source-linking-exception (replacing disclose-source) for
> the LGPL and OSL seems like it would let us drop the blanket tooltip
> caveat in favor of more specific requirements for those licenses,
> without requiring the large changes [3] to other permissive-linking
> licenses.  Can you take another stab at explaining why
> disclose-source-linking-exception is a bad idea?

Ah, maybe we're looking for:

required:
…
- description: Other packages that link to this product count as derivative works.
  label: Linking is Derivative Work
  tag: linking-is-derivative

Which would apply to the GPL and similar, but not to the LGPL and
similar.  It's not as clear about what restrictions that imposes on
the derivative work, leaving that to other properties such as
disclose-source, etc.

mlinksva:
Thanks for explaining patiently explaining. I'm just noticing yet another problem: EPL and MPL also currently have the disclose-source property, and the disclosure requirement is in MPL case even narrower than LGPL's.

That's fine though: disclose-source is accurate for all of these (and GPL/AGPL). But the tooltip as it currently stands is very odd.

linking-is-derivative reads like a definition not a requirement, but I suppose that we already have similar in network-use-disclose. But I don't think linking-is-derivative is quite the right property to separate GPL and LGPL -- LGPL doesn't say linking is not derivative in general.

So far I think
- library-usage property should be removed entirely
- LGPL/OSL mention in disclose-source tooltip should be removed
- some yet to be named property should be added to AGPL and GPL; linking-is-derivative is close.

wking:
On Wed, Feb 03, 2016 at 04:59:04PM -0800, Mike Linksvayer wrote:

> linking-is-derivative reads like a definition not a requirement…
> …
> LGPL doesn't say linking is not derivative in general.

In fact, v2.1 explicitly says that linking (static or dynamic) _is_
derivative, and just that that sort of derivative work has relaxed
requirements [1](https://www.gnu.org/licenses/old-licenses/lgpl-2.1.html):

  When a program is linked with a library, whether statically or using
  a shared library, the combination of the two is legally speaking a
  combined work, a derivative of the original library.

I think that hints at a useful distinction between “derivative” and
“combined”.  The OSL aims for a similar target by tightly scoping
“derivative” [2](http://rosenlaw.com/OSL3.0-explained.htm#_Toc187293087), and the LGPL v3.0 language talks about “Combined
Work” and not about “Derivative Work” [3](https://www.gnu.org/licenses/lgpl.html).  How about picking up the
“combined” language and using ‘combination-share-alike’ or
‘combination-restrictions’?

> - library-usage property should be removed entirely
> - LGPL/OSL mention in disclose-source tooltip should be removed
> - some yet to be named property should be added to AGPL and GPL;
>   linking-is-derivative is close.

This all sounds good to me.

mlinksva:
I'm not sure a property concerning combined work is much more useful than one concerning derivative to distinguish AGPL/GPL from weaker copylefts, or at least LGPL. For the latter, whether the term is derivative or combination, there's no one-word summary for when and what license conditions apply that I can think of.

I wonder if spectrum of copyleft strength from MPL through EPL and LGPL to GPL best left to textual descriptions on http://choosealicense.com/licenses/ (which themselves may need tweaking) and left out of binary properties?

wking:
On Thu, Feb 04, 2016 at 10:38:37AM -0800, Mike Linksvayer wrote:

> I'm not sure a property concerning combined work is much more useful
> than one concerning derivative to distinguish AGPL/GPL from weaker
> copylefts, or at least LGPL. For the latter, whether the term is
> derivative or combination, there's no one-word summary for when and
> what license conditions apply that I can think of.

They all seem to pick different wording for the concept, but they seem
to put the distinction in a similar place and put similar restrictions
on the combined work.  I'd like to use a binary label if possible, but
I'm not wedded to the idea ;).

> I wonder if spectrum of copyleft strength from MPL through EPL and
> LGPL to GPL best left to textual descriptions on
> http://choosealicense.com/licenses/ (which themselves may need
> tweaking) and left out of binary properties?

Can you file a PR with your recommended wording?  Maybe that will help
clarify the distinctions that make a binary label untenable.

I've pushed f8a23aa → ffa5387 adjusting this PR to remove
library-usage.  The commit message explains why, and also tries to
explain why it can land before a subsequent commit replacing the
disclose-source description caveat with either license descriptions or
a new label.

mlinksva:
Thanks! I'll merge this now and create a new PR referring to this discussion later.

Issue #335, opened 2016-01-31
Feedback from John Sullivan talk on license choosers


bkeepers:
John Sullivan from FSF gave a great talk at FOSDEM on license choosers. He had some helpful observations and recommendations for choosealicense.com that I've attempted to summarize here:
- Make the agenda explicit. Any attempt to simplify and summarize options will have implicit biases, so at least name them.
- MIT
  - Being the first one listed puts it in a (potentially) privileged position
  - The heading of being "Simple & Permissive" could be seen as more appealing
  - "short and to the point" is not normative
  - 2 examples listed instead of 3
- Apache:
  - description is not as compelling as MIT
  - "I’m concerned about patents” is not a distinguishing feature from GPL, creating a false dichotomy
  - description builds on MIT
  - Patent text is unclear
- GPL
  - fix name: _GNU_ GPL
  - none of the examples are GPL v3
  - “further restricts” is slanted phrase
  - “hardware that forbids software alterations” is not entirely correct
  - Recommended change: “It lets people do anything they want with your code as long as, if they give your code or something based on it to others, they give those people the same 

Anyone else that saw the talk: please correct or add to these points.

bkeepers:
> Make the agenda explicit. Any attempt to simplify and summarize options will have implicit biases, so at least name them.

The [README](https://github.com/github/choosealicense.com#immediate-goals) lists the immediate goals, but it may be worth further clarifying. In a comment at the end of the session, I described our agenda as (in this order):
1. Get project maintainers to choose a license (which means simplifying as much as necessary)
2. Reflect consensus of community and license authors as accurately as possible

mlinksva:
Seems like reasonable feedback. I can begin to implement, or rather continue to implement, once #336 which addresses the lack of GPLv3 examples, is merged.

As I've [mentioned elsewhere](https://github.com/github/choosealicense.com/issues/331#issuecomment-176884853), I'd rather not have GPLv2 at all on the home page. That would solve needing to try to differentiate between v2 and v3 on the home page, which is hard to get correct and way too much detail for 99%. Wonder if John mentioned any preference there?

mlinksva:
[Abstract](https://fosdem.org/2016/schedule/event/license_pickers/), [video](http://video.fosdem.org/2016/ud2218a/pick-a-peck-of-license-pickers.mp4)

I'm mildly surprised http://home.ccil.org/~cowan/floss/ wasn't mentioned. It's the best.

[...]

PR #344, opened 2016-02-05
Copyleft tweaks

mlinksva:
-     remove note about LGPL and OSL
-     make same-license requirement, add to all copyleft licenses
-    ~~Note differences in copyleft scope in description field. Use suggestion from John Sullivan in #335 as base description language.~~

Based on discussion with @wking in #343

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,7 @@ The license properties (rules) are stored as a bulleted list within the licenses
 * `disclose-source` - Source code must be made available when distributing the software.
 * `network-use-disclose` - Users who interact with the software via network are given the right to receive a copy of the corresponding source code.
 * `rename` - You must change the name of the software if you modify it.
+* `same-license` - Modifications must be released under the same license when distributing the software. In some cases a similar or related license may be used.
 
 #### Permitted
 
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ The license properties (rules) are stored as a bulleted list within the licenses
 
 * `include-copyright` - Include a copy of the license and copyright notice with the code.
 * `document-changes` - Indicate significant changes made to the code.
-* `disclose-source` - Source code must be made available when distributing the software. In the case of LGPL and OSL 3.0, the source for the library (and not the entire program) must be made available.
+* `disclose-source` - Source code must be made available when distributing the software.
 * `network-use-disclose` - Users who interact with the software via network are given the right to receive a copy of the corresponding source code.
 * `rename` - You must change the name of the software if you modify it.
 
--- a/_data/rules.yml
+++ b/_data/rules.yml
@@ -14,6 +14,9 @@ required:
 - description: You must change the name of the software if you modify it.
   label: Rename
   tag: rename
+- description: Modifications must be released under the same license when distributing the software. In some cases a similar or related license may be used.
+  label: Same License
+  tag: same-license
 
 permitted:
 - description: This software and derivatives may be used for commercial purposes.
--- a/_data/rules.yml
+++ b/_data/rules.yml
@@ -5,7 +5,7 @@ required:
 - description: Indicate significant changes made to the code.
   label: State Changes
   tag: document-changes
-- description: Source code must be made available when distributing the software. In the case of LGPL and OSL 3.0, the source for the library (and not the entire program) must be made available.
+- description: Source code must be made available when distributing the software.
   label: Disclose Source
   tag: disclose-source
 - description: Users who interact with the software via network are given the right to receive a copy of the corresponding source code.
--- a/_licenses/agpl-3.0.html
+++ b/_licenses/agpl-3.0.html
@@ -18,6 +18,7 @@ required:
   - document-changes
   - disclose-source
   - network-use-disclose
+  - same-license
 
 permitted:
   - commercial-use
--- a/_licenses/epl-1.0.html
+++ b/_licenses/epl-1.0.html
@@ -18,6 +18,7 @@ using:
 required:
   - disclose-source
   - include-copyright
+  - same-license
 
 permitted:
   - commercial-use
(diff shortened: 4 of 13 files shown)


bkeepers:
I like the spirit of the wording of John's suggestion from #335, but I find the wording somewhat awkward…

> Lets people do anything they want with your code as long as, if they give your code or something based on it to others, they give those people the same freedoms.

This isn't any better, but here's an iteration on it…

> Lets people do anything they want with your code as long as they give the same freedoms to anyone whom they give your code or something based on it.

/cc @amateurhuman because I know you had some feels about this when we were chatting.

wking:
On Fri, Feb 05, 2016 at 11:55:43AM -0800, Brandon Keepers wrote:

> > Lets people do anything they want with your code as long as they
> > give the same freedoms to anyone whom they give your code or
> > something based on it.

“Lets people do anything they want with your code as long as …” is a
lot of text for “I'm about to list a license requirement…”.  How
about:
—
  Licensees who distribute your code or something based on it must
  also release their code under the same (or similar) terms.

mlinksva:
@wking "Lets people..." is for emphasizing permission (or freedom if you prefer) not merely a garrulous way of saying "I'm about to list a license requirement".

That said I don't particularly care what form the descriptions take, but would like more consistency than what we have now.

wking:
On Fri, Feb 05, 2016 at 12:46:05PM -0800, Mike Linksvayer wrote:

> @wking "Lets people..." is for emphasizing permission (or freedom if you prefer)…

The freedom promoted by copyleft licenses is the freedom for
grandchild licensees and later, not the freedom of direct licensees.
I don't think emphasizing the direct licensee's freedom is useful, but
if someone has text for emphasizing indirect licensee freedom, I'm all
for it.

> That said I don't particularly care what form the descriptions take,
> but would like more consistency than what we have now.

Yeah, +1 for consistency.  It makes it easier to spot the actual
differences.

mlinksva:
Copyleft licenses like all free/open source licenses promote freedom of direct licensees. Their difference is that they also promote (or attempt to anyway) freedom for people who receive software from additional licensors.

But I take your point that descriptions should make seeing differences easy, and a long text doesn't do that. I'll make another attempt adding to this PR later. Thanks!

wking:
On Fri, Feb 05, 2016 at 01:13:34PM -0800, Mike Linksvayer wrote:

> Copyleft licenses like all free/open source licenses promote freedom
> of direct licensees.

Yeah, and that gets covered very explicitly by private-use (especially
since #115).  The thing that sets $LICENSE apart from other free/open
licenses is what the per-license description should focus on.

mlinksva:
Removed commit that updated descriptions. I don't think it makes sense to limit updating descriptions to copyleft licenses only. Will attempt general description update in another PR.

wking:
On Fri, Feb 05, 2016 at 02:37:49PM -0800, Mike Linksvayer wrote:

> Removed commit that updated descriptions.

2a01884f6 and previous look good to me :).

wking:
On Fri, Feb 05, 2016 at 02:44:55PM -0800, W. Trevor King wrote:

> Fri, Feb 05, 2016 at 02:37:49PM -0800, Mike Linksvayer:
> 
> > Removed commit that updated descriptions.
> 
> 2a01884f6 and previous look good to me :).

Actually, 2a01884f6 by itself looks good to me.  5fb1cde7 (qualifying
which works same-license applies to) probably should get shifted to a
same-license caveat.  I don't think we can drop an LGPL / OSL /
etc. caveat of some sort without adjusting per-license descriptions.

wking:
disclose-source and same-license are strongly correlated.  As of
2a01884f6, the only license with disclose-source but not same-license
is the LPPL, and the only licenses with same-license but not
disclose-source are the SIL OFL and the MS-RL.  Is this accidental?

The SIL OFL note makes same-license but no disclose-source look
intentional.  Should we drop the same-license reference in the note
when adding the same-license label?

It looks like the MS-RL §3.a contains a disclose-source requirement.
It looks like it has similar weak same-license scoping similar to the
LGPL and OSL.

The LPPL's §6 and §10 look like they require same-license and
disclose-source (with some link/compilation caveats) for the Work, but
neither (with some license-compatibility restrictions) for Derived
Works.  Maybe we should add language to same-license and
disclose-source trying to qualify whether they apply to the “just the
Work”, “the Work and any Derived Works”, or “the Work, and possibly to
Derived Works”?

mlinksva:
>  I don't think we can drop an LGPL / OSL /
> etc. caveat of some sort without adjusting per-license descriptions

I think we can: the differences are mentioned in per-license descriptions, just not with a great deal of consistency. And I think the extra tooltip text is a net negative as it stands: it's misleading for MPL and EPL, and mentioning the OSL is weird as it is otherwise hidden.

> disclose-source and same-license are strongly correlated.  As of
> 2a01884, the only license with disclose-source but not same-license
> is the LPPL, and the only licenses with same-license but not
> disclose-source are the SIL OFL and the MS-RL.  Is this accidental?

It's not accidental. They are highly correlated in practice but not theory. I'm not 100% sure `disclose-source: true` is correct for LPPL, but that combination is right for OFL and MS-RL, and [CC-BY-SA-4.0 when it is added](https://github.com/github/choosealicense.com/issues/33#issuecomment-173067929) will be the same (false, true).

wking:
On Sat, Feb 06, 2016 at 10:49:37AM -0800, Mike Linksvayer wrote:

> >  I don't think we can drop an LGPL / OSL / etc. caveat of some
> > sort without adjusting per-license descriptions
> 
> I think we can: the differences are mentioned in per-license
> descriptions…

Ah, you're right.  So I'm back to +1 on both the tooltip trimming from
5fb1cde and the same-license addition from 2a01884.

mlinksva:
Thanks for your attention and eye for detail @wking, really needed for a project like this.

Tangentially I notice you're involved in [Software Carpentry](https://github.com/swcarpentry), one of the projects/organizations I most admire. Thanks again for all your contributions. :smile:

PR #345, opened 2016-02-08
Test that only open licenses are included per CONTRIBUTING

### LINKED ISSUE #1: MIT and BSD also allow private use

mlinksva:
https://github.com/github/choosealicense.com/blob/gh-pages/CONTRIBUTING.md#adding-a-license
- check that all licenses have minimum permissions
- remove non-open (and unused) forbiddens
- closes #1 with confidence

### DIFF
--- a/spec/license_spec.rb
+++ b/spec/license_spec.rb
@@ -29,16 +29,19 @@ describe "licenses" do
       end
 
       context "minimum permissions" do
-        permissions = license["permitted"]
+        let(:permissions) { license["permitted"] }
         it "should allow commercial use" do
           expect(permissions).to include("commercial-use")
         end
+
         it "should allow modification" do
           expect(permissions).to include("modifications")
         end
+
         it "should allow distribution" do
           expect(permissions).to include("distribution")
         end
+
         it "should allow private use" do
           expect(permissions).to include("private-use")
         end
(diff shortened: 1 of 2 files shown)


spec/license_spec.rb
```diff
@@ -27,6 +27,22 @@
           expect(approved_licenses).to include(id), "See https://git.io/vzCTV."
         end
       end
+
+      context "minimum permissions" do
+        permissions = license["permitted"]
```
benbalter: I'll be honest, I don't fully get it, but stylistically, for tests, this should be `let(:permissions) { license["permitted"] }`, similar to the `id` line, a few lines above. It's a Ruby/Rspec thing and has something to do with scopes or cleanup, IIRC.

spec/license_spec.rb
```diff
@@ -27,6 +27,22 @@
...
+        it "should allow commercial use" do
+          expect(permissions).to include("commercial-use")
+        end
+        it "should allow modification" do
```
benbalter, line 37: Technically, technically, Ruby style says the inner blocks should have whitepace around them (e.g., lines 36, 39, 42), like https://github.com/github/choosealicense.com/blob/gh-pages/spec/license_category_spec.rb#L21-L28.

benbalter:
Really good idea. Two minor style notes, not to be too pedantic, but :+1:.

PR #346, opened 2016-02-08
Add Rubocop

benbalter:
As they say at GitHub:

> Pedantic Robots > Pedantic Humans

Inspired by trying to be helpful over in https://github.com/github/choosealicense.com/pull/345, this pull requests adds Rubocop, to enforce Ruby style.

It's relatively straightforward because it's mostly just tests and ad-hoc scripts, but I find those are the most likely to be overlooked, and with multiple maintainers, it'd be best to clean things up.

As an added benefit, this fixes some not-really-global global variables that were used for caching, meaning tests are now _significantly_ faster.

I went with the default tests, and disabled a few that'd be too onerous to enforce. There should be no change to the resulting site or test logic. It should all be stylistic.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,16 @@
+AllCops:
+  TargetRubyVersion: 2.1
+  Exclude:
+    - _site/**/*
+
+Metrics/LineLength:
+  Enabled: false
+
+Metrics/MethodLength:
+  Enabled: false
+
+Metrics/AbcSize:
+  Enabled: false
+
+Style/Documentation:
+  Enabled: false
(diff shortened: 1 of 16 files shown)


mlinksva:
I appreciate pedantic feedback from you and will also from a bot.

Works for me locally, not sure why run by CI its seeming to check all gems used by the project.

benbalter:
> I appreciate pedantic feedback from you and will also from a bot.

Ha. Awesome. Equally so for my own sake, as looking back at some of my old "ehhh, lets just make this work for now" code, was not pretty, and I wouldn't want to subject anyone else to having to deal with it. :smile:

> Works for me locally, not sure why run by CI its seeming to check all gems used by the project.

Travis vendored vendored gems into the vendor folder. Excluded it via the rubocop config, and tests are now green.

Issue #290, opened 2015-09-08
relationship between "tab-slug" and "redirect_from" properties, and the filename is unclear / inconsistent.


waldyrious:
They should be harmonized / clarified, and possibly merged (i.e. using the filename instead of the tab-slug property, or renaming the files according to the tab-slug, or something else along those lines).

Here's a table of the current values, to allow a better overview:

| filename | tab-slug | redirect_from | category |
| --- | --- | --- | --- |
| agpl-3.0 | agpl-v3 | agpl | gpl |
| apache-2.0 |  | apache |  |
| artistic-2.0 |  | artistic |  |
| bsd-2-clause | bsd | bsd | bsd |
| bsd-3-clause | bsd-3 |  | bsd |
| cc0-1.0 | cc0 | cc0 | pd* |
| epl-1.0 |  | eclipse |  |
| gpl-2.0 | gpl-v2 | gpl-v2 | gpl |
| gpl-3.0 | gpl-v3 | gpl-v3 | gpl |
| isc | isc |  | bsd |
| lgpl-2.1 | lgpl-v2_1 | lgpl-v2.1 | lgpl |
| lgpl-3.0 | lgpl-v3 | lgpl-v3 | lgpl |
| mpl-2.0 |  | mozilla |  |
| ofl-1.1 |  | ofl |  |
| unlicense | unlicense |  | pd* |

\* - shortened from "Public Domain Dedication" to make the table shorter. See #289 for a proposal to change this.

---

Looking at the "name" and nickname properties, one can also notice a trend:

Additionally, it seems that where the nickname is the same as the tab-slug or redirect_from, it is not included, the exception being the Apache license. So either the nickname should be removed in that case, or added to all licenses regardless of coinciding with the tab-slug/redirect_from version.

| nickname | title |
| --- | --- |
| GNU Affero GPL v3.0 | GNU Affero General Public License v3.0 |
| Apache | Apache License 2.0 |
|  | Artistic License 2.0 |
| Simplified BSD | BSD 2-clause "Simplified" License |
| New BSD | BSD 3-clause "New" or "Revised" License |
| CC0 1.0 Universal | Creative Commons Zero v1.0 Universal |
|  | Eclipse Public License 1.0 |
| GNU GPL v2.0 | GNU General Public License v2.0 |
| GNU GPL v3.0 | GNU General Public License v3.0 |
|  | ISC License |
| GNU LGPL v2.1 | GNU Lesser General Public License v2.1 |
[...]

PR #348, opened 2016-02-12
Remove need for tab-slug

mlinksva:
Partially address #290 following up on @bkeepers suggestion to use liquid filters https://github.com/github/choosealicense.com/issues/290#issuecomment-139019502

### DIFF
--- a/_includes/license-overview.html
+++ b/_includes/license-overview.html
@@ -24,7 +24,7 @@
       <ul class="nav-pills js-nav-pills">
         {% assign variations = site.licenses | where:"hidden",false | where:"family",license.family %}
         {% for variation in variations %}
-        <li{% if license.id == variation.id %} class="active"{% endif %}><a href="#{{ variation.id | remove: '.' | remove: '/licenses/' }}" data-selected-tab="variation-{{ variation.id | remove: '.' | remove: '/licenses/' }}">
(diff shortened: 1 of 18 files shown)


_includes/license-overview.html
```diff
@@ -43,7 +43,7 @@ <h3 class="license-family-name">
     {% endif %}
 
     {% for variation in variations %}
-    <div {% if variation.tab-slug %}id="{{ variation.tab-slug }}" {% endif %}class="variation-tab {% if variation.tab-slug %}variation-{{ variation.tab-slug }}{% endif %} js-variation-tab{% if license.tab-slug == variation.tab-slug %} active{% endif %}">
+    <div {% if variation.id %}id="{{ variation.id | remove: '.' | remove: '/licenses/' }}" {% endif %}class="variation-tab {% if variation.id %}variation-{{ variation.id | remove: '.' | remove: '/licenses/' }}{% endif %} js-variation-tab{% if license.id == variation.id %} active{% endif %}">
```
benbalter: I believe `{{ variation.id | remove: '/licenses/' | slugify }}` should achieve roughly the same ends by replacing all non-alphanumeric characters (and replace `.`s with `-`s to somewhat preserve existing behavior).

benbalter:
Made a minor implementation selection that can rely a bit more on Jekyll's internals, but otherwise, nice work. Always glad to see cruft removed.

PR #350, opened 2016-02-18
Make BSD nicknames self-descriptive and conform to OSI site use

mlinksva:
- don't require knowledge of 'new' or 'simplified' relative to _what_
- rm nickname for BSD 3-clause Clear License which is hidden and
  deserves to remain little known


benbalter:
> don't require knowledge of 'new' or 'simplified' relative to what... deserves to remain little known

Made me LOL. :+1: :smile:

waldyrious:
+:100:! This has always been quite confusing to me.

PR #352, opened 2016-02-20
Add Code of Conduct

mlinksva:
Project doesn't have one, seems like an oversight. Copied
http://contributor-covenant.org/version/1/4/code_of_conduct.txt


bkeepers:
:+1:

PR #358, opened 2016-02-26
Add project links

thewarpaint:


mlinksva:
Interesting idea. Should probably consistently link to one of project home or project repo or LICENSE file rather than varying by project.

I like linking to LICENSE file; as these are well-known projects a generic project link doesn't inform anyone much, but actually seeing how they've deployed the license they use is informative and topical.

I'd bet every project does have a LICENSE (or similarly named with the license) file that can be linked to, eg one that has a project repo link in this PR that could be replaced with https://github.com/dotnet/corefx/blob/master/LICENSE

thewarpaint:
Cool, will change it to point to the actual LICENSE file. Couldn't find links for some projects, that's why I added a project link, but I'll give it another try.

index.html
```diff
@@ -33,7 +33,7 @@
       The <a href="licenses/apache-2.0/">Apache License 2.0</a> is a permissive license similar to the MIT License, but also provides an express grant of patent rights from contributors to users.
     </p>
     <p>
-      <strong>Apache</strong>, <strong>Swift</strong>, and <strong>Android</strong> use the Apache&nbsp;License&nbsp;2.0.
+      <a href="https://httpd.apache.org/">Apache</a>, <a href="https://github.com/apple/swift/blob/master/LICENSE.txt">Swift</a>, and <a href="https://github.com/android/platform_system_core/blob/master/NOTICE">Android</a> use the Apache&nbsp;License&nbsp;2.0.
```
wking: The LICENSE file in the Apache source seems to be [here](http://svn.apache.org/viewvc/httpd/httpd/trunk/LICENSE?view=markup).

index.html
```diff
@@ -45,7 +45,7 @@
       The GNU GPL (<a href="licenses/gpl-2.0/">V2</a> or <a href="licenses/gpl-3.0/">V3</a>) is a copyleft license that requires anyone who distributes your code or a derivative work to make the source available under the same terms. V3 is similar to V2, but further restricts use in hardware that forbids software alterations.
     </p>
     <p>
-      <strong>WordPress</strong>, <strong>Linux</strong>, and <strong>Bash</strong> use the GNU GPL.
+      <a href="https://github.com/WordPress/WordPress/blob/master/license.txt">WordPress</a>, <a href="https://github.com/torvalds/linux/blob/master/COPYING">Linux</a>, and <a href="https://www.gnu.org/software/bash/">Bash</a> use the GNU GPL.
```
wking: I don't see any mentions of “mirror” in the GitHub page, but I'd guess a better canonical link for Linux would be [this](https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/COPYING).

[This](http://git.savannah.gnu.org/cgit/bash.git/tree/COPYING) looks like a good link for Bash.  [Chet doesn't actually develop in public](http://thread.gmane.org/gmane.comp.shells.bash.bugs/13695/focus=13827), but that repo is where things get thrown over the wall.

thewarpaint:
Added direct links to all projects, only [Rails](https://github.com/rails/rails) doesn't have a file. Thanks, @wking!

mlinksva:
Rails has LICENSE files in subprojects (toplevel has been [suggested](https://github.com/rails/rails/pull/21812) and rejected multiple times). I'd suggest https://github.com/rails/rails/blob/master/activerecord/MIT-LICENSE since I think that's the most well known component.

thewarpaint:
Makes perfect sense. Added that too.

mlinksva:
Thanks @thewarpaint. I'll leave this open for feedback for a day, merge over weekend assuming none.

thewarpaint:
Cool, thanks @mlinksva.

Issue #313, opened 2015-12-17
Open hardware licenses and choosealicense.com?


mithro:
The choosealicense.com site is pretty awesome! 

I was wondering with the growth of open hardware (which are increasingly hosted on github) if it would make sense to add another link in the "My content isn’t code." section for hardware projects. I would have sent a pull request but I couldn't figure out were the right place to link too would be, there doesn't seem to be a similar site - the "best" I found so far was http://www.inmojo.com/licenses/ and that isn't exactly clean/simple like this site.

Generally it seems for hardware;
- For "simple+permissive" most people seem go with CC-BY or BSD/MIT license
- For "care about sharing" the 3 major choices seem to be;
  - CC-BY-SA
  - TAPR Open Hardware License (OHL) - http://www.tapr.org/OHL
  - CERN Open Hardware License - http://www.ohwr.org/projects/cernohl/wiki
- I don't know enough to figure out the "I care about patents" bit..

Thoughts?

stapelberg:
+1.

I was about to submit a pull request for the CERN OHL, just so that it can be detected by https://github.com/benbalter/licensee and exposed by the GitHub license API as a consequence :). Would such a PR be accepted?

mlinksva:
By [use](http://www.oshwa.org/oshw-community-survey-2013/#section5) (2013, but license use changes slowly; anyone with data better in recency or other dimension, please share) the major copyleft family options are CC-BY-SA and GPL. Note the most recent versions of those recently were [made compatible](https://creativecommons.org/weblog/2015/10/08/46186) in a limited (one-way) fashion. The CERN and TAPR licenses are incompatible with everything else, including each other. Promoting such fragmentation would be a bad thing to do.

The best options for hardware designs in any category are the same as the best options for software. Perhaps this should be communicated on the site in some manner. This also answers "care about patents" -- Apache 2. CC instruments should not be recommended for hardware or software due to their explicit caring about patents in an anti-sharing direction: explicit non-grant.

I agree it would be nice to be able to detect the CERN OHL and a whole bunch of other licenses, which can be otherwise [hidden](https://github.com/github/choosealicense.com/issues/273). But I suspect a general decision to support (or not) detecting lots more licenses should be made before adding such one at a time.

benbalter:
>  But I suspect a general decision to support (or not) detecting lots more licenses should be made before adding such one at a time.

[...]

Issue #261, opened 2015-03-27
Licenses for documentations


ghost:
Developing a framework or a library implicitly implies to deliver a documentation and many well known frameworks like Qt uses that kind of licenses to cover their documentions.

By going from pages to pages on http://choosealicense.com, i've noticed that it only covers software licenses. I think that i'd be be interesting to also cover documentation licenses like GNU FDL.

Here is a resource i've found on the web that gives a list of available documentation licenses: https://dreamsongs.com/IHE/IHE-50.html

benbalter:
Most licenses explicitly include documentation (e.g., MIT says "and associated documentation files", Apache says "generated documentation" and "documentation source"). Choosealicense is about making it easier to license software (which includes documentation). I believe for most use cases having a separate documentation license may be over-engineering the problem and will be confusing for folks new to licensing.

ghost:
Yes, I've noticed that some licenses like your examples are giving protection to the documentation files, but some like the GNU GPL (any revision) doesn't cover documentation files.

I thought that explicitly covering documentation licenses on http://choosealicense.com could be interesting for developers, not folks ones as you said and also explain them the purpose of that kind of licenses.

strugee:
Perhaps we can find a compromise - e.g., we could link to the GNU FDL in the GPL explanation text. Something along the lines of, "unlike many other licenses, the GPL does not cover documentation. Consider licensing your documentation under the GNU FDL, which licenses documentation in a manner similar to how the GPL licenses software."

ghost:
Yep, it sounds fine to me. No need to create new pages dedicated to documentation licenses but the information is given to the developer.

mlinksva:
Addressed with #359

PR #359, opened 2016-02-26
non-software licenses page

mlinksva:
building on https://github.com/github/choosealicense.com/pull/307

non-software rather than non-code because software is eating the
    world, the main topic is open source software, and I find
    software/non-software slighly less trying than code/non-code if
    only because it isn't a
    [thing](http://www.adamhyde.net/perhaps-stop-using-non-code-whatayarekin/).

tweak text of page, also addressing #261 and #313


non-software.md
```diff
@@ -0,0 +1,15 @@
...
+title: Non-Software Licenses
+layout: default
+permalink: /non-software/
+---
+
+Open source software licenses can be used for non-software works, and often are the best choice. This is particularly the case when the works in question can be edited and versioned as source, e.g., **[open source hardware](http://www.oshwa.org/definition/) designs**. Choose an [open source software license](/licenses/).
+
+[CC0-1.0](/licenses/cc0-1.0/), [CC-BY-4.0](/licenses/cc-by-4.0/), and [CC-BY-SA-4.0](/licenses/cc-by-sa-4.0/) are [open](http://opendefinition.org) licenses used for **non-software material ranging from datasets to videos**. Note that Creative Commons licenses should [not be used for software](https://creativecommons.org/faq/#can-i-apply-a-creative-commons-license-to-software.).
+
+Any open source software license or open Creative Commons license is applicable to **software documentation**. If you use different licenses for your software and its documentation, you may wish to take care that the licenses are compatible, or that source code examples in the documentation is also licensed under the software license.
```
benbalter: > you may wish to take care that the licenses are compatible, or that source code examples in the documentation is also licensed under the software license.

How do I do that?

non-software.md
```diff
@@ -0,0 +1,15 @@
+---
+title: Non-Software Licenses
+layout: default
+permalink: /non-software/
+---
+
+Open source software licenses can be used for non-software works, and often are the best choice. This is particularly the case when the works in question can be edited and versioned as source, e.g., **[open source hardware](http://www.oshwa.org/definition/) designs**. Choose an [open source software license](/licenses/).
+
+[CC0-1.0](/licenses/cc0-1.0/), [CC-BY-4.0](/licenses/cc-by-4.0/), and [CC-BY-SA-4.0](/licenses/cc-by-sa-4.0/) are [open](http://opendefinition.org) licenses used for **non-software material ranging from datasets to videos**. Note that Creative Commons licenses should [not be used for software](https://creativecommons.org/faq/#can-i-apply-a-creative-commons-license-to-software.).
```
benbalter: I thought CC0 was okay for software?

benbalter:
Nice. I feel like this is throwing a lot of information at the user, and not necessarily _demystifying_ the process. 

Two suggestions:
- Headings and organization to break things into logic groups and make things easier to skim / ignore non-applicable sections. After re-reading things, I think you're doing this with the bold text, but I'd front-load the subject matter / make it explicit.
- Instead of just stating facts, explain _how_ the user could use that fact, or more to the point, what they should do. Instead of "SIL accounts for this", even as simple as "Use a font-specific license, like the SIL". 

I'm torn, as to be helpful, we almost have to be prescriptive, but don't take that tone elsewhere in the site. Perhaps we can make it more like a decision tree?

non-software.md
mlinksva: Heh. I think I'll just remove the bit about compatibility.

non-software.md
mlinksva: I think it's marginal and wouldn't recommend it for a large project due to patent non-grant. But I'll tweak the language.

mlinksva:
@benbalter thanks for the feedback, I'll try to address all in next commit. I was indeed trying to be non-prescriptive.

benbalter:
>  I was indeed trying to be non-prescriptive.

Yeah, I'm torn and am not sure the best answer here. Either way, very glad you're doing this.

MakerTobey:
I am a bit lost in the process. Does this mean that Open Source Hardware Licences, most prominently and widely used the CERN OHL will not become available on GitHub? It would be a highly desired addition in the choosealicence tool by many of the hardware developers active on GitHub.

mlinksva:
@MakerTobey feel free to open an issue or make a PR about adding CERN OHL or any other license. The process is documented at https://github.com/github/choosealicense.com/blob/gh-pages/CONTRIBUTING.md#adding-a-license ... note there are a few criteria for inclusion. If an open hardware license meets all of the criteria but has not been approved by FSF or OSI or OD, we can look into adding an equivalent open hardware license approver to the criteria.

PR #362, opened 2016-03-02
rm gplv2 from homepage

mlinksva:
- remove descrption of v2 v3 difference
- add to description v3 express patent grant
- update example projects to only include v3 ones
- move v2 projects to gplv2 license using property

partially addresses feedback in #335

### DIFF
--- a/_includes/license-overview.html
+++ b/_includes/license-overview.html
@@ -35,7 +35,7 @@
       </ul>
       {% endif %}
     </div>
-    
+
     {% if license.family %}
       {% assign variations = site.licenses | where:"hidden",false | where:"family",license.family %}
     {% else %}
--- a/_includes/using-sentence.html
+++ b/_includes/using-sentence.html
@@ -1,0 +1,17 @@
+{% assign id = "/licenses/" | append: include.license-id %}
+{% assign licenses = site.licenses | where:"id", id %}
+{% for license in licenses %}
+  {% assign using = license.using %}
+  {% if using.size > 0 %}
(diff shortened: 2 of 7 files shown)


wking:
Looks good to me.

Are there criteria for populating the ‘using:’ field?  I'd expect any
consumers listed on the home page (e.g. Bash) to also show up in the
using section on that license's detail page.  To stay DRY, the home
page might be able to pull the first three entries from a license's
‘using:’ automatically, but I'm not familiar enough with Jekyll to
know how to do that off the top of my head (if it's possible at all).

mlinksva:
There aren't AFAIK criteria for the 'using:' field (and it is barely used) but probably linking to license implementation as we discussed in #358 would make sense. Either that or removing 'using:' altogether.

Assuming we keep 'using:' I agree it'd be great for home page to pull from license records. I also don't know how to do that, but quick testing shows it to be possible. There's probably a better way, but I'll make another commit soon with my horrible way.

mlinksva:
DRY but there could well be a better way implemented in ab3ee00 

@benbalter if you have a chance to would love to have your critique.

wking:
On Wed, Mar 02, 2016 at 06:31:34PM -0800, Mike Linksvayer wrote:

> DRY but there could well be a better way implemented in ab3ee00 

I like this.  I'm surprised we need all the {% assign … %} calls, but
I'm not particularly familiar with Jekyll.  Anyhow, DRY is enough for
me; improvements to _includes/using-sentence.html (if possible) can
always land in follow-up commits.

PR #369, opened 2016-03-14
Unmyth

mlinksva:
Don't immediately give user first encountering open source licenses reason to think they are scary. Further in commit messages.

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -4,7 +4,6 @@
           <a href="/terms-of-service/">Terms of Service</a>
         </nav>
         <p>
-          If you have questions or issues, it is always best to consult a legal professional.<br>
           The content of this site is licensed under the <a href="https://creativecommons.org/licenses/by/3.0/">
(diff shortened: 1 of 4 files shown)


_includes/footer.html
```diff
@@ -4,12 +4,11 @@
           <a href="/terms-of-service/">Terms of Service</a>
         </nav>
         <p>
-          If you have questions or issues, it is always best to consult a legal professional.<br>
           The content of this site is licensed under the <a href="https://creativecommons.org/licenses/by/3.0/">
           Creative Commons Attribution 3.0 Unported License</a>.
         </p>
         <div class="with-love">
-          Demystified with &lt;3 by <a href="https://github.com">GitHub, Inc.</a>
+          Made with &lt;3 by <a href="https://github.com">GitHub, Inc.</a>
```
benbalter: Is there a better, more descriptive word than "made"? Published? Curated?

mlinksva: Curated is indeed more descriptive, thanks.

PR #371, opened 2016-03-15
Add name of license to ISC license text

adius:
This makes it immediately clear which license this is without reading and comparing the whole text.

### DIFF
--- a/_licenses/isc.txt
+++ b/_licenses/isc.txt
@@ -22,6 +22,8 @@ limitations:
 hidden: false
 ---
 
+ISC License
+
 Copyright (c) [year], [fullname]
 
 Permission to use, copy, modify, and/or distribute this software for any


mlinksva:
Good idea, and probably is for BSD-2 and 3-clause as well, matching MIT (for which "The MIT License (MIT)" perhaps could be shortened to "MIT License").

I don't know (but there could be) that there's any longstanding different practice among these license to include or exclude the license name; suspect MIT is named here just because people have paid more attention to it.

But I'd like to investigate more and also see what impact adding the title will have on license detection with https://github.com/benbalter/licensee before merging this.

benbalter:
> suspect MIT is named here just because people have paid more attention to it.

See https://github.com/github/choosealicense.com/pull/9. (We added it).

wking:
On Tue, Mar 15, 2016 at 11:28:12AM -0700, Mike Linksvayer wrote:

> Good idea, and probably is for BSD-2 and 3-clause as well, matching
> MIT (for which "The MIT License (MIT)" perhaps could be shortened to
> "MIT License").

My guess is that “The MIT License (MIT)” is a cut-and-paste from [1](https://opensource.org/licenses/MIT).

adius:
@mlinksva "The MIT License (MIT)" should definitely be shortened to "MIT License"
The "The" and "(MIT)" is completely redundant and useless 🙈

The original site for the ISC license also has the title: https://opensource.org/licenses/ISC

mlinksva:
@benbalter someday I will remember to search past issues thoroughly. I see that you also tested the add in licensee. I did the same for this and additional changes in #373 and the new tests pass. So I'm pretty confident nothing would be harmed to go with these changes but I'll wait to give you or others chance to point out something worng.

PR #374, opened 2016-03-18
upgrade to html-proofer 3

mlinksva:
Want to eventually end annoying CI failures on status code 0 'something wrong' for external link checks, eg https://travis-ci.org/github/choosealicense.com/builds/112981705 ... first upgrade to current version of html-proofer so as to debug with that.


benbalter:
:shipit:

PR #379, opened 2016-03-29
Curated with ❤️

cyhsutw:
Replace `<3` with heart Octicons.

screen shot 2016-03-29 at 4 49 19 pm


mlinksva:
Hi, thanks for the pull request.

I'm not thrilled with adding 188 files to the repository and having clients make at least one additional request (I haven't investigated) when a single unicode character would do, if folks would...love seeing a ♡ in the site footer. If we were already using octicons to enhance the site look throughout, that'd be another thing.

Personally, when I see size blowup for tiny or no benefit on sites, I interpret the unspecified love as a love of pinpricking me, the contributor and reader. But, this is just one of my pet peeves.

I'll leave this open for awhile in case anyone else has an opinion.

cyhsutw:
@mlinksva Thanks for pointing that out. Adding whole bunches of icons just for a single :heart: is not worthy. 

I'll close this for now.

Issue #355, opened 2016-02-23
Description and nickname on individual license pages


mlinksva:
Individual license pages don't display the license description or (if available) nickname provided in the front matter of files in `_licenses`, e.g., http://choosealicense.com/licenses/eupl-1.1/ and https://github.com/github/choosealicense.com/blob/gh-pages/_licenses/eupl-1.1.txt

This info seems somewhat useful, especially if one gets to an individual license page through a search engine rather than by navigating the site.

These fields are accessible as `page.description` and `page.nickname` and might be added to `_layouts/license.html` or `_includes/sidebar.html`, depending on where they ought be displayed.

I'm terrible at making this look nice. If anyone wishes to tackle this, make a pull request. Alternately, if anyone feels this is a terrible idea, scream.

devGHBot:
Hi @mlinksva,
I think this is a good idea. I attached something I came up with, I tried to keep it minimal and give more space for any future additions if needed.  I saw that the site is using qtip2 so I was going to use it when the user hovered over the link.  Let me know what you think.
![issue-355-include-info_03](https://cloud.githubusercontent.com/assets/8132772/14206260/d845eccc-f7c6-11e5-87fb-730e21ee933d.jpg)

mlinksva:
@LeoFNaN thanks for coming up with this! Seems reasonable to me for the description. I don't think hover-over to see nickname often shorter than "nickname" is very reasonable. But maybe the nickname could be in the title area of the hover over box, rather than the duplicative "Description"?

waldyrious:
As a user, I would expect the description to appear in a paragraph below the title and before the full text of the license. I can't see any good reasons to use a "hover to see" mechanism when we could just plainly add the text there.

By the way, the nickname could be in the sidebar, before the usage instructions.

mlinksva:
I generally dislike hovering or clicking to see, but in this case I like that it would very crisply separates the summary from the license text.

Of course the description could also go in the sidebar. Maybe that's the easiest thing to do.

devGHBot:
Ok you guys bring up some good points, I was in a make it subtly interactive mode last night :)  So description block in the sidebar, then the nickname and the rest of the content after.

mlinksva:
Would you like to make a PR implementing in the sidebar?

devGHBot:
I'm actually trying to run the repo on my machine but having issues with script/server throwing an error...I wanted to take a look at it before I created a PR.

[...]

PR #382, opened 2016-04-02
Display description and nickname in sidebar (#355)

devGHBot:
Added if for nickname since some of the licenses may not have a nickname.


mlinksva:
I didn't notice before that "how to apply..." is an h5 and the other sidebar headings are h3. I think that might be resulting in a couple pixels less space above "how to apply..." than the other headings. Could you make that one "how to apply..." an h3 also in this PR?

I'll try to merge over the weekend, to give others a chance to look, and myself to clean up any descriptions that need cleaning up before being exposed on the site.

Thanks!

PR #386, opened 2016-04-08
replace /licenses listings with spectrum

### LINKED ISSUE #331: Recommend AGPL3, LGPL3 and GPL3
AGPL3, GPL3 and LGPL3 really differ.

My gut feeling tells me most of the projects on github want either AGPL3 or LGPL3 and possibly not GPL3 (that would be for the tiny fraction that are "desktop" or "mobile" open source binary/packaged/distributed/standalone apps)
- GPL3 enforces share-back on software that is being distributed as a package or binary (App Store, Download, Install, whatever)
- AGPL3 enforces share-back and considers server-only software (PaaS, cloud, webhosting, etc) to be a distribution. This is the case for a big majority of apps on github.
- LGPL3 goes more towards the MIT road of being easy to integrate with closed source infrastructure but still enforces share-back if you want to distribute software, for the library part. E.g. it **is** ~~not~~ **a good** license for a library that is being bundled with other apps while it is a bad license for libraries that you want to have share-back on but that run as part of sever software only (see AGPL3).

E.g. LGPL3 is very much the competition for MIT, while AGPL3 is **MOST LIKELY** the choice for those that **might accidentally chose GPL3**. The current info page totally misses on the last point and hides the competition and differences between LGPL3 and MIT. Both must be fixed in order to respect premise # 1:

> Non-judgmental. Our goal is to help you find a license that meets your goals.

Last but not least, CC is not only for creative stuff. It is very customaziable to go towards public domain (CC0) or non-commercial-shake-alike and things like that.

The current overview is very biased towards the MIT, not only by its primary position in the list, but also because GPL (packaged software) LGPL (packaged software library) and AGPL (server software) is not differentiated, e.g. not communicating what benefits a developer, organization or company might get by choosing one of those.

On 16-01-29 13:03, Ivan Žužak wrote:

> Hi Jonas,
> 
> Can you open an issue on the project's issue tracker?
> 
> https://github.com/github/choosealicense.com
> 
> Thanks,
> Ivan
> 
> > > http://choosealicense.com/
> > > 
> > > This should REALLY mention AGPL3 and LGPL3 instead of GPL3 as these two are probably much more fitting for all the projects on github than just GPL3.

Edit: fixed wrong negation above.

mlinksva:
(on home page, non-hidden) -> (agplv3 to unlicense)
- emphasize macro (spectrum of oss licenses) rather than micro (families/variations) structure
- rm <100% accurate 'family'/'variant', and 'featured'/'hidden' fields no longer needed to control display
- closes #331 (not a pure fix; feel free to open new issue)
- reverse order as on home page, #335 claims order potentially privileges


benbalter:
@mlinksva would you be able to post a few screenshots and talk a bit more about the motivation? Suspect this is the right move, but hard to tell from a few low-level bullets.

mlinksva:
The first bullet "emphasize macro (spectrum of oss licenses) rather than micro (families/variations) structure" is the big picture motivation.

The choices between GPLvN, LGPLvN and MIT/ISC/BSD-x aren't ones someone who doesn't already know about should be confronted with when there's a big picture to get. Eclipse and Artistic aren't licenses anyone should choose unless they already know about them because they're in the relevant communities.

before (left) after (right) screenshots

choosealicense-montage

waldyrious:
Am I missing something? How does one get to the full listing of licenses? I believe the summary in terms of permissions/conditions/limitations is very useful, and even though I get the reasoning for not providing an exhaustive list (of the project as a hole, and of this change in particular), I think completely hiding such useful summarized views from visitors does a disservice to them, as that information is hardly available elsewhere.

mlinksva:
@waldyrious there never has been a full listing of licenses presented on the site. I can't find it now, but I recall suggesting or reading a suggestion that every license we have a page for be linked at the end of /licenses for discoverability purposes. I don't feel strongly one way or the other whether that's a good idea so I haven't opened an issue or implemented, but it'd be easy to do.

waldyrious:
> there never has been a full listing of licenses presented on the site.

I understand. I meant the "fuller" list that is displayed now. In effect, curating a smaller subset as this PR proposes has a similar (and arguably higher) value as the curated subset that is currently displayed, so having as available options only the proposed shortlist and an exhaustive dump does lose the value of the intermediate selection. But I am not strongly opposed to that, I just want to make sure that the decision to get rid of the intermediate-size listing is explicit.

> every license we have a page for be linked at the end of /licenses for discoverability purposes.

:+1:

benbalter:
Ah awesome thanks. A few questions:
1. Why are we going from most restrictive to least restrictive? Assuming people have a short attention span, shouldn't we default to the most open?
2. It looks like we're no longer displaying some of the licenses currently displayed. What's the rationale behind that? Why those licenses? Are we now "recommending" a smaller subset of (family-agnostic) licenses?
3. Was there a discussion about Unlicense versus CC0 a while back? I can't seem to find it. Thinking maybe we should prefer CC0?
4. It'd require some small design effort, but what if we made the continuum explicit with a TLDR on the side of the delta between each licenses (or at last "more restrictive" and "least restrictive"). E.g., between unlicense and MIT, you retain copyright. Between MIT and Apache, you also grant patents, etc.

mlinksva:
1. (a) first bullet "reverse order as on home page, #335 claims order potentially privileges"; re short attention span -- home page has correct order from your perspective -- but an aside: I wouldn't phrase as "more open", if I were trying to be neutral (b) unconditional licenses least well developed/accepted of spectrum (eg neither CC0 nor Unlicense on OSI list) so would rather not start there as much as I personally favor
2. Except for CC0 the ones that would no longer appear on /licenses are covered by above "The choices between GPLvN, LGPLvN and MIT/ISC/BSD-x aren't ones someone who doesn't already know about should be confronted with when there's a big picture to get. Eclipse and Artistic aren't licenses anyone should choose unless they already know about them because they're in the relevant communities." CC0/Unlicense choice is similar -- to the MIT/ISC/BSD-x one -- cultural baggage, not substantive or helpful to present learner with.
3. Unlicense is defaulted to in the present UI, so just going with that. CC0 is more problematic than Unlicense due to CC0's explicit patent non-grant.
4. I've got zero design skills, but would be happy to see something like that. I think the decreasing number of conditions is already a pretty strong visual indicator of what is happening.

ionas:
I like the new design to the right. It allows you to compare very quickly what you want and not just choose something you know (MIT or GPL2 for instance). Thanks!

I also like the logical ordering. The top license enforces "about" the most, the most bottom enforces the least - allowing you to move back and forth between what you requirements and/or believes are.

>

wolftune:
I wish I could +100 instead of +1. This is a _major_ improvement.
- It takes what was a complex, confusing 2-factor system and makes it a single linear list with one factor.
- On one side is the most _protective_ of openness, on the other side is the most permissive.
- Being the opposite order as the homepage provides a balance between which side of this list is shown first.
- It's visually easy to compare
- removing Artistic and Eclipse is definitely correct. There's ZERO sense in recommending those to anyone who isn't already planning to use them.
- removing Public Domain in favor of Unlicense is alright

It would be ideal from a design perspective if the one new bullet for each item going from top to bottom was highlighted somehow to indicate "moving up to this adds this bullet" more visually. But I think that would probably fit into a more full redesign.

_Note of concern:_ as much as I favor going with the updated latest licenses, it remains the case that GPLv2 is indeed still used and provides some compatibility, especially as GPLv2+, that GPLv3 doesn't. Incidentally, this is less relevant to AGPL (which had no v2) and LGPL (which is less common and less compatibility issues). So, I would recommend that in the one case of GPLv3, there be an extra little link that would flip open an explanation about GPLv2 saying that it's still in use, and there are some compatibility reasons to stick to GPLv2 even though it lacks the updates of GPLv3; and there could be a mention about the "or later" option for all the GPL family somewhere. I don't think this stuff belongs in the core list, but for someone trying to make sense of things from this site, at _least_ having a link to explain about the version situation with GPL is warranted.

wolftune:
Another idea to consider: have a "similar licenses" dropdown under various things. This does _not_ distract from the linear list because this would only be used where it really is somewhat inconsequential. Basically the idea would be to state that, e.g. BSD-2 and BSD-3 are effectively interchangeable with MIT, because this is not _strictly_ about choosing a license, it can also be about learning to recognize licenses. I.e. I'd hope people can quickly see BSD licenses in use and recognize "oh, that's comparable to MIT in that spectrum of licenses". I'd only support this for BSD and maybe any other _really_ popular licenses. And I'd not mind if this was rejected.

ionas:
I don't really buy into any argument based on popularity. But I'd like to see a list of "similar lics" tags as well. However I'd simply link to wikipedia or whatnot because maintaining nitpicky differences between them will be hard.

wolftune:
@ionas agreed overall. @mlinksva work here is great, and for anything beyond this with more really thorough stuff, it could make sense to just link to the well-designed page at https://www.gnu.org/licenses/license-list.html#SoftwareLicenses

ionas:
Don't get me wrong, I'd personally link gnu.org but I think to be on neutral ground I'd rather link to the license entries on Wikipedia. There is also multi language options available over there for most of them.

wolftune:
that particular gnu.org page is just superior to about any other in its clarity and thoroughness.

Anyway, to not get too tangential, my concern about GPL version from https://github.com/github/choosealicense.com/pull/386#issuecomment-207825053 above is outstanding too

mlinksva:
I don't think educating users about why they might want GPLv2+ instead of GPLv3(+) is a net win, at least on /licenses. Talk about a lot to digest. I'd love to swap in copyleft-next at some point, when it has a version that meets criteria for including in the site. There's a related unfortunate issue that I have yet to open but will soon...

kentfredric:
> removing Artistic and Eclipse is definitely correct. There's ZERO sense in recommending those to anyone who isn't already planning to use them.

How? Surely the value of a license stands on the merit of what the license allows you to do/restricts, and if somebodies interest falls squarely in line with an artistic license, then..... surely, they should use artistic, because it reflects what they want.

wolftune:
@kentfredric you didn't provide any concrete reasoning here. Are you even aware of a single aspect of the Artistic license that would make it a better fit than others already on the list? Your comment could be applied to any of hundreds of licenses that you don't even bother to read. Most licenses are redundant. Do you actually know of a single real-world case where Artistic and Eclipse licenses do something someone wants over the licenses this change keeps, other than being the license a particular community already uses?

wolftune:
@kentfredric (posted a bit hasty initially), the trouble is that there are hundreds of licenses if someone were to pick just right one for each preference. This list is necessarily a subset and small selection. Having re-read Artistic v2, I see it has some merit as being between MIT and MozillaPL, but still not sure it really is valuable to be included.

mlinksva:
> Artistic v2, I see it has some merit as being between MIT and MozillaPL

Apache License 2.0 fills this space. Note at the granularity currently described by permissions/conditions/limitations fields here, these two licenses are exact matches. Of course the licenses have subtle differences, but I don't see any that are worth calling out for people who aren't license nerds, something which contributors to open source should not be required to become.

I've no objection to anyone using Artistic 2.0 if it's what they want, but listing it on /licenses just makes the page noisier.

benbalter:
Noted over in https://github.com/github/choosealicense.com/issues/434, but similar to https://github.com/github/choosealicense.com/pull/398, there's value in exposing the `hidden` field (and using it for the site's templating, rather than hard-coding preferences, to distinguish license prominence.

blahah:
@mlinksva is there an issue already about re-adding CC0 to the UI (I only found closed ones and don't want to create new ones if you don't want to entertain them)? Removing it has killed the usefulness of the github new repo license UI for me and collabs. The unlicense is much more problematic than CC0 in its lack of fallback for jurisdictions where total public domain dedication is not possible. (happy to discuss in detail)

trucktank (commented):
`https://github.com/AdobeDocs/analytics-1.4-ap`

PR #392, opened 2016-04-13
rm patent-use permission field from GPLv2 and LGPLv2.1 metadata

mlinksva:
There's a strong argument they have implied patent licenses, but
this site doesn't annotate any other implied patent licenses, as
one would expect given the description of the patent-use field "This
license provides an express grant of patent rights from the contributor
to the recipient."

### DIFF
--- a/_licenses/gpl-2.0.txt
+++ b/_licenses/gpl-2.0.txt
@@ -24,7 +24,6 @@ permissions:
   - commercial-use
   - modifications
   - distribution
-  - patent-use
   - private-use
 
 limitations:
--- a/_licenses/lgpl-2.1.txt
+++ b/_licenses/lgpl-2.1.txt
@@ -19,7 +19,6 @@ permissions:
   - commercial-use
   - modifications
   - distribution
-  - patent-use
   - private-use
 
 limitations:


waldyrious:
Should we have an "implied" status for some properties like the "patent-use" and "trademark-use"?

mlinksva:
Maybe. That would probably mean adding something like implicit-patent-use to permissions for licenses which do not have an express patent-use permission or limitation, and with a little less certainty, implicit-trademark-use to limitations for licenses which do not have an express trademark-use limitation.

Recent talk covering both, for anyone interested https://fosdem.org/2016/schedule/event/trademark_not_implied/

This wouldn't necessarily be a huge departure from the current set of fields in the sense that most catalog things explicitly stated in licenses, but some things are taken (such as private and commercial use) even if licenses don't explicitly call them out because what they do call out is more broad. On the other hand it would be, because although implied patent/lack of implied trademark license is widely assumed, I'm not sure they're considered settled.

BTW we could also add various properties to cover additional patent terms present in a variety of licenses, such as [L]GPLv2 possible jurisdiction limitations, [L]GPLv2+ condition that parties with conditions imposed on them that are incompatible with their obligations as licensees they cannot distribute, and possible jurisdiction limitations, *GPLv3 downstream shielding provision, various termination upon patent litigation clauses, and varying scope of the express patent grant that the patent-use permission currently describes...I think this level of detail would be mostly out of scope of the site's goals, but if anyone wishes to submit issues or pull requests on any of these, I think we should at least consider them.

waldyrious:
Thanks for the pointer, I'll have a look.

Meanwhile, just to clarify: when you say

> implicit-patent-use to permissions for licenses which do not have an express patent-use permission or limitation

you're referring to defaulting to the patent-use limitation, rather than the patent-use permission, right? (I wonder if the nomenclature could be clarified here...)

mlinksva:
No, defaulting to implicit patent license, ie implicit patent-use permission, if the open source license says nothing about patents.

waldyrious:
In that case doesn't it contradict with what you said later on?

> implied patent/lack of implied trademark license is widely assumed

mlinksva:
No, but what I said before was probably unclear.

If an open source license doesn't mention patents, arguably it includes an implied patent license.

If it doesn't mention trademark, arguably it doesn't include an implied trademark license.

The talk I linked above addresses why these (which you might think of as 'defaults', if a license does not mention the topics) outcomes are different for patents and trademarks.

waldyrious:
Ok, thanks. I recall reading that the default was more restrictive in both cases, but I'll check the talk then, once I get some time :)

PR #395, opened 2016-04-22
rm symlinks, maybe fix #383 (Windows build)

### LINKED ISSUE #383: Conversion error when running the site locally
_Summarized (continued from another topic)_

Let me state my lack of Jekyll know-how and as @mlinksva said this can help for any future contributors.  

Following [Run it on your machine](https://github.com/github/choosealicense.com#run-it-on-your-machine) (Win7) to view some changes for a PR (see #355 ) when executing _script/server_ I get this error:

`Conversion error: Jekyll::Converters::Scss encountered an error while converting 'assets/css/application.scss': Invalid CSS after "...s/normalize.css": expected "{", was "" on line 1`

**Steps taken:**
- Updated all dependencies (Jekyll, ruby, bundler...)
- Went through [Jekyll on Windows by juthilo](http://jekyll-windows.juthilo.com/) (ensuring proper encoding is used in the terminal and the file)
- Tried a vagrant box, while ssh'd the script throws same error [Jekyll on Win via Vagrant](http://www.jamessturtevant.com/posts/running-jekyll-in-windows/)
- Tried to run jekyll serve without any flags and same result
- Removing the yaml front matter block from application.scss, the script works but it doesn't generate a application.css when viewing the site in browser.

Besides throwing windows out the window :+1:
...I'll keep troubleshooting.

mlinksva:
Not certain this is right thing to do, but
https://jekyllrb.com/docs/assets/ says "Note that the sass_dir
becomes the load path for Sass imports, nothing more" so setting
it to `assets/vendor` and changing imports accordingly ought and
seems to work without disturbing other files under `assets/vendor`,
even though the docs continue "This folder should only contain
imports."

Motivated by #383

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -41,7 +41,6 @@ gems:
   - jekyll-coffeescript
 
 sass:
-    sass_dir: assets/vendor
     style: :compressed
 
 twitter:
--- a/_config.yml
+++ b/_config.yml
@@ -41,7 +41,7 @@ gems:
   - jekyll-coffeescript
 
 sass:
-    sass_dir: _sass
+    sass_dir: assets/vendor
     style: :compressed
 
 twitter:
--- a/_sass/jquery.qtip.scss
+++ b/_sass/jquery.qtip.scss
@@ -1,1 +1,0 @@
-../assets/vendor/qtip2/jquery.qtip.css
--- a/_sass/normalize.scss
+++ b/_sass/normalize.scss
@@ -1,1 +1,0 @@
-../assets/vendor/normalize-css/normalize.css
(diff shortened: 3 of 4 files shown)


parkr:
:shipit:

parkr:
screen shot 2016-04-22 at 11 51 22 am

mlinksva:
Hmm, I'm afraid this broke something. The tooltips styled by qtip are now showing up at bottom of page rather than in context.

isc license - choose a license

benbalter:
If it works, :+1:. It was a not-so-clean hack at the time, which I didn't realize would affect windows users. Hopefully, when Gem-based themes ship, this won't be a problem.

mlinksva:
I failed to make it work initially; I hadn't understood sass' magic and I can't make that work outside a dedicated single level directory. I made the tooltips work again at price of 2 (now successful) additional http requests (qtip and normalize get referenced from rather than bundled into application.css).

Issue #238, opened 2015-01-03
Explain common terms like "copyright", "patent" and "license"


strugee:
Most people have a basic grasp of licensing-related stuff, but are hazy on the exact details of what the different terms mean. It would be nice if we could have a FAQ or something to explain all of it. I'd send a patch, but I'd like to solicit some feedback on what people think the exact format should be.

Here's a list of things we might want to put on such a page:
- "Copyright"
- "License"
- "Public Domain" (specifically the fact that waiving copyright law includes waiving your right to relicense, IIUC)
- "Patent" (especially since the distinction between a patent and a copyrighted work is lost on a lot of people)
- "Copyleft"
- Warranty disclaimers/liability (maybe?)

bhuga:
This is a good idea, but I'd like to hear your thoughts on two concerns I have:
- These words mean different things in different jurisdictions. Can we make a definition of "public domain" or "patent" that is meaningful for everyone without misrepresenting for at least some readers?
- How do we expose this information? I think the site would become a little muddy to read if every instance of the word "license" was a link to a glossary entry, but I also feel like a link on the front page wouldn't be very helpful. By the time you knew you needed a glossary, it'd be too late.

strugee:
> These words mean different things in different jurisdictions. Can we make a definition of "public domain" or "patent" that is meaningful for everyone without misrepresenting for at least some readers?

I think so? I'm not enough of an expert on non-US intellectual property law to say definitively, but I think if we make it general enough we can find a balance. E.g. "patent" could be defined as something along the lines of, "Something that places legal protections on ideas such that the law treats the patented idea similar to how it treats property. Patents " That blurb, while getting the main idea across, doesn't really say anything specific about the law.

Alternately, we could add a note stating that the definitions provided are based on U.S. IP law.

JeromSar:
I think it's also important to clarify several other terms which use the word "patent".
For example, the Apache license includes a so called "patent grant", but the term is never further explained. For me, this was confusing and raised several questions:
- What is a patent?
- Does my code automatically have a patent when I publish it?
- What happens when my license includes a patent grant?
- What is restricted if the license doesn't contain a patent grant?
- What is the difference between patents and copyright?

I think an explanation for some of these could be included in the popovers that appear at the right side of the screen when hovering over the different characteristics.

ionas:
distribution, installation, service should also be cleared.

waldyrious:
[...]

PR #396, opened 2016-04-24
Rework descriptions of 7 listed on /licenses

mlinksva:
- primarily functional
- drop self-naming
- minimize requiring significant understanding of other licenses or
  copyright
- should be excruciatingly bland for anyone who already knows the licenses
  well; newcomers shouldn't have to deal with baggage immediately

Probably a few more words should be added to the xGPLv3s about their
stronger patent terms.

Licenses not listed on /licenses could be described in similar style.

Suggestions and criticisms welcome!

### DIFF
--- a/_licenses/agpl-3.0.txt
+++ b/_licenses/agpl-3.0.txt
@@ -4,7 +4,7 @@ nickname: GNU AGPLv3
 redirect_from: /licenses/agpl/
 source: http://www.gnu.org/licenses/agpl-3.0.txt
 
-description: The GNU GPL family of licenses is the most widely used free software license and has a strong copyleft requirement. When distributing derived works, the source code of the work must be made available under the same license. GNU AGPLv3 is distinguished from GNU GPLv3 in that hosted services using the code are considered distribution and trigger the copyleft requirements.
+description: The license with the strongest conditions for preserving openness. Complete source code of licensed works and modifications, which include larger works using a licensed work, must be made available under the same license when distributed. Copyright and license notices must be preserved. Contributors provide an express grant of patent rights to users. Additionally, when a modified version is used to provide a service over a network, the complete source code of the modified version must be made available to users of the service.
 
 how: Create a text file (typically named LICENSE or LICENSE.txt) in the root of your source code and copy the text of the license into the file.
 
--- a/_licenses/agpl-3.0.txt
+++ b/_licenses/agpl-3.0.txt
@@ -4,7 +4,7 @@ nickname: GNU AGPLv3
 redirect_from: /licenses/agpl/
 source: http://www.gnu.org/licenses/agpl-3.0.txt
 
-description: The license with the strongest conditions for preserving openness. Complete source code of licensed works and modifications, which include larger works using a licensed work, must be made available under the same license when distributed. Copyright and license notices must be preserved. Contributors provide an express grant of patent rights to users. Additionally, when a modified version is used to provide a service over a network, the complete source code of the modified version must be made available to users of the service.
+description: Permissions of this strongest copyleft license are conditioned on making available complete source code of licensed works and modifications, which include larger works using a licensed work, under the same license when distributed. Copyright and license notices must be preserved. Contributors provide an express grant of patent rights to users. Additionally, when a modified version is used to provide a service over a network, the complete source code of the modified version must be made available to users of the service.
 
 how: Create a text file (typically named LICENSE or LICENSE.txt) in the root of your source code and copy the text of the license into the file.
 
--- a/_licenses/agpl-3.0.txt
+++ b/_licenses/agpl-3.0.txt
@@ -5,7 +5,7 @@ nickname: GNU AGPLv3
 redirect_from: /licenses/agpl/
 source: http://www.gnu.org/licenses/agpl-3.0.txt
 
-description: Permissions of this strongest copyleft license are conditioned on making available complete source code of licensed works and modifications, which include larger works using a licensed work, under the same license when distributed. Copyright and license notices must be preserved. Contributors provide an express grant of patent rights to users. Additionally, when a modified version is used to provide a service over a network, the complete source code of the modified version must be made available to users of the service.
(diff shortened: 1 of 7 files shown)


waldyrious:
It might be worth considering adding a "glossary" page describing terms like "copyleft", "permissive", etc. That way, more verbose descriptions (such as those introduced in this PR) need not be in the first sentence describing licenses, but we can instead use such jargon while linking to the appropriate entry in the glossary (or even showing the contents of the definition, if we ensure they're all sufficiently short, in a tooltip).

The more verbose descriptions could still remain, for those new to licensing, but we'd be able to provide a quick overview in the first sentence for reasonably license-savvy people.

mlinksva:
@waldyrious a glossary could be a good addition to the site; I've left #238 open for that purpose.

I had hoped to avoid using the terms 'copyleft', 'permissive', and 'public domain' entirely in part because they help people mistake commitment for savviness,  but in c208a53 they are included. Better?

waldyrious:
:+1:, although ideally those terms would have links to the glossary. I've added a comment on #238 to help keep that in mind.

_licenses/agpl-3.0.txt
```diff
@@ -4,7 +4,7 @@ nickname: GNU AGPLv3
 redirect_from: /licenses/agpl/
 source: http://www.gnu.org/licenses/agpl-3.0.txt
 
-description: The GNU GPL family of licenses is the most widely used free software license and has a strong copyleft requirement. When distributing derived works, the source code of the work must be made available under the same license. GNU AGPLv3 is distinguished from GNU GPLv3 in that hosted services using the code are considered distribution and trigger the copyleft requirements.
+description: Permissions of this strongest copyleft license are conditioned on making available complete source code of licensed works and modifications, which include larger works using a licensed work, under the same license when distributed. Copyright and license notices must be preserved. Contributors provide an express grant of patent rights to users. Additionally, when a modified version is used to provide a service over a network, the complete source code of the modified version must be made available to users of the service.
```
benbalter: I wouldn't assume the reader knows what "copyleft" is.

>  which include larger works using a licensed work

This reads a bit like legalese to me. There's traditionally been a conflict between completeness / thoroughness and readability. When possible to still be accurate, we've erred on the site of readability. The idea being, that the target user for the site is a first-or-second time project creator, who's just looking to license their project (e.g., coming from the dropdown on .com), and we should optimize for their 80% use case. Yes, there will be edges that won't be 100% accurate or fleshed out, but we're not trying to solve for every possible licensing concern, or write the canonical treatise on open source licensing. We're trying to help the user choose a license for their project.

benbalter: Put another way, our approach has historically been closer to [explain like I'm five](https://www.reddit.com/r/explainlikeimfive/) than to explain-like-I'm-about-to-go-to-court.

mlinksva: re expectation of knowing what "copyleft" is; I agree! But it's always been on the [home page](http://web.archive.org/web/20130716093023/http://choosealicense.com/) which I think is far more important to be understandable. Suggestions welcome on how to get rid of it from either or both places.

On the bit like legalese bit, I agree. But it's hard to distinguish between licenses with some degree of accuracy without sounding like that. But I'd love to improve it. I think the main mitigation is that most users shouldn't get to /licenses, they should select one of the licenses on the home page, and among those the distinctions are more stark, so one can be accurate and very high level.

I've heard the phrase explain to me like I'm 5 before but never looked at a site dedicated to the concept. Not sure the reddit succeeds. I'd love to see proposed explanations of the main licenses as if reader were 5. Of the current descriptions, I'd say only MIT comes close, and that description has by far the easiest time of it (yet I don't know what "proper attribution" means). I think I'll open an issue with that challenge.

Finally I think there's some value in having an accurate and brief summary of each license even if it reads a bit like legalese, as those details matter and are really hard to extract through cultural osmosis or even just reading a license. But possibly that should be different from the description, especially if we had EILI5 descriptions.

benbalter: By all means, I'm not suggesting we should assume our reader is actually five (or write as if they were), but likewise, we shouldn't assume that the reader absolutely needs an accurate and complete summary, or that they have the patience/stomach to grok through legalease. 

We should be comfortable being a little wrong if it means making things easier for the vast majority of users. Our job is to absorb at least some of that complexity on behalf of the reader and find a happy medium (where on that continuum we land is obviously up to you, but I think we can both agree that both 100% accurate/complete and 100% readable - the two extremes - are not good goals to strive for).

> as those details matter

They do for the edge cases, but do they for 50% of users? [80%?](http://ben.balter.com/2016/03/08/optimizing-for-power-users-and-edge-cases/) 90%?

mlinksva: Actually I do think 100% accuracy is a worthy goal. But I'm willing to drop _lots_ of precision in the interest of readability. The text in this PR can be whittled down in that direction, I'll work on it some more.

Anything not on the home page is an edge case. I'd like the home page to work for 80%+.

_licenses/mpl-2.0.txt
```diff
@@ -3,7 +3,7 @@ title: Mozilla Public License 2.0
 redirect_from: /licenses/mozilla/
 source: https://www.mozilla.org/media/MPL/2.0/index.txt
 
-description: The Mozilla Public License (MPL 2.0) is maintained by the Mozilla foundation. This license attempts to be a compromise between the permissive BSD license and the reciprocal GPL license.
+description: Permissions of this weak copyleft license are conditioned on making available source code of licensed files and modifications of those files under the same license (or in certain cases, the GNU AGPLv3, GPLv2, LGPLv2.1, or later versions of those licenses) when distributed. Copyright and license notices must be preserved. Contributors provide an express grant of patent rights to users. However, a larger work using work may be distributed under different terms and without source code for other files the larger work.
```
Zimmi48: I very much agree with this rewording. (Previous wording did not carry at all the meaning of MPL 2.0.)

Maybe `(or in certain cases, the GNU AGPLv3, GPLv2, LGPLv2.1, or later versions of those licenses)` could be shortened. For instance, to `(or in certain cases, under a license of the GNU family)`.

The last sentence has some language mistakes. I would correct it to `However, a larger work using the licensed work may be distributed under different terms and without source code for the files added in the larger work.`

mlinksva:
Removed some mostly unnecessary words. I'm going to go ahead and merge now as this is a big improvement in accuracy and consistency over what's on the site now, but I admit it isn't completely ideal. Further improvements _very_ welcome in the direction of _more friendly/basic/just less legalese-esque words while still being conceptually accurate_.

PR #399, opened 2016-04-27
Restructure individual license pages, added styles to improve alignment

devGHBot:
This is in reference to #355 and an update to #382 (which essentially makes it obsolete)

Always wonderful visual:
solo-license-page-restructure


mlinksva:
Looks great to me. Wonderful! Only quibble is I'm not sure the License Text heading is needed.

Will leave this open to collect other feedback since it is a pretty big change to the individual license pages.

benbalter:
Really like this change, and agree with @mlinksva that the license text should be self-explanatory

PR #400, opened 2016-04-28
don't use the same color for headings and links

waldyrious:
Changed to a slightly-blue grey, which IMO fits well with the current style.

Screenshot:
screenshot

### DIFF
--- a/assets/css/application.scss
+++ b/assets/css/application.scss
@@ -80,7 +80,7 @@ strong {
 }
 
 .home h2 {
-  color: #149ad4;
+  color: #687072;
   font-size: 27px;
   line-height: 70px;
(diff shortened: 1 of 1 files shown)


mlinksva:
I'm for this. A distinct color for links is the way things are supposed to be. 😄 

I've zero idea whether slightly-blue grey is the best for headings instead of the same color as links. Looks fine to me. Will leave open a bit to see if anyone has an informed different opinion.

Thanks for opening this @waldyrious !

waldyrious:
> Thanks for opening this @waldyrious !

As usual with FOSS, I'm merely scratching a personal itch (I instinctively attempted to click the "What if none of these work for me" heading), so no need for thanks :)

PR #405, opened 2016-05-05
Better limitations:trademark-use

ShaneCurcuru:
Existing phrase is likely confusing to new readers; trying to better clarify the fact that copyright licenses don't grant you trademarks anyway, but that these licenses are explicit about letting you know trademarks aren't copyrights.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ The license properties (rules) are stored as a bulleted list within the licenses
 
 #### Limitations
 
-* `trademark-use` - While this may be implicitly true of all licenses, this license explicitly states that it does NOT grant you any rights in the trademarks or other marks of contributors.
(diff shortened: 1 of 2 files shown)


mlinksva:
The existing language is a bit unwieldy and this is a net improvement. But I like "may" in the current language; not sure it can be stated with 100% certainty no implicit trademark license is granted by any common licenses (and why mention 'common' licenses; are there less common licenses that arguably do grant an implicit trademark license?) How about

> This license explicitly states that it does NOT grant you any trademark rights, even though licenses without such a statement probably do not grant you any implicit trademark rights.

?

Also, this text is duplicated in `README.md`. Would you mind adding a commit to change there too?

Thanks!

PR #406, opened 2016-05-09
Appendix/table of all repo licenses; closes #388

mlinksva:
This is quite ugly. Improvements or something better welcomed.

Closes #388 see there for motivation.

Additionally I wish I had made a table like this a long time ago, would have caught more license description errors earlier.

PR #411, opened 2016-05-14
Legend for appendix table of licenses

mlinksva:
I don't claim it looks good. Improvements and criticism most welcome.

cal-appendix-legend

As mentioned in https://github.com/github/choosealicense.com/pull/406#issuecomment-218225779 (I think it complements the tooltips) and https://twitter.com/yschaus/status/730674072129048576

### DIFF
--- a/appendix.md
+++ b/appendix.md
@@ -53,3 +53,33 @@ All licenses described in the choosealicense.com [repository](https://github.com
   </tr>
 {% endfor %}
 </table>
+
+## Legend
+
+<p>Open source licenses grant to the public <b>permissions</b> (<span class="license-permissions"><span class="license-sprite"></span></span>) to do things with licensed works copyright or other "intellectual property" laws might otherwise disallow.</p>
+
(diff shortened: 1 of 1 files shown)


benbalter:
What does having the legend on the page provide the user that the tool tips does not?

Would "🔴 Limitations 🔵 conditions 🍏 permissions" suffice?

mlinksva:
If a legend doesn't add anything not provided by the tool tips, an abbreviated legend is not justified, for the tooltips state "Limitation" etc in their title bars.

If I were to encounter a table like this for a domain I didn't know, I'd want a legend so that I could read and not only hunt and peck, to mix metaphors.

I'll leave this open for other feedback. If anyone can point to applicable evidence on whether legends help or harm comprehension or some other desirable outcome please do point that out.

mlinksva:
> Bolding the specific legend term whenever a header is clicked on so the user can find their place faster with some JS.

Yes, I always like that as a reader, wish were default behavior. Only requires CSS, done in #426 

Thanks for the suggestions @LeoFNaN!

I have no sense of what font size or colors would make sense for your other two ... please open an issue or pull request if you have an idea.

PR #414, opened 2016-05-16
Add a note about more info available on hover

phanimahesh:
It wasn't obvious to me, but expected something to happen on click/hover on circles because there is no legend anywhere. Let's remove the guesswork.

### DIFF
--- a/appendix.md
+++ b/appendix.md
@@ -5,7 +5,7 @@ title: Appendix
 class: license-types
 ---
 
-All licenses described in the choosealicense.com [repository](https://github.com/github/choosealicense.com), in a table.
+All licenses described in the choosealicense.com [repository](https://github.com/github/choosealicense.com), in a table. Hover on the circles for more info.
 
 <table border style="font-size: xx-small">
 {% assign types = "permissions|conditions|limitations" | split: "|" %}


mlinksva:
Possibility of adding a legend in #411 

Do you mean you didn't expect something to happen on click/hover over the circles and discovered it by accident?

If we don't add a legend I agree taking guesswork out of discovering tool tips would be good. "Hover on the circles for more info" probably works.

I'll leave this open for discussion along with #411. Thanks for the pull request!

phanimahesh:
No, I was searching for some legend, and after not finding one, tried click/hover just to see if there's anything that'll happen. I browse primarily using keyboard.

Semi-related note: Screen readers will be easier to use if empty cells also contained aria text or an item with hover similar to the current ones. It can say something along the lines of 'There is no mention of X in the license'

mlinksva:
@phanimahesh how about just "not applicable" aria text in empty cells? If you want to make a pull request along these lines, would be greatly appreciated.

mlinksva:
Closing this as legend was added in #411. @phanimahesh I'm not sure how to implement your semi-related note, but feel free to open another pull request on that matter. Thanks!

PR #417, opened 2016-05-23
Allow auto filling like for MIT license {yyyy}{name of copyright owne…

gwpl:
Allow auto filling like for MIT license
 `{yyyy}{name of copyright owne…r}`->`[year] [fullname]`

TL;DR- Bug this Commit is fixing: When Autogenerating MIT , year and name are automatically filled. When autogenerating Apache 2.0 they are not. I would expect behaviour to be consistent - either autofill everwhere, either nowhere.

Actually two fields: `{yyyy} {name of copyright owner}` in Apache
License are not automatically substituted.

I see them automatically substituted in case of X11 AKA MIT License,
here it is, initial commit for same workflow but by selecting MIT
instead of Apache 2.0:
https://github.com/gwpl/test_autogen_MIT_LICENSE/commit/6efc5748d77c528c976e0457dbdb3ea7f6298115
as you see in MIT license both: year and author got substituted.

Here is how it is defined for MIT `[year] [fullname]`: https://github.com/github/choosealicense.com/blob/gh-pages/_licenses/mit.txt#L30

I would expect behaviour to be consistent across all licenses.

### DIFF
--- a/_licenses/apache-2.0.txt
+++ b/_licenses/apache-2.0.txt
@@ -219,7 +219,7 @@ limitations:
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright {yyyy} {name of copyright owner}
+   Copyright [year] [fullname]
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.


mlinksva:
Thanks for this PR, I appreciate your attention to detail.

The boilerplate example in Apache 2.0 is not analogous to the copyright notice in MIT. I don't think the former is meant to be modified in the LICENSE file and I don't recall seeing it modified.

I'll leave this open for awhile in case anyone wants to point out evidence to the contrary.

gwpl:
Still, if one creats new project in github and select "autogenerate LICENSE file" , IMHO github should behave consistently. Therefore, either auto-fill it in all cases, or leave it as template in all cases. Eventually display warning in case of exception, e.g. "Please Remember To Modify LICENSE Template appropriately".

(btw. I've accepted GitHub CLA and still check is lagging)

gwpl:
Still IMHO Github should behave consistently for all license types, when one creates new project and select "autogenerate LICENSE". Either fill it always, either never, either display clear message "Remember to fill LICENSE template" if required.

(Btw. If just accepted GitHUB CLA, and still displays this check as failing...)

mlinksva:
> Therefore, either auto-fill it in all cases, or leave it as template in all cases.

I see where you're coming from, but MIT and Apache are different. MIT is a template, Apache is not. The Apache license provides example boilerplate, but this is never modified in the copy of the license.

I'm closing this for now, but if you can find widespread practice to the contrary, please document in another comment and I'll reconsider.

PR #418, opened 2016-05-24
Make nickname a required meta field

mlinksva:
Add missing nicknames

Use SPDX ID if no customary nickname (eg GNU GPLv3) exists

This ensures that a relatively compact name is always available

I may be missing some obvious customary names, e.g., is "Eclipse
1.0" customary? For now I've used the SPDX ID, EPL-1.0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,8 @@ Licenses sit in the `/_licenses` folder. Each license has YAML front matter desc
 
 #### Required fields
 
-* `title` - The SPDX-compliant, human-readable license name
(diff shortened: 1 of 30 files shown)


benbalter:
👍 Good idea.

mlinksva:
I think I have a better idea though: make a new `spdx-id` field required. This would provide the case-sensitive version people [sometimes want](https://github.com/benbalter/licensee/issues/72) and make for a default short display value. Nickname would still be available for licenses with customary names, and user interfaces that wanted to could prefer nicknames.

PR #420, opened 2016-05-29
add scope=col or row to appendix table

mlinksva:
As recommended by
https://www.w3.org/WAI/tutorials/tables/two-headers/
http://webaim.org/techniques/tables/data

Prompted by @phanimahesh [comment](https://github.com/github/choosealicense.com/pull/414#issuecomment-219495783) I searched for info on table accessibility and found above recommendations. I didn't find anything on empty table cells other than to not use them to obtain borders or spacing, which we don't.

Further accessibility issues and pull requests would be most welcome.

### DIFF
--- a/appendix.md
+++ b/appendix.md
@@ -10,7 +10,7 @@ All licenses described in the choosealicense.com [repository](https://github.com
 <table border style="font-size: xx-small">
 {% assign types = "permissions|conditions|limitations" | split: "|" %}
 <tr>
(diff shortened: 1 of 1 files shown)


phanimahesh:
Thanks! Sorry I couldn't follow up on my previous comment, got caught up with things. This is definitely an improvement. I'll try testing with a few screenreaders later. :)

PR #421, opened 2016-05-31
License identifiers and package descriptions hint

mlinksva:
Now that licenses' case-sensitive SPDX ID are [available](https://github.com/github/choosealicense.com/pull/418#issuecomment-221404630) I thought it might be useful to expose on individual license pages, along with a hint about how this can be used. Screenshot:

package-metadata-license

I also used this to sneak in an instruction about noting "or later" for _GPLs:

package-metadata-gpl

### DIFF
--- a/_includes/sidebar.html
+++ b/_includes/sidebar.html
@@ -14,7 +14,7 @@
     {% endif %}
     {% assign xgpl = false %}
     {% if page.spdx-id contains 'GPL' %}{% assign xgpl = true %}{% endif %}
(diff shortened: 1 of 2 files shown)


benbalter:
Do we want to encourage users to _only_ add the SPDX identifier to their package manager? If not, I'd make it very clear that it's an _additional_ step, and if so, what the advantage is, both to the publisher and to consumers.

If we tell users to do something, we have to assume that they're going to do it and plan accordingly. In terms of license detection on GitHub.com, that means we'll have to begin parsing package manager files in order to identify a project's license. Practically, that'd mean that there'd be no cross-platform way of describing a project's license (as each package manager has a different standard). And legally, [I'm not sure simply adding the SPDX ID is sufficient](https://github.com/benbalter/licensee/issues/4).

> Now that licenses' case-sensitive SPDX ID are available

The above is to suggest that we should be purposeful about adding complexity to the site (both cognitive and in the best practices we encourage), rather than adding something simply because we can.

mlinksva:
Thanks for the feedback! Added optional and reason:

optional-package-desc

I'll make a detailed comment on benbalter/licensee#4 eventually but in short I don't see evidence of a bright line. More intentionality, prominence, and non-ambiguity are all good, for both grant from original licensor and subsequent contributors. How much project owners do, and how hard tools try to detect, are tradeoffs, no right or wrong.

benbalter:
> Thanks for the feedback! Added optional and reason:

✨ I like that approach a lot, and agree on your analysis below the screenshot.

Issue #401, opened 2016-04-28
New feature: Suggest license directly from site to third-party repository?


kfogel:
It would be wonderful if one could go to http://choosealicense.com/, pick a license, and click a button to create a new pull request that suggests that license to a currently unlicensed third-party repository.

For example, today I ran across this repository on GitHub: https://github.com/ys/forked (the code behind the very useful [Stargazers forks](http://forked.yannick.io/) web service).  The repository doesn't have any license file at all.  It's quite likely that @ys meant it to be open source software, and simply never bothered to put an explicit license on the tree.  But the easiest way to ask @ys would just be to make a PR that adds a license he would be likely to find acceptable.  So I was going to create a PR that adds the text of the MIT license in a LICENSE.md file at the top level of the `ys/forked` tree... And then I thought, gosh, it would be so much easier if I could just go to http://choosealicense.com/, pick the obvious license (in this case MIT), and click a button to do create the PR for me!  A form field would open up, asking me for name of the repository where the PR should be created.  If I'm a member of one or more GitHub organizations, then there would be an intermediate chooser UI widget to ask me where the PR should come from (the same widget that comes up when you're logged in to GitHub and you click "fork" on a repository, and GitHub asks you where you want the new fork to be).

Once I've answered these easy one or two questions, the rest is fully automatic.  No manual copy-and-paste of license text, no clicking around the manual PR creation interface, no need for me to write the commit message -- all of this can be automated by choosealicense.com.

I think if we had this feature, it could result in a lot more repositories getting clear open source licenses.

Thoughts?

-Karl

mlinksva:
I've wanted this to use and to point others to, eg when I wrote https://github.com/github/choosealicense.com/commit/72f852f4ebf5f668c622356a152c3ea41a902884#diff-4b7241b75e16fdab61fb0f7ff4fa6f7fR25

I imagine (but have made zero attempt to verify) that it'd be possible to implement all of this on choosealicense.com with client side javascript, perhaps using https://github.com/philschatz/octokat.js ... otherwise substantial changes would be needed, eg a non-static site.

Completely welcome more informed feedback.

Just to state the obvious, choosealicense.com may not be the only site from which this functionality would be useful, e.g., opensource.org, github.com, though those are doubtless much harder for various reasons.

[...]

PR #422, opened 2016-05-31
expansion/clarification of non-license permissions

bill-auger:
I made some suggestions to this documentation today because I have been wondering about these ambiguities for years and am recently, trying to convince a repo maintainer to add a license to his repo which has over 1000 regular users; but this document (even the section: "Ask the maintainers nicely to add a license") offers nothing convincing to that end.  In fact, it is barely self-consistent, and the GitHub TOS is so terse on the topic that it is not at all clear what this document implies specifically for GitHub users.

For example, the one paragraph states that GitHub public repos are forkable (and actually download-able) regardless of the license or lack thereof.  Then the very next paragraph states that without a license users may not use the material in ANY way.  This avoids blatant contradiction only by omitting that, strictly speaking, without a license the rights do not exist to copy or fork either.

Again I am myself not crystal clear on the specifics and I would be as interested to hear where my assumptions may not be accurate as I am to push this PR through.

I hope this edit will serve to inform users until perhaps GitHub more clearly defines the limits of the permissions granted in section F1 of the TOS.  Until then, I hope that the drafters of the GitHub TOS would read this PR mindfully and note that it raises some important issues.


mlinksva:
@bill-auger thanks for the suggestion! I cut your long addition down to its core points (IIUC):
- (for potential licensors) limited permission to fork doesn't foster collaboration an open source license does
- (for potential users) limited permission to fork doesn't grant permission to use, modify, share

See #423 

>  trying to convince a repo maintainer to add a license to his repo which has over 1000 regular users; but this document (even the section: "Ask the maintainers nicely to add a license") offers nothing convincing to that end

Thanks for trying! What kind of response are you getting? Does the maintainer think they don't need a license because the ToS is enough? Something else?

bill-auger:
i have not actually raised the issue yet - it was in writing up the proposal that i realized it was not convincing even to myself when i could not find anything in GitHub documentation clearly stating (without apparent contradiction) that the TOS is insufficient for sharing - if either this or PR #423 goes through that would be something to point to

i think it is too strong to say that users perceive the TOS as sufficient for their needs - it is probably more correct to say that because downloading is technically possible (indeed trivial) that this alone is "enough" to make proper licensing seem unimportant

mlinksva:
I wouldn't even mention the TOS when requesting someone add a license. I've seen a fair number of add-a-license issues/pull requests and don't think I've ever seen TOS mentioned in the rationale/request or subsequent discussion. But as I mentioned in https://github.com/github/choosealicense.com/issues/401#issuecomment-217308431 I'd love to gather a sampling of such discussions and perhaps create some boilerplate that people can use to request that repo maintainers add a license in the most effective way. In any case I hope you're comfortable with making the request now, and good luck!

PR #423, opened 2016-05-31
Make non-license (non-)permissions more explicit

mlinksva:
Building on/abbreviation of #422 by @bill-auger


bill-auger:
thank you for your prompt attention - this PR satisfies my primary concern which was to make these two paragraphs more harmonious and less apparently contradictory by somehow making more clear that the GitHub TOS alone is insufficient for a repo that is intended to be widely and regularly used

```
.... as fostered by a copyleft or open-source license.
```

i do wish that the word "copyleft" had not been omitted - i specifically included both "copyleft" and "open source" as a compromise but these are not equivalent - if only one term were to be used "open source" should be replaced with "copyleft" because a non-copyleft-style license does not actually foster collaboration - in its current form this qualifier is a mis-statement and would be better removed entirely because only a subset of the licenses listed on this site suggest that anyone share their improvements or contribute back to the project

secondly - though it is not my main concern - the warnings that were removed should be of great concern to GitHub and users should be aware of this trap that GitHub has laid out - perhaps my assumptions are incorrect (that is why that section was perhaps more verbose than necessary) but the issue seems pretty clear
- default copyright does not allow copying
- the TOS grants exceptions to the default copyright allowing users to "view" and "fork" on the server only (but not to "copy" locally)
- GitHubs place an un-removable "download" button on all public repos regardless of the license or lack thereof
- therefore anyone who clicks that button (or clones locally) does so in violation of the copyright

it seems to me that GitHub could avoid this by simply adding ".... and download" to the permission granted in section F1 and i dont think many would object - although as it has been clarified here - the downloaded work would still be of absolutely no usefulness to the downloader

again that was not my main concern but i would be interested to know if the line of reasoning above is accurate - and if so the users should be wary and some warning should remain in this document - this section was not abbreviated - an abbreviation would be something like: "... and you are not permitted to download even if the website makes this easy to do"  - this edit is a complete omission of perhaps the most significant issue

mlinksva:
> non-copyleft-style license does not actually foster collaboration

So the collaboration that happens to build many thousands of non-copyleft but open source (which includes copyleft) projects is something other than collaboration, or not fostered by the license?

99% of what a license needs to do to foster collaboration is remove the possibility of collaborators suing each other; that's why open source and free software definitions mandate freedoms; regulations such as copyleft are optional and their different impacts on fostering collaboration subject to evidence, which is lacking.

> you are not permitted to download even if the website makes this easy to do

This interpretation is probably not correct and in any case choosealicense.com tries mainly to document express provisions not interpret what is not stated. In any case I'll pass along your feedback about the GitHub TOS.

PR #436, opened 2016-06-15
Correct the spelling of "license" in wtfpl

EmilGedda:
Just a minor spelling mistake I found while browsing the site.


mlinksva:
Not exactly a spelling mistake, but we do use the U.S. spelling elsewhere, so for consistency, good find.

PR #103, opened 2013-10-23
Use the plaintext formatting of the Artistic License 2.0

IQAndreas:
Another case where the license was copied from HTML, but had a nicely
indented plaintext version (this is the last one, I checked).

This time the source was taken from 
http://www.perlfoundation.org/attachment/legal/artistic-2_0.txt

PR #440, opened 2016-07-06
Fix instructions for Artistic 2.0 (follow-up #333)

waldyrious:
cf. #103, #310 and #333
cc @mlinksva, @benbalter, @bkeepers and @Haacked


mlinksva:
Thanks, good catch.

waldyrious:
Sorry, there's a small mistake in the text: where it says "they refer to", it should say "it refers to". Can you make that fix directly? I don't think a new PR is warranted.

mlinksva:
> Do not replace the copyright notice (year, author) — it refers to the license itself, not the licensed project.

It.itself reads a bit awkwardly. How about:

> Do not replace the copyright notice (year, author), which refers to the license itself, not the licensed project.

waldyrious:
Sounds good :)

Issue #445, opened 2016-08-20
No License page is vague about copyright


Wuzzy2:
I am not really happy with the No License page http://choosealicense.com/no-license/.
It is written in a way that makes it seem that copyright works exactly the same globally.
This is obviously false, as there are different copyright laws in different countries.

Asking you to go into the national details if of course a bit too much but at least it should be acknowledged that copyright is not the same everywhere, and that therefore the particular claims about copyright made on this page may or may not apply to a particuar country. Maybe some weblinks would help.

I'm not very well read on this subject, so I will not rewrite the text which will probably be wrong anyway.

PR #450, opened 2016-09-09
Update no-license

### LINKED ISSUE #444: Website claims that copyright forbids using software
From http://choosealicense.com/no-license/:

> If you find software that doesn’t have a license, that generally means you have no permission from the creators of the software to use, modify, or share the software.

Wait, what? Even not _using_? I am not a lawyer but this seems surprising to me. But is this actually true? And even if such a restriction applies, is it actually because of copyright law? Or maybe its because of something else?

My gut feeling tells me this claim is false because by the same logic, we would not be allowed by copyright law to “use” copyrighted books (by reading them) because we didn't get permission. ;-)
Can someone who knows more about things like this clear it up, ideally with sources and stuff?

I am also unsure about the “modify” part. Note the text only mentions the modification itself, not the distribution of such a modification. As far my layman knowledge goes, I think copyright is only concerned about copies.

mlinksva:
- qualify "no [actions]" with "without being at risk..."
- mention exceptions and limitations, jurisdiction-specific
- mention that collaborators don't give you permission, either
- remove trivial example copyright notice, suggest adding statement symmetric with for-users section below
- suggest exploring contributor agreement so non-licensor has permission from contributors

closes #444 #445


Wuzzy2:
Thank you.

Issue #449, opened 2016-08-30
EPL 1.0 not appearing in license lists


waynebeaton:
I swear that the EPL used to be an option in the license selection drop-down for a new GitHub repository, but it's not showing up anymore.

AFAICT, the license file meets the criteria outlined on the CONTRIBUTING page.

I don't see how the list of licenses is being fed to GitHub. What do I need to do to make this happen?

mlinksva:
The license is at http://choosealicense.com/licenses/epl-1.0/

The license selection drop-down on GitHub eventually (via benbalter/licensee, which vendors github/choosealicense.com) is informed by changes here.

The change you noticed is due to https://github.com/github/choosealicense.com/pull/386

See #413 and #427 for similar discussion.

I can't promise anything but I'll pass along the desire of people who already know what license they want to have that license available in the license selection UI somehow.

Thanks for the feedback! I'm closing this now as I don't think think there's anything to do for this project, but feel free to continue conversation if you have further questions.

mlinksva:
BTW @waynebeaton since I see that you're on the Eclipse Foundation staff, I look forward to adding the next version of the EPL when it is eventually done; made me check the [mailing list](http://dev.eclipse.org/mhonarc/lists/epl-discuss/msg00070.html). Also very nice to see the new [contributor agreement](https://www.eclipse.org/legal/ECA.php) utilizing the DCO!

waynebeaton:
I cannot figure out how to even navigate to the EPL page on the site. Am I just not seeing it? Is there an actual link on the site?

The notion expressed in #386 that the EPL isn't a license that "anyone should choose unless they already know about them because they're in the relevant communities" is incorrect. I just did a quick search to compare hits of the EPL vs the AL and while there are certainly more hits against AL, the numbers are in the same order of magnitude. Lots of people who aren't building Eclipse plug-ins select the Eclipse Public License.

How do we appeal this decision?

mlinksva:
Eclipse and Clojure are pretty big communities, so I'm not surprised that EPL gets lots of hits. Why should someone who is just learning about open source licenses and not in one of those communities (and maybe others I don't know about) learn about EPL?

It isn't linked prominently but can be gotten to at http://choosealicense.com/appendix/ which is linked from the top of http://choosealicense.com/licenses/ and http://choosealicense.com/about/

Also see just added https://github.com/github/choosealicense.com/issues/413#issuecomment-243605410 we're considering how to make more licenses accessible from the drop-down.

Issue #413, opened 2016-05-14
Is there a reason why BSD was removed from the list of licenses


sandeepraju:
I see that the page here (http://choosealicense.com/licenses/) no more lists the BSD license. Is there any specific reason why it was removed?

mlinksva:
BSD-2 and 3-Clause as well as ISC were removed from /licenses/ in https://github.com/github/choosealicense.com/pull/386 because they aren't substantively different from MIT and more redundant choices adds cognitive overhead for someone learning about and choosing an open source license.

They're all still on the site linked from http://choosealicense.com/appendix/ and if you prefer one of them, please continue using it; they're all fine licenses.

I'm closing this but feel free to follow up if this doesn't make sense.

svenevs:
Would it be appropriate to bring back a choice of GPL v2 or v3?  My understanding (as a non-legal expert) is they are quite different.  Different enough to stir a enough banter online as to whether it should even be v3 instead of a different name...

Realistically, all of these licenses can be included manually with minimal effort.  I just feel that having a GPL v2 as well as GPL v3 option would at least inspire potential users to search online to compare the two.

mlinksva:
@svenevs inspiring users to search online and compare GPLv2 and GPLv3 falls squarely under "adds cognitive overhead" if anything does. People who want to try to discern the differences between these licenses from reading online banter are not in the main target audience for a site helping people to choose a license, perhaps for the first time and without lots of background knowledge.

For you or future searchers who do want to compare, I recommend [Chapter 9: GPL Version 3](http://copyleft.org/guide/comprehensive-gpl-guidech10.html#x13-610009) of _Copyleft and the GNU General Public License: A Comprehensive Tutorial and Guide_.

svenevs:
Well, when you put it that way I couldn't agree more x0

Thanks for sending that link along, that was a very well put together document.  Dense, but very thorough!

scopatz:
I am posting this here on [the suggestion](https://twitter.com/arfon/status/770408199270699008) of @arfon.

**Disclaimer:** This is not about the contents of the choosealicense.com website, which I veiw as disitnctly outside of my perogative.

[...]

PR #456, opened 2016-09-26
hidden: false for 'popular' licenses

mlinksva:
- document criteria for whether a license is hidden
  - needed for license spectrum on /licenses OR
  - on 'popular' list at https://opensource.org/licenses (some other list could be used in the future)
- adjust license properties and tests accordingly

This gets non-hidden list back close to what it was before #386 and (pending licensee vendoring this change, licensee release, and github.com licensee dependency version bump) some commonly requested licenses (eg #413 #449) will reappear in the github.com license drop-down.

cc @arfon

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ Licenses sit in the `/_licenses` folder. Each license has YAML front matter desc
 #### Optional fields
 
 * `featured` - Whether the license should be featured on the main page (defaults to false)
-* `hidden` - Whether the license is hidden from the license list (defaults to true)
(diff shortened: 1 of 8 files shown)


arfon (approved):
👍 Looks good to me.

waldyrious:
Potential additional lists to use are github-specific statistics like [this](https://github.com/blog/1964-open-source-license-usage-on-github-com) (maybe that very list could be used for now, since the proportions aren't likely to change too dramatically over time), and cross-host stats like [this](https://www.blackducksoftware.com/top-open-source-licenses) (maybe just the top 10).

Any thoughts about these?

mlinksva:
@waldyrious yes those sorts of stats, though I'd prefer ones that are based on public datasets; GitHub-specific ones could be obtained from the GitHub datasets in Google BigQuery, see  https://github.com/blog/2201-making-open-source-data-more-available

PR #459, opened 2016-10-10
Adding a link in the about page to FSF's advice on how to choose a license

anandtrex:
Adding a link in the about page to FSF's advice on how to choose a license. While it does strongly recommend using the GPL, I think it does a good job in explaining where using GPL may not be necessarily appropriate.


mlinksva:
Thanks for the suggestion! I combined in one bullet with the existing FSF license list.

PR #462, opened 2016-10-29
Update MIT license based on source

remarkablemark:
Source: https://opensource.org/licenses/MIT

### DIFF
--- a/_licenses/mit.txt
+++ b/_licenses/mit.txt
@@ -28,7 +28,7 @@ limitations:
 
 ---
 
-MIT License
+The MIT License (MIT)
 
 Copyright (c) [year] [fullname]
 


remarkablemark:
Hmm... Travis says that an [external link failed](https://travis-ci.org/github/choosealicense.com/builds/171650485#L364).

However, I'm able to see it in my browser: https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/COPYING

Should the test be rerun?

mlinksva:
@remarkablemark thanks for noticing ambiguity in the `source` field here -- our license texts don't all match precisely the texts the fields point to. I'd like to fix that eventually, probably by treating license texts curated by SPDX as canonical.

In the meantime for MIT I'd prefer to keep the current text, arrived at in https://github.com/github/choosealicense.com/pull/371 and agreeing with https://spdx.org/licenses/MIT.html

As I've [mentioned before](https://github.com/benbalter/licensee/issues/98#issuecomment-220885877) there are many variations on the MIT text. There's no reason to perpetuate a redundant title just because another prominent site does.

PR #467, opened 2016-12-12
no-liability -> liability, warranty

### LINKED ISSUE #412: "hold liable" found horribly confusing
https://twitter.com/richardfontana/status/731542765637652484

> I find "hold liable" horribly confusing

I'm not sure whether this comment is in reference to:
- The heading "Hold Liable" on http://choosealicense.com/appendix/
- The description currently displayed with a tooltip: "Software is provided without warranty and the software author/license owner cannot be held liable for damages."
- Implication that the concept makes for an inaccurate or otherwise inappropriate description of open source licenses.

Suggested fixes for any of the above that may be pertinent welcome.

If only the heading, adding a legend as proposed in #411 might help.

mlinksva:
rename and add tag and associated descriptions
 
fixes #412

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,14 +8,14 @@ gem 'github-pages', versions['github-pages']
 
 group :development do
   gem 'colored'
-  gem 'terminal-table'
   gem 'fuzzy_match'
+  gem 'terminal-table'
 end
 
 group :test do
   gem 'html-proofer', '~> 3.0'
+  gem 'nokogiri'
   gem 'rake'
   gem 'rspec'
-  gem 'nokogiri'
   gem 'rubocop'
 end
--- a/README.md
+++ b/README.md
@@ -95,8 +95,9 @@ The license properties (rules) are stored as a bulleted list within the licenses
 #### Limitations
 
 * `trademark-use` - This license explicitly states that it does NOT grant you trademark rights, even though licenses without such a statement probably do not grant you any implicit trademark rights.
-* `no-liability` - Software is provided without warranty and the software author/license owner cannot be held liable for damages.
+* `liability` - This license includes a limitation of liability.
 * `patent-use` -  This license explicitly states that it does NOT grant you any rights in the patents of contributors.
+* `warranty` - The license explicitly states that it does NOT provide any warranty.
 
 ## License
 
--- a/_data/rules.yml
+++ b/_data/rules.yml
@@ -36,9 +36,12 @@ limitations:
 - description: This license explicitly states that it does NOT grant you trademark rights, even though licenses without such a statement probably do not grant you any implicit trademark rights.
   label: Trademark Use
   tag: trademark-use
-- description: Software is provided without warranty and the software author/license owner cannot be held liable for damages.
-  label: Hold Liable
-  tag: no-liability
+- description: This license includes a limitation of liability.
+  label: Liability
+  tag: liability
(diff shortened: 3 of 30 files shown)


waldyrious:
For convenience, quoting https://github.com/github/choosealicense.com/issues/412#issuecomment-266566232:
> I thought almost all (except WTFPL) public copyright licenses have both [liability and warranty limitations], but searching the ones cataloged here, it seems MS-PL and MS-RL do not include a limitation on liability.

mlinksva:
Thanks @waldyrious indeed I should've included that info. I also should've cc'd @richardfontana @benbalter @lee-dohm

lee-dohm:
Looks good to me! :tada: 👍

benbalter:
> I thought almost all (except WTFPL) public copyright licenses have both [liability and warranty limitations], but searching the ones cataloged here, it seems MS-PL and MS-RL do not include a limitation on liability.

Why is it necessary to have distinct `warranty` and `liability` properties? Two reasons I ask:

1. Is it worth the added complexity to the vast majority of use cases for the incremental clarity provided for the (hidden) MP-* licenses? The fact that the MP-* licenses have a slightly weaker "don't sue me" clause (that covers contract, but not tort causes of action), seems too far in the weeds to make it worth adding the additional field to every other license. Perhaps it could be better served by a note in the sidebar clarifying the weaker protections?

2. The other limitations are `liability`, `copyright`, and `trademark`, and are in the negative, meaning you *can't* use my copyright, you *can't* use my trademark, and you *can't* hold me liable. In the case of `warranty`, it's that I've affirmatively disclaimed implied warranties (or that you can't bring a cause of action under one of those implied warranties, which again feels in the weeds for me... most would just call that `liability`, and if you need more specifics, you should get a lawyer to talk through your particular use case).

mlinksva:
It's not _necessary_ to have distinct `warranty` and `liability` properties. It's not necessary to have either of them. The only real outlier is WTFPL and that could be noted in its description. The MS-PL/RL difference is trivia, not a motivation for adding `warranty`; I only noticed in process of adding. In general we don't require variation among licenses in order to make something a tag. Indeed several permissions are mandatory in order for a license to be cataloged, ie they're the same for all licenses.

The reason for including any tag should be that it helps provide an accurate and clear description of licenses for licensors and licensees. Not necessarily a precise description, as that would entail getting into the weeds.

I agree that both limitation on liability and disclaimer of warranty serve same overall objective -- reduce potential costs to licensor and to some extent others. https://github.com/github/choosealicense.com/issues/412#issuecomment-266175475 made me think that level of abstraction might make it harder to be accurate, and harder to be clear. I guess that most people have a good intuition about what no warranty means, and I also guess that most people would _not_ intuitively think of a warranty as a liability, simply because they've never even considered the question.

Again, I don't consider either tag really necessary here. The benefit to including either is probably to inform reader that these are issues that maybe you wouldn't have thought about, but by using a standard open source license, they are covered. For that benefit, I guess that stating both `liability` and `warranty` is better than folding them together.

I don't feel particularly strongly, happy to close this without merging if above is nonsense. :smile:

mlinksva:
I'm going to merge this in the next few days if I don't get other feedback. I'm somewhat attracted to not modeling limitations at all (explicit tm and patent non-permissions could be modeled as such) but that would be a bigger change and I think there's some benefit to calling liability/warranty limitation/disclaimer as a feature of most open source licenses.

PR #470, opened 2016-12-22
add closed-source permission

### LINKED ISSUE #410: Difference between GPLv3 and LGPLv3 not modeled
"The table shows no difference between GPLv3 and LGPLv3?" - https://twitter.com/weigandtLabs/status/730676399678324736

It's true, we don't make any attempt to model the different scopes of derivative works that license conditions apply to. I don't know of any attempt to model these differences (pointers wanted). I've seen many times where the differences are important, people categorize licenses, often as "strong" or "weak" copyleft, but that's different from modeling the differences.

I'd planned to bring this up eventually as a use case for potentially adding some kind of qualifiers to our currently boolean (present or not) permissions, conditions, and limitations described in `_data/rules.yml`. But I'll take above question as opportunity to sketch out solutions to modeling the differences between GPLv3 and LGPLv3, and a bit more.

(1) qualified closed source derivatives permission

We could have a `closed-derivatives` permission, which [lax] permissive licenses (eg MIT) would have without qualification, and would be completely absent from strong copyleft (eg GPLv_) licenses. Qualifications could be something like `larger-works` (eg LGPLv_), `other-modules` (EPL-1.0), and `other-files` (MPL-2.0).

To minimize changes, the "qualifications" would actually be new fields described in `_data/rules.yml` with some naming convention, e.g., `closed-derivatives:larger-works`. Display code would recognize those different fields as qualifications and render in same place unqualified field would be rendered, probably accompanied by a yellow or yellow-green circle instead of green.

I think this could be useful to users: whether closed (alternatively, proprietary) derivatives are permitted is something people want to know, whether their aim is to prevent or allow closed derivatives, and a key difference among licenses, and choosealicense.com doesn't directly state this key difference. One has to ponder the implications of `same-license` to really get it.

Calling out whether proprietary derivatives are [dis]allowed mentioned in the talk referenced at https://github.com/github/choosealicense.com/issues/335#issuecomment-191952315 and license decisions being motivated by this question documented in paper referenced at https://github.com/github/choosealicense.com/issues/335#issuecomment-210152954

Just to state the obvious:
- `closed-derivatives` would be permission to distribute such; making such permitted by all open licenses; description could state that
- could directly qualify `modifications`; this would be somewhat confusing due to the point directly above, and might needlessly anger copyleft advocates (copyleft licenses would get a yellow bullet for that field).

(2) qualified same license condition

The existing `same-license` condition could be qualified. Probably strong copyleft and permissive licenses would see no change, but weak copyleft licenses would be qualified similarly to above, with "only" replacing "other", with the blue bullets getting lighter.

Relative to (1) this would be a smaller change and would make the model a bit more precise but not make the key difference among licenses mentioned above any more explicit.

(3) categories

Categorize to capture interesting features, but don't try to model those features. Display categories wherever license info is displayed, along with permission/condition/limitation fields. As mentioned above, this is an approach others have taken (e.g., "strong copyleft", "weak copyleft"). Categories could have [additional uses](https://github.com/github/choosealicense.com/issues/404#issuecomment-217277778).

Categories could even be used by display code to change the rendering of modeled fields -- achieving same display impact as (2).

(4) descriptions

No change. Don't attempt to model or display graphically differences  more subtle than those we already model and display. Mention those differences where possible in textual descriptions. #396 attempts to make sure those differences are described accurately in text, but needs work to be readable.

I don't plan to work on any of 1-3 soon but would enjoy feedback, other ideas, even speculative pull requests.

mlinksva:
Present for all licenses except strong copyleft ones (weak copyleft ones have this permission and same-license condition), attenuation left to description.

Fixes #410 (model l/gpl differences) while keeping completely flat/unqualified tags, simplified version of option (1) there.

Also exposes most practical licensing consideration (whether work can be used to make closed source applications), presently only mentioned in free-form license descriptions.

I'm not particularly committed to this, it is just the simplest option to to fix #410. Feedback wanted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ The license properties (rules) are stored as a bulleted list within the licenses
 
 #### Permissions
 
-* `closed-derivatives` - You may distribute proprietary software based on this software. Pay very close attention to license conditions, especially _same license_, if present, to understand when this permission applies.
+* `closed-derivatives` - You may distribute proprietary derivatives of this software. Pay very close attention to license conditions, especially _same license_, if present, to understand when this permission applies.
 * `commercial-use` - This software and derivatives may be used for commercial purposes.
 * `modifications` - This software may be modified.
 * `distribution` - You may distribute this software.
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ The license properties (rules) are stored as a bulleted list within the licenses
 
 #### Permissions
 
-* `closed-source` - You may distribute closed source software based on this software. Pay very close attention to license conditions, especially _same license_, if present, to understand when this permission applies.
+* `closed-derivatives` - You may distribute proprietary software based on this software. Pay very close attention to license conditions, especially _same license_, if present, to understand when this permission applies.
 * `commercial-use` - This software and derivatives may be used for commercial purposes.
 * `modifications` - This software may be modified.
 * `distribution` - You may distribute this software.
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ The license properties (rules) are stored as a bulleted list within the licenses
 * `document-changes` - Indicate changes made to the code.
 * `disclose-source` - Source code must be made available when distributing the software.
 * `network-use-disclose` - Users who interact with the software via network are given the right to receive a copy of the corresponding source code.
-* `same-license` - Modifications must be released under the same license when distributing the software. In some cases a similar or related license may be used or this condition may only apply to certain kinds of modificaitons.
+* `same-license` - Modifications must be released under the same license when distributing the software. In some cases a similar or related license may be used or this condition may only apply to certain kinds of modifications.
 
 #### Limitations
 
--- a/README.md
+++ b/README.md
@@ -78,6 +78,7 @@ The license properties (rules) are stored as a bulleted list within the licenses
 
 #### Permissions
 
+* `closed-source` - You may distribute closed source software based on this software. Pay very close attention to license conditions, especially _same license_, if present, to understand when this permission applies.
 * `commercial-use` - This software and derivatives may be used for commercial purposes.
 * `modifications` - This software may be modified.
 * `distribution` - You may distribute this software.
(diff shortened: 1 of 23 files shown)


_data/rules.yml
```diff
@@ -28,7 +31,7 @@ conditions:
 - description: Users who interact with the software via network are given the right to receive a copy of the corresponding source code.
   label: Network Use is Distribution
   tag: network-use-disclose
-- description: Modifications must be released under the same license when distributing the software. In some cases a similar or related license may be used.
+- description: In some cases a similar or related license may be used or this condition may only apply to certain kinds of modificaitons.
```
waldyrious: typo: "modificaitons" --> "modifications"

README.md
```diff
@@ -90,7 +91,7 @@ The license properties (rules) are stored as a bulleted list within the licenses
 * `document-changes` - Indicate changes made to the code.
 * `disclose-source` - Source code must be made available when distributing the software.
 * `network-use-disclose` - Users who interact with the software via network are given the right to receive a copy of the corresponding source code.
-* `same-license` - Modifications must be released under the same license when distributing the software. In some cases a similar or related license may be used.
+* `same-license` - Modifications must be released under the same license when distributing the software. In some cases a similar or related license may be used or this condition may only apply to certain kinds of modificaitons.
```
waldyrious: Maybe "In some cases *only* a similar or related license may be used"?

mlinksva: Adding sometimes only would make sense if this were called `similar-license`. I don't know what it adds given the name and rest of description now. Am I missing something?

waldyrious: No, you're right, it was me who had misread the entire description.

mlinksva:
Here's what this would look like (present for all licenses except AGPL, GPL and CC-BY-SA):

closed source tag

closed source tag label

Would love additional feedback.

mlinksva:
@waldyrious reading https://github.com/tldr-pages/tldr/issues/1076#issuecomment-270199712 ... CC-BY-SA is far weaker than GPL: former doesn't require modifiable form. It isn't a great fit for this schema. Though one shouldn't use CC-BY-SA for software, one could make openly licensed but without source (and thus not open source even in spirit, if we let not being an OSI-vetted license slide). Maybe `closed-source` should be `closed-derivatives`. Or maybe CC-BY-SA should get `closed-source`.

waldyrious:
Thanks for the clarifications. From a layman POV, `closed-derivatives` seems intuitive and appropriate.

mlinksva:
I'm not happy enough with this to merge it. Though there's some value in calling out closed derivatives, I'm not sure it also wouldn't be confusing and even misleading in some cases, and it's a bit redundant.

I'll open another PR just taking the slightly expanded description of same license here, and later a different solution for #410 that'll take more work to implement.

mlinksva:
I got happier with this, and not enthused about other options. Currently looks like below.

screen shot 2017-02-19 at 1 14 59 pm

mlinksva:
Alternatively see #484 which I am slightly more enthused about after all. :smile:

PR #472, opened 2017-01-03
Update ruby version in travis.yml

guiscaranse:
Nokogiri requires ruby >= 2.1.0 to be used, and fails to complete travis build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ script: "./script/cibuild"
 #environment
 language: ruby
 rvm:
-  - 2.1.0
+  - 2.3.3
 
 branches:
   only:
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ script: "./script/cibuild"
 #environment
 language: ruby
 rvm:
-  - 2.0.0
+  - 2.1.0
 
 branches:
   only:


guiscaranse:
`http://git.savannah.gnu.org/cgit/bash.git/tree/COPYING` was down while TravisCI was building, and caused the build to fail. The pull request build should be restarted when the `http://git.savannah.gnu.org/cgit/bash.git/tree/COPYING` get back online again

mlinksva:
Thanks. Looks like >= 2.1 is new in https://rubygems.org/gems/nokogiri/versions/1.7.0-x86-mingw32 released last week.

I'm not sure what best practice is for CI ruby versions. I see that https://pages.github.com/versions/ currently uses 2.3.3. Maybe @benbalter or @parkr can say for sure what the right thing to do is and merge this if it is already the right thing?

parkr:
Generally, if you are running CI against a web page which eventually is run on GitHub Pages, you should use the same version that GitHub Pages uses. I would recommend bumping this further to Ruby 2.3.3 to match GitHub Pages. 👍

mlinksva:
Thanks @guiscaranse and @parkr. I went ahead and bumped and will merge this now.

I suppose to be really complete we could have a test that grabs the version used by pages at the time the test runs, and fails if `.travis.yml` doesn't have the same version.

parkr:
You could certainly do that. You can use https://pages.github.com/versions.json if you want to write a program. A good idea to add to the Pages gem, actually.

mlinksva:
For the heck of it, #473

PR #475, opened 2017-01-16
Update apache.txt

xproax:
patent grant from permitted to required (contributors are required to give a patent grant to recipients)


mlinksva:
Thanks for opening this. I'd seen the very old branch but hadn't gotten around to deleting it.

It's true that contributors give a patent grant, but then it's also true that contributors grant a copyright license that includes all of the permissions. I don't think it's helpful to mirror all of these as conditions.

PR #476, opened 2017-01-17
Check images for consistent use of http/https

bkeepers:
https is now enforced on choosealicense.com.  This adds a test to ensure `img` tags don't generate mixed content warnings.

cc @gjtorikian

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -13,7 +13,7 @@ task :test do
   HTMLProofer.check_directory('./_site',
                               check_html: true,
                               validation: { ignore_script_embeds: true },
-                              url_swap: { %r{http://choosealicense.com} => '' },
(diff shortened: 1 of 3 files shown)


Rakefile
```diff
@@ -13,7 +13,8 @@ task :test do
   HTMLProofer.check_directory('./_site',
                               check_html: true,
                               validation: { ignore_script_embeds: true },
-                              url_swap: { %r{http://choosealicense.com} => '' }).run
+                              url_swap: { %r{http://choosealicense.com} => '' },
```
mlinksva: Doesn't seem to make any practical difference, but so it reflects what's deployed, should this (and `url` in `_config.yml`) be `https`?

PR #484, opened 2017-02-19
same-license--variations

### LINKED ISSUE #410: Difference between GPLv3 and LGPLv3 not modeled

mlinksva:
Adds 3 variations on `same-license`:

* `same-license--file` (MPL-2.0 and MS-RL)
* `same-license--library` (LGPL-2.1 and -3.0)
* ~~`same-license--source` (EPL-1.0)~~ (Not useful enough distinction in practice to include here; maybe [eventual](https://dev.eclipse.org/mhonarc/lists/epl-discuss/msg00112.html) EPL-2.0 will clearly fit with `same-license--file`)

This is an alternative to #470 for which also would fix #410 

This alternative doesn't highlight closed derivatives as a choice point, which is a plus (avoids redundancy, maybe some controversy and misunderstanding) and a minus (leaving the most important choice point for many licensing decisions implied).

I think I like this alternative better overall, in part because it conveys information about the variations in weaker copylefts that are hard to learn about, and avoids minus above.

The two alternatives aren't mutually exclusive, but I think if we go for this one I'd not feel any pull to implement the one in #410.

Screen shot of LGPL-3.0 with this one:

screen shot 2017-02-19 at 2 51 48 pm

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,6 @@ The license properties (rules) are stored as a bulleted list within the licenses
 * `same-license` - Modifications must be released under the same license when distributing the software. In some cases a similar or related license may be used.
 * `same-license--file` - Modifications of existing files must be released under the same license when distributing the software. In some cases a similar or related license may be used.
 * `same-license--library` - Modifications must be released under the same license when distributing the software. In some cases a similar or related license may be used, or this condition may not apply to works that use the software as a library.
-* `same-license--source` - Modifications must be released under the same license when distributing the software in source form.
 
 #### Limitations
 
--- a/README.md
+++ b/README.md
@@ -91,6 +91,9 @@ The license properties (rules) are stored as a bulleted list within the licenses
 * `disclose-source` - Source code must be made available when distributing the software.
 * `network-use-disclose` - Users who interact with the software via network are given the right to receive a copy of the corresponding source code.
 * `same-license` - Modifications must be released under the same license when distributing the software. In some cases a similar or related license may be used.
+* `same-license--file` - Modifications of existing files must be released under the same license when distributing the software. In some cases a similar or related license may be used.
+* `same-license--library` - Modifications must be released under the same license when distributing the software. In some cases a similar or related license may be used, or this condition may not apply to works that use the software as a library.
+* `same-license--source` - Modifications must be released under the same license when distributing the software in source form.
 
 #### Limitations
 
--- a/_data/rules.yml
+++ b/_data/rules.yml
@@ -31,6 +31,15 @@ conditions:
 - description: Modifications must be released under the same license when distributing the software. In some cases a similar or related license may be used.
   label: Same License
   tag: same-license
+- description: Modifications of existing files must be released under the same license when distributing the software. In some cases a similar or related license may be used.
+  label: Same License (File)
+  tag: same-license--file
+- description: Modifications must be released under the same license when distributing the software. In some cases a similar or related license may be used, or this condition may not apply to works that use the software as a library.
+  label: Same License (Library)
+  tag: same-license--library
+- description: Modifications must be released under the same license when distributing the software in source form.
+  label: Same License (Source)
+  tag: same-license--source
 
 limitations:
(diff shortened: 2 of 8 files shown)


README.md
```diff
@@ -91,6 +91,8 @@ The license properties (rules) are stored as a bulleted list within the licenses
 * `disclose-source` - Source code must be made available when distributing the software.
 * `network-use-disclose` - Users who interact with the software via network are given the right to receive a copy of the corresponding source code.
 * `same-license` - Modifications must be released under the same license when distributing the software. In some cases a similar or related license may be used.
+* `same-license--file` - Modifications of existing files must be released under the same license when distributing the software. In some cases a similar or related license may be used.
```
benbalter, line 94: Curious, why the double `-`s?

mlinksva: @benbalter to mark as refinement without having to loop through rules to check for substring, or having another field indicating refinement, or having knowledge of specific rules in software using the rules. I use this in the super-ugly appendix code. If `--` is itself super-ugly I can switch to `-` and use slightly different logic.

PR #486, opened 2017-03-03
Change copyright date 2004 -> [year] in wtfpl

jacobmischka:
I believe this one is supposed to be the year that the software under the license is published, not the year the license itself was written. But I'm not positive. If I'm wrong please feel free to close!

### DIFF
--- a/_licenses/wtfpl.txt
+++ b/_licenses/wtfpl.txt
@@ -22,7 +22,7 @@ limitations: []
             DO WHAT THE FUCK YOU WANT TO PUBLIC LICENSE
                     Version 2, December 2004
 
- Copyright (C) 2004 [fullname] <[email]>
+ Copyright (C) [year] [fullname] <[email]>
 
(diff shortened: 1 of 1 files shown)


mlinksva:
Good catch! But the year is actually right, it's the fill-in fields that are wrong. That part of the license is about the license text itself, see  http://www.wtfpl.net/faq/

jacobmischka:
I believe I fixed the correct one, can you please double check my commit?

jacobmischka:
Oh, you meant their existence is wrong? Hm.

jacobmischka:
Yeah I guess you're right, interesting.

mlinksva:
@jacobmischka Thanks a bunch for noticing this! It's really easy to confuse a copyright notice for the license text itself for a copyright notice for the licensed work, a previous instance in https://github.com/github/choosealicense.com/issues/310

PR #491, opened 2017-03-14
focus additional resources on choosing an open license, not lists of licenses

### LINKED ISSUE #490: Differentiation of licensing resources on /about/
From John Sullivan at FSF re https://choosealicense.com/about/:

> We're concerned that the OSI and FSF license lists are incorporated with unlike licensing
resources, and think that some differentiation would be good there -- maybe some statement that these are the two places that people generally look to for license evaluations.

cc @zacchiro

mlinksva:
Fixes #490 which was concerned that the additional resources list did not include all like items, I understand driven by concern that the SPDX license list includes non-open licenses. So does the FSF license list -- worse, it even recommends some.

This fix focuses the list on resources about choosing an open license, not resources listing a bunch of licenses. Fortunately both OSI and FSF have pages directly addressing choosing a license, neither of which recommends non-open licenses.

I've removed the SPDX and English Wikipedia lists.

### DIFF
--- a/about.md
+++ b/about.md
@@ -18,10 +18,8 @@ See our [appendix](/appendix) for a table of all of the licenses cataloged in th
 
 {: .bullets}
 
-* Open Source Initiative's [list of licenses](https://opensource.org/licenses/) approved as conforming to the [Open Source Definition](https://opensource.org/osd)
-* Free Software Foundation's [comments on various licenses](http://www.gnu.org/licenses/license-list.html) and [advice on how to choose a license](https://www.gnu.org/licenses/license-recommendations.en.html)
-* Linux Foundation's [SPDX License List](https://spdx.org/licenses/)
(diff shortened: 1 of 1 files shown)


waldyrious:
@mlinksva do you have any suggestions that would improve the Wikipedia entry? What do you envision as a resource that would be useful to be added to this list?

mlinksva:
@waldyrious not really, other than generic suggestion for Wikipedia comparison pages -- fill out all of the cells, probably add some relevant rows, remove some obscure ones...and in the glorious future, render from Wikidata queries.

I'm thinking about re-adding some of the list-focused (as opposed to choice-focused) resources as part of making our [criteria](https://github.com/github/choosealicense.com/blob/gh-pages/CONTRIBUTING.md#adding-a-license) apparent on the about page.

PR #489, opened 2017-03-05
WIP: Vendor license text from SPDX API

benbalter:
[Whitespace insensitive diff](https://github.com/github/choosealicense.com/pull/489/files?w=1)

PR #494, opened 2017-03-17
Add ECL-2.0 License

buckett:
The Educational Community License version 2.0 ("ECL") consists of the Apache 2.0 license, modified to change the scope of the patent grant in section 3 to be specific to the needs of the education communities using this license.

### DIFF
--- a/_licenses/ecl-2.0.txt
+++ b/_licenses/ecl-2.0.txt
@@ -2,8 +2,6 @@
 title: Educational Community License v2.0
 spdx-id: ECL-2.0
 source: https://opensource.org/licenses/ECL-2.0
-featured: false
-hidden: true
 
 description: The Educational Community License version 2.0 ("ECL") consists of the Apache 2.0 license, modified to change the scope of the patent grant in section 3 to be specific to the needs of the education communities using this license.
 
--- a/_licenses/ecl-2.0.txt
+++ b/_licenses/ecl-2.0.txt
@@ -1,0 +1,240 @@
+---
+title: Educational Community License v2.0
+spdx-id: ECL-2.0
+source: https://opensource.org/licenses/ECL-2.0
+featured: false
+hidden: true
+
+description: The Educational Community License version 2.0 ("ECL") consists of the Apache 2.0 license, modified to change the scope of the patent grant in section 3 to be specific to the needs of the education communities using this license.
+
+how: Create a text file (typically named LICENSE or LICENSE.txt) in the root of your source code and copy the text of the license into the file.
+
+note: The Apereo Foundation recommends taking the additional step of adding a boilerplate notice to the header of each source file. You can find the notice at the very end of the license in the appendix.
+
+using:
+  - Sakai: https://github.com/sakaiproject/sakai/blob/master/LICENSE
+  - OAE: https://github.com/oaeproject/Hilary/blob/master/LICENSE
+  - Opencast Matterhorn: https://bitbucket.org/opencast-community/matterhorn/src/e5b1684a822c0bd9bb07f3e53b4e1f22932da5ef/LICENSE?fileviewer=file-view-default
+
+permissions:
+  - commercial-use
+  - modifications
+  - distribution
+  - patent-use
+  - private-use
+
+conditions:
+  - include-copyright
+  - document-changes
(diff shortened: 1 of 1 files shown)


buckett:
There's some background on the license on it's Wikipedia page: https://en.wikipedia.org/wiki/Educational_Community_License

benbalter:
@buckett Does the ECL license meet [the criteria for adding a new license](https://github.com/github/choosealicense.com/blob/gh-pages/CONTRIBUTING.md#adding-a-license)?

buckett:
1. The license must have an SPDX identifier. If your license isn't registered with SPDX, please request that it be added. **(yep, ECL-2.0)**
1. The license must be listed on one of the following approved lists of licenses:
    1. List of OSI approved licenses **yep on** https://opensource.org/licenses/alphabetical
    1. GNU's list of free licenses (note: the license must be listed in one of the three "free" categories)
    1. Open Definition's list of conformant licenses (non-code)
1. A GitHub code search must reveal at least 1,000 public repositories using the license  **Possibly** https://github.com/search?p=3&q=ECL-2.0+filename%3ALICENSE&ref=searchresults&type=Code&utf8=%E2%9C%93 although there are 365 forks of the Sakai project codebase, does this count as one or 365?

What spurred me on in suggesting adding this is that GitHub is wrongly marking projects with the ECL license as being Apache 2, eg: https://github.com/ox-it/wl-sakai

inoas:
Maybe the parser should look for something like "The MIT License (MIT)" and just take that **in first line** to identify a license _**instead**_ of trying to be intelligent.

https://github.com/cakephp/cakephp/pull/10430/files#diff-61e0bdf7e1b43c5c93d9488b22e04170L3 for instance is a problem that stopped github to detect the obvious existence of MIT, whereas for you/wl-sakai it assumed Apache V2 but it was ECL.

_licenses/ecl-2.0.txt
```diff
@@ -0,0 +1,238 @@
...
+  - include-copyright
+  - document-changes
+
+limitations:
+  - trademark-use
+  - liability
+  - warranty
+
+---
+Educational Community License, Version 2.0 (ECL-2.0)
```
mlinksva: I don't think this line is supposed to be part of the license text; it's fully redundant with the next lines and not included in https://github.com/spdx/license-list/blob/788e54b566dc78358e3e1c3f9b7187ad4005114e/ECL-2.0.txt

buckett: Yeah, the duplication seems redundant, however that's what on https://opensource.org/licenses/ECL-2.0 so I didn't want to create another variant of the ECL.

NB: Sadly the plain text version doesn't seem to be working any more: https://opensource.org/licenses/ecl2.txt

mlinksva: The OSI site is kind of ambiguous as to whether page headings are part of license texts. SPDX seems to be more careful, and eventually we will [source texts from them](https://github.com/github/choosealicense.com/pull/489), so I'd prefer text we import to be as close to them to start with.

buckett: Done.

buckett:
Side node: I did change the [..] to {..} in the header template, as the tests were failing without that change.

mlinksva:
@buckett thanks for submitting ECL-2.0. It may not quite get to 1k repos, but we don't have a perfect way of counting, so it's probably close enough, and I know Sakai is a giant project.

I'll leave this PR open over the weekend to give others a chance to notice any problems with the description and merge next week if none.

The project you note is mis-identified as Apache-2.0 was detected before [licensee](https://github.com/benbalter/licensee) got even stricter about rejecting potential false positives; I'm pretty sure the license would not be detected today. For any wrongly detected license on GitHub, feel free to [contact support](https://github.com/contact) and ask them to try re-detection.

@inoas just looking for a license name in the first line of a `LICENSE` file would generate too many false positives; a text file can contain all kinds of things, such as multiple licenses, and often does.

In any case, this (choosealicense.com) project just curates license descriptions, it isn't the place to debate license detection strategies.

inoas:
I do agree that false positives are a lot worse than false negatives. Thanks for explaining.

PR #497, opened 2017-03-24
Adding the University of Illinois/NCSA License 

narendasan:
This PR should add the University of Illinois/NCSA License into the set of licenses on choosealicense.com. Its a license that combines part of the MIT and BSD-3 Clause licenses and used in projects like Clang and LLVM (and many of my own projects).

License Requirements: 
1. [SPDX Identifier](https://spdx.org/licenses/NCSA.html): NCSA
2. Inclusion in both [OSI Approved License List](https://opensource.org/licenses/NCSA) and [GNU's list of GPL Compatible licenses](https://www.gnu.org/licenses/license-list.en.html#NCSA) 
3. [Github Search](https://github.com/search?utf8=%E2%9C%93&q=+%22University+of+Illinois%22+%22Open+Source%22+filename%3ALICENSE+%22NCSA%22&type=Code) returns 13K results (some duplicated repos)

Question about the metadata though, for limitations does trademark refer to only prohibiting the use of trademarks associated with the project or is it supposed to cover limitations on usage of developer and organization names as well? 

If it doesn't is that a limitation worth noting as its is a major condition of not only this license but also other BSD-3 Clause-like licenses?

### DIFF
--- a/_licenses/ncsa.txt
+++ b/_licenses/ncsa.txt
@@ -23,7 +23,6 @@ conditions:
   - include-copyright
 
 limitations:
-  - trademark-use
   - liability
   - warranty
 
--- a/_licenses/ncsa.txt
+++ b/_licenses/ncsa.txt
@@ -6,7 +6,7 @@ source: https://opensource.org/licenses/NCSA
 hidden: false
 featured: false
 
-description: The University of Illinois/NCSA Open Source License, or UIUC license, is a permissive free software license, based on the <a href="/license/mit/">MIT/X11 license</a>  and the <a href="/license/bsd-3-clause">BSD 3-clause License</a>. By combining parts of these two licenses, it attempts to be clearer and more concise than either. It's conditions include requiring the preservation of copyright and license notices both in source and in binary distributions and the prohibtion of using the names of the authors or the project organization to promote or endorse derived products.
+description: The University of Illinois/NCSA Open Source License, or UIUC license, is a permissive free software license, based on the <a href="/licenses/mit/">MIT/X11 license</a>  and the <a href="/licenses/bsd-3-clause/">BSD 3-clause License</a>. By combining parts of these two licenses, it attempts to be clearer and more concise than either. It's conditions include requiring the preservation of copyright and license notices both in source and in binary distributions and the prohibtion of using the names of the authors or the project organization to promote or endorse derived products.
 
 how: Create a text file (typically named LICENSE or LICENSE.txt) in the root of your source code and copy the text of the license into the file. Replace [year] with the current year and [fullname] with the name (or names) of the copyright holders. Replace [project] with the project organization, if any, that sponsors this work.
 
--- a/_licenses/ncsa.txt
+++ b/_licenses/ncsa.txt
@@ -6,7 +6,7 @@ source: https://opensource.org/licenses/NCSA
 hidden: false
 featured: false
 
-description: The University of Illinois/NCSA Open Source License, or UIUC license, is a permissive free software license, based on the <a href="/license/mit/">MIT/X11 license</a>  and the <a href="/license/bsd-3-clause">BSD 3-clause License</a>. By combining parts of these two licenses, it attempts to be clearer and more concise than either.
+description: The University of Illinois/NCSA Open Source License, or UIUC license, is a permissive free software license, based on the <a href="/license/mit/">MIT/X11 license</a>  and the <a href="/license/bsd-3-clause">BSD 3-clause License</a>. By combining parts of these two licenses, it attempts to be clearer and more concise than either. It's conditions include requiring the preservation of copyright and license notices both in source and in binary distributions and the prohibtion of using the names of the authors or the project organization to promote or endorse derived products.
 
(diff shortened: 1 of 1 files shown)


_licenses/ncsa.txt
```diff
@@ -0,0 +1,65 @@
+---
+title: University of Illinois/NCSA Open Source License
+spdx-id: NCSA
+nickname: UIUC/NCSA
+source: https://opensource.org/licenses/NCSA
+
+description: The University of Illinois/NCSA Open Source License, or UIUC license, is a permissive free software license, based on the <a href="/licenses/mit/">MIT/X11 license</a>  and the <a href="/licenses/bsd-3-clause/">BSD 3-clause License</a>. By combining parts of these two licenses, it attempts to be clearer and more concise than either. It's conditions include requiring the preservation of copyright and license notices both in source and in binary distributions and the prohibtion of using the names of the authors or the project organization to promote or endorse derived products.
```
mlinksva: Suggest removing the "By combining...' sentence, which appears to be copied from https://en.wikipedia.org/wiki/University_of_Illinois/NCSA_Open_Source_License ... both claims are somewhat questionable, and the citation on Wikipedia doesn't support them.

"It's" should be "Its".

_licenses/ncsa.txt
```diff
@@ -0,0 +1,65 @@
...
+
+how: Create a text file (typically named LICENSE or LICENSE.txt) in the root of your source code and copy the text of the license into the file. Replace [year] with the current year and [fullname] with the name (or names) of the copyright holders. Replace [project] with the project organization, if any, that sponsors this work.
+
+using:
```
mlinksva: Are there any good examples not from the llvm-mirror org? The first two of these 3 contain extra text that makes them not great as examples for a new project to start from.

_licenses/ncsa.txt
```diff
@@ -0,0 +1,65 @@
...
+  - liability
+  - warranty
+
+---
+
+University of Illinois/NCSA Open Source License 
+
+Copyright (c) [year], [fullname] 
+All rights reserved.
+
```
mlinksva: Please make the punctuation and newlines match https://github.com/spdx/license-list/blob/master/NCSA.txt (they don't above, also check the rest of the file)

_licenses/ncsa.txt
```diff
@@ -0,0 +1,65 @@
...
+Developed by: [fullname]
+              [project]
+              [email]
```
mlinksva: I guess `email` is best that can be done with fields available, but a URL is clearly the practice here for this license. I wonder if we should/could make `project_url` or similar available (the github.com interface would probably substitute with the repo URL)? cc @benbalter

mlinksva (changes requested):

_licenses/ncsa.txt
narendasan: The NCSA License does take the license grant and the warranty disclaimer from the MIT License (the text is nearly identical), and it does take inspiration from the BSD licenses requirements. Not sure if they modified the original BSD license and came out with a similar out come to BSD-3 clause or took the BSD 3-clause requirements, not well informed about the history here.

mlinksva: Yes, "based on" seems to be true. I'm only suggesting removing "By combining parts of these two licenses, it attempts to be clearer and more concise than either."

narendasan: ok

mlinksva:
@narendasan thanks for submitting this license. All of the fill-in fields make it kind of a pain. 😅 

I know I said to follow the SPDX record's formatting, but I think they're wrong about how the "Developed by" list is formatted, just filed an issue at https://github.com/spdx/license-list/issues/9

I welcome other feedback on what's the right thing to do on this for now, and also on the URL element of the same list, as noted at https://github.com/github/choosealicense.com/pull/497#discussion_r107978668

narendasan:
Should I change the formatting back to what OSI had?

mlinksva:
Yes, for the "Developed by" list.

Also, I think it makes sense to replace `[email]` with `[project_url]` even if github.com doesn't support auto-filling it. An email address just isn't what others using the license do, and I'd hate to start/spread another idiosyncratic way of deploying this license.

After those changes I'll leave open a few days before merging in case others have feedback.

narendasan:
Also because for BSD 3-Clause, this license, licenses for projects like curl as well as many open source licenses customized by companies make explicit limitations on endorsement, would it be worth having a ```non-endorsement``` tag in the set of limitations? It's a main differentiator from a license like MIT. TLDRLegal throws non-endorsement and trademark limitations into a single ```use trademark``` tag so that is another option.

mlinksva:
@narendasan probably not. If anything, I'd consider removing the `trademark-use` limitation tag. Licenses that don't say anything about trademark probably shouldn't be assumed to give permission to use trademarks or anyone's endorsement. For a good talk on this, see https://archive.fosdem.org/2016/schedule/event/trademark_not_implied/ (previously mentioned at https://github.com/github/choosealicense.com/pull/392#issuecomment-209554190).

mlinksva (approved):

PR #506, opened 2017-04-22
Example project moved its license file, update

mlinksva:


waldyrious:
x-ref: https://github.com/openstreetmap/iD/pull/3961

Issue #508, opened 2017-04-24
Add 3 nontrivial and regular/detectable uses to criteria for cataloging a license


mlinksva:
Extracting from https://github.com/github/choosealicense.com/issues/498#issuecomment-296753533

> I suspect we should add producing 3 solid (not trivial or totally unknown projects) and straightforward (using just the license text, that could be detected) examples to the required criteria. If there's no evidence that a license is used in a regular enough way that it could be detected then there's less value to adding it here, and I wouldn't even be confident that the use we'd be promoting here would be aligned with whatever the idiosyncratic way the license is used in the wild.

Current criteria are at https://github.com/github/choosealicense.com/blob/b356254902142b07e897c31cdf505c6ca8dbcbcd/CONTRIBUTING.md#adding-a-license in case anyone wants to propose wording to add above in a comment or a PR.

Note related documentation of `using` field at https://github.com/github/choosealicense.com#optional-fields which could be reused or referenced by added criterion.

PR #438, opened 2016-06-29
zlib license

pchaigno:
This pull request adds the Zlib license to choosealicense.com. It was first used by the zlib project (see the end of [its README](https://github.com/madler/zlib/blob/master/README)). It's now used by many other projects (e.g., certain syntax definitions for programming languages).

Requirements:
1. `Zlib` is the [SPDX identifier](https://spdx.org/licenses/Zlib.html).
2. Zlib is [part of the list of OSI approved licenses](https://opensource.org/licenses/Zlib) and [GNU's list of free licenses](https://www.gnu.org/licenses/license-list.en.html#ZLib).
3. [This GitHub search result](https://github.com/search?utf8=%E2%9C%93&q=misrepresented+%22Altered+versions%22+filename%3ALICENSE+%22commercial+applications%22&type=Code&ref=searchresults) displays ~2,830 Zlib licenses. I downloaded 2,534 of them, from which I counted 1,580 repositories (from 1,439 different users).

PR #511, opened 2017-05-27
Add BSD-4-Clause license

okeuday:
* Fixes https://github.com/benbalter/licensee/issues/195

The requirements are met:
1. BSD-4-Clause
1. https://www.gnu.org/licenses/license-list.en.html#GPLIncompatibleLicenses lists it (2nd free category)
1. https://github.com/search?q=BSD+filename%3ALICENSE&type=Code (difficult to distinguish between the different BSD types, the BSD-4-Clause is commonly used to ensure credit is received for open-source work that may be used commercially)

### DIFF
--- a/_licenses/bsd-4-clause.txt
+++ b/_licenses/bsd-4-clause.txt
@@ -1,0 +1,58 @@
+---
+title: BSD 4-clause "Original" or "Old" License
+spdx-id: BSD-4-Clause
+source: https://directory.fsf.org/wiki/License:BSD_4Clause
+hidden: true
+
+description: A permissive license similar to the <a href="/licenses/bsd-3-clause/">BSD 3-Clause License</a>, but with a 4th clause (known as the "advertising clause") that requires an acknowledgment of the original source in all advertising material.
+
+how: Create a text file (typically named LICENSE or LICENSE.txt) in the root of your source code and copy the text of the license into the file. Replace [year] with the current year and [fullname] with the name (or names) of the copyright holders. Replace [project] with the project organization, if any, that sponsors this work.
+
+permissions:
+  - commercial-use
+  - modifications
+  - distribution
+  - private-use
+
+conditions:
+  - include-copyright
+
+limitations:
+  - liability
+  - warranty
+
+---
+
+BSD 4-Clause License
+
+Copyright (c) [year], [fullname]
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+* Redistributions of source code must retain the above copyright notice, this
+  list of conditions and the following disclaimer.
+
+* Redistributions in binary form must reproduce the above copyright notice,
+  this list of conditions and the following disclaimer in the documentation
+  and/or other materials provided with the distribution.
+
+* All advertising materials mentioning features or use of this software must
+  display the following acknowledgement:
+    This product includes software developed by [project].
+
+* Neither the name of the copyright holder nor the names of its
+  contributors may be used to endorse or promote products derived from
+  this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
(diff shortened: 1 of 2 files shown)


okeuday:
The CI build fails due to wanting to use ruby 2.4.0 instead of 2.3.3, though it has test failures if it is changed to 2.4.0.  So, this failure should be expected in the master branch and is not related to these changes.

mlinksva:
Thanks @okeuday ! I fixed the failures.

https://github.com/search?utf8=%E2%9C%93&q=%22All+advertising+materials+mentioning+features+or+use+of+this+software+must+display+the+following+acknowledgement%22+filename%3ALICENSE&type=Code is more specific but still qualifies.

2 things:

- There are some differences between this text and https://spdx.org/licenses/BSD-4-Clause.html (numbered bullets, words in disclaimer). I'd prefer to go with SPDX version unless there's evidence they have it wrong because an eventual goal is to source texts from them.
- Related to figuring out whether the SPDX version is correct, 3 `using` examples would be helpful (does use in the wild match?); this will be required soon, see https://github.com/github/choosealicense.com/issues/508 I just haven't gotten around to writing the tests.

Could you update the PR with examples and license text matching the SPDX text or a reason that is wrong?

okeuday:
@mlinksva The BSD-4-Clause was used in NetBSD until 2008 (OpenBSD and FreeBSD changed their license around that time too), but despite that change, the older BSD licensed source code has found itself info other projects like ReactOS (https://www.reactos.org/) and uClinux (http://www.uclinux.org/).  These projects don't appear to use github for their repositories, though their source code has found its way into the github search due to other people creating repositories with the source code stuck at a certain point in the past.  I have used it in CloudI (http://cloudi.org) which does have a github repo at https://github.com/CloudI/CloudI/ but I plan on changing it, after having read more about it.  This is difficult to check, since you may unintentionally limit yourself to github or usage of git.

I updated the license to match the SPDX version.  I am not sure about the best github search to find "nontrivial and regular/detectable uses" (to satisfy https://github.com/github/choosealicense.com/issues/508) but I would like to think the information above is enough.

To some degree I would hope that GitHub isn't attempting to provide legal advice with its decisions, since attempting to provide legal advice without being a lawyer should be a problem (in the US at least).  To avoid that problem, it seems simplest to support any license and just rely on popularity to show what is best (natural selection).

mlinksva:
GitHub isn't attempting to provide legal advice, see the site disclaimer. 'Natural selection' is fine, but there are a lot of licenses, and for older ones especially, even defining what counts as an instance of a particular license is [fuzzy](https://github.com/benbalter/licensee/issues/98#issuecomment-220885877). So we defer as much as possible to external authorities and evidence of a license's use 'in the wild'. Note `using` examples do not need to be on GitHub or use git; their `LICENSE` file just needs to have a URL.

okeuday:
@mlinksva I am no longer interested in pursuing this.  Feel free to close the issue or to pursue it separately.

mlinksva:
@okeuday appreciated.

If anyone else reading now or later wishes to follow up on this, please do.

waldyrious:
I tried to find projects on GitHub licensed under BSD-4-Clause via Google BigQuery's GitHub tables, using the same method described [here](https://github.com/github/choosealicense.com/pull/438#issuecomment-230150424), but the query returned no results :/

``` sql
SELECT gh_licenses.repo_name, gh_since2015.num_stars
FROM [bigquery-public-data:github_repos.licenses] AS gh_licenses
JOIN (
    SELECT repo.name, COUNT(*) AS num_stars
    FROM TABLE_DATE_RANGE([githubarchive:day.], TIMESTAMP('2015-01-01'), CURRENT_TIMESTAMP())
    WHERE type = "WatchEvent"
    GROUP BY repo.name
) AS gh_since2015
ON (gh_licenses.repo_name == gh_since2015.repo.name)
WHERE license="bsd-4-clause"
ORDER BY num_stars DESC
```

Perhaps GitHub's new GraphQL API could eventually help here, but I couldn't figure out a way to search for repos by license (in any case, I suppose it would only work after BSD-4-Clause is added to choosealicense.com anyway)

waldyrious:
FYI, the BigQuery method does work for other licenses -- here are the ten most starred repos with the bsd-3-clause license:

- [FreeCodeCamp/FreeCodeCamp](https://github.com/FreeCodeCamp/FreeCodeCamp)
- [facebook/react](https://github.com/facebook/react)
- [facebook/react-native](https://github.com/facebook/react-native)
- [facebookincubator/create-react-app](https://github.com/facebookincubator/create-react-app)
- [golang/go](https://github.com/golang/go)
- [d3/d3](https://github.com/d3/d3)
- [django/django](https://github.com/django/django)
- [ecomfe/echarts](https://github.com/ecomfe/echarts)
- [facebook/immutable-js](https://github.com/facebook/immutable-js)
- [scikit-learn/scikit-learn](https://github.com/scikit-learn/scikit-learn)

mlinksva:
@waldyrious nice. If you feel like it, please make a PR to add example uses to BSD-3-Clause. Lots of good choices there, but offhandedly react, go, and django nicely demonstrate the apparent prestige of the license.

mlinksva:
Closing this as nobody has expressed interest in following up, including me. :smile:

Issue #498, opened 2017-03-26
Add Vim License


othree:
Vim uses a special GNU GPL [compatible][1] license called: [Vim License][2].
There are many Vim scripts [use][3] [the][4] [same][5] license.
I think it will be great if you can provide (and detect in GitHub).

[1]:https://github.com/vim/vim/blob/master/runtime/doc/uganda.txt#L115-117
[2]:https://github.com/vim/vim/blob/master/runtime/doc/uganda.txt#L24-L105

[3]:https://github.com/tpope/vim-pathogen#license
[4]:https://github.com/mileszs/ack.vim/blob/master/LICENSE
[5]:https://github.com/kien/ctrlp.vim/blob/master/doc/ctrlp.txt#L1262

mlinksva:
@othree, thanks for the suggestion.

Deployment of this license seems highly variable, with lots of text added before and after in examples I've seen -- these wouldn't be detected. Can you verify that it meets our [criteria](https://github.com/github/choosealicense.com/blob/gh-pages/CONTRIBUTING.md#adding-a-license) for cataloging a license here, and point out some examples that use only the [license text](https://spdx.org/licenses/Vim.html#licenseText) in a `LICENSE` file such that they could be detected?

If so feel free to make a pull request to add the license.

othree:
Hi

Sorry I almost forgot about this issue.
I have checked the criteria, yes Vim fits:
* SPDX Identifier: Vim
* GNU's free licenses <https://www.gnu.org/licenses/license-list.en.html#Vim>
* More than 1000 results 
    * <https://github.com/search?utf8=%E2%9C%93&q=%22vim+license%22++filename%3ALICENSE&type=Code>
    * <https://github.com/search?utf8=%E2%9C%93&q=%22same+as+vim%22&type=Code>

I think its not possible to support all examples I provided.
But we can start it from LICENSE file.
Maybe we can also detects [simple type][s]?

[s]:https://github.com/dracorp/aur-packages/blob/e2c03dcccd14209f1d2190e66dd040be7488b27c/vim/vim-cecutil/LICENSE

mlinksva:
@othree thanks for that info. On the "simple type" no, that wouldn't be detected; licensee only matches full copies of licenses (and optionally license IDs in package files, but not brief statements in a `LICENSE` file in any case).

[...]

PR #514, opened 2017-05-30
Require 3 detectable examples for newly added licenses

### LINKED ISSUE #508: Add 3 nontrivial and regular/detectable uses to criteria for cataloging a license

mlinksva:
`using:` examples are documented as optional, but going forward I'd like to make them mandatory. Almost any license proposed now will be somewhat obscure, as we have all the common ones. This additional requirement ensures that anything we add has examples 'in the wild' that could be detected. This came up a couple times recently: #498 #511 

Fixes #508

Note there are 20 licenses marked as 'legacy' which test for 3 examples skips. I'd be happy to see examples added to those over time.

Note absolutely no testing is done to ensure that 3 examples could be detected (it'd be possible to rig up a test using licensee and the licenses cataloged in a branch) or even are URLs. This seems good enough as a first pass.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,6 +29,7 @@ If your proposed license meets the above criteria, here's a few other things to 
 * The license files should be in the format of `_licenses/[lowercased-spdx-id].txt` (e.g., `_licenses/mit.txt`)
 * Each license has both [required and optional metadata](https://github.com/github/choosealicense.com#license-metadata) that should be included.
 * The text of the license should be wrapped to a 78 character width.
+* The text of the license should match the corresponding text found at https://spdx.org/licenses/
 * The body of the file should be the text of the license in plain text.
 
 ## Making Changes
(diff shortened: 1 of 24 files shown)


benbalter (approved):
Nice idea! (and well implemented)

PR #522, opened 2017-06-20
clarify what open license for documentation means

### LINKED ISSUE #521: Documentation section of Non-Software Licenses has no links.
https://choosealicense.com/non-software/#documentation

It states "Any open source software license or open Creative Commons license", however it offers no assistance in finding those.  Technically a couple of Creative Commons licenses were mentioned in the previous section (data/media), but only by short name (ex. CC-BY-4.0) so uninformed users likely won't even know they are Creative Commons licenses.  The result is that it is more effective to google "Creative Commons" at that point if you want to find licenses because there is no quick way to the [Appendix](https://choosealicense.com/appendix/) which is the only place to get the information.

I think either more links or license details would be the solution, but I am not sure what the best implementation would be.

mlinksva:
Fix #521

### DIFF
--- a/non-software.md
+++ b/non-software.md
@@ -12,7 +12,7 @@ Open source software licenses can be used for non-software works, and often are 
 
 ### Documentation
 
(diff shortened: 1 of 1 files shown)


adamvoss (approved):
Works for me :+1:

PR #524, opened 2017-06-23
New PostgreSQL license

dlecan:


_licenses/postgresql.txt
```diff
@@ -0,0 +1,50 @@
...
+conditions:
+  - include-copyright
+
+limitations:
+  - liability
+  - warranty
+
+---
+
+The PostgreSQL Licence (PostgreSQL)
```
mlinksva: Maybe this should just be "PostgreSQL License"?

The headings on license pages at opensource.org aren't super reliable indicators of what should be included in the license text. Some related discussion for similar licenses at https://github.com/github/choosealicense.com/pull/371

_licenses/postgresql.txt
```diff
@@ -0,0 +1,50 @@
...
+featured: false
+hidden: true
+
+description: A very short, BSD-style license, used specifically for PostgreSQL.  
+
+how: To use it, say that it is The PostgreSQL License, and then substitute the copyright year and name of the copyright holder into the body of the license. Then put the license into a prominent file ("COPYRIGHT", "LICENSE" or "COPYING" are common names for this file) in your software distribution.
+
+using:
+  - PostgreSQL: https://www.postgresql.org/about/licence/
+  - PGAudit: https://github.com/pgaudit/pgaudit/blob/master/LICENSE
```
mlinksva: This is not an ideal example. It does not contain the license text and would not be matched by [licensee](https://github.com/benbalter/licensee).

_licenses/postgresql.txt
```diff
@@ -0,0 +1,50 @@
...
+The PostgreSQL Licence (PostgreSQL)
+
+Copyright (c) [year] [fullname]
+
+Permission to use, copy, modify, and distribute this software and its
+documentation for any purpose, without fee, and without a written agreement is
+hereby granted, provided that the above copyright notice and this paragraph
+and the following two paragraphs appear in all copies.
+
+IN NO EVENT SHALL [fullname] BE LIABLE TO ANY PARTY FOR DIRECT, INDIRECT,
```
mlinksva, line 39: Is this (and fullname in next 2 paragraphs) typically replaced in the wild? Old versions of similar licenses were typically templated throughout, but generic terms (like "copyright holder") are now typically found in the license body.

_licenses/postgresql.txt
dlecan: Ok

_licenses/postgresql.txt
dlecan: Ok, I will provide better examples.

_licenses/postgresql.txt
dlecan: The new examples are compatible with the template. But we can find many projects with Postgresql license which aren't.

mlinksva: How do the many projects with the Postgresql license that don't match the template vary from the template? By just not including the license text at all (like one removed example)? Or by not treating the fields I mention above as fill-ins? Or something else? I just want to be sure the text we catalog here represents modern best practice. Thanks!

dlecan: Some projects:
1. mention "Postgresql Licen**c**e" (not licen**s**e) and add a simple link to the license, such as  [pgAudit](https://github.com/pgaudit/pgaudit_analyze/blob/master/LICENSE)
1. don't mention "Postgresql" at all and provide the full license with "fullname" substitution, such as [pgBadger](https://github.com/dalibo/pgbadger/blob/master/LICENSE) or [pgAdmin4](https://github.com/postgres/pgadmin4/blob/master/LICENSE)

With Licensee, I tried different license declarations to try to match those situations: only pgBadger was recognized :-/

Notice that removing the `[fullname]` field improve the Licensee's Confidence rate, but only pgBadger is still recognized.

Honestly, I don't know what to do.

mlinksva: Fedora generally has up to date licensing info, so I take from  https://fedoraproject.org/wiki/Licensing/PostgreSQL_License that fill-in fields are expected. licensee doesn't know about templated fields in a license body, so it might not do too well, but we'll have to live with that for now. Thanks for helping dig into this!

mlinksva:
https://github.com/search?utf8=%E2%9C%93&q=%22provided+that+the+above+copyright+notice+and+this+paragraph+and+the+following+two+paragraphs+appear+in+all+copies.%22+filename%3ALICENSE&type= obtains 1001 matches.

I think all the [criteria](https://github.com/github/choosealicense.com/blob/gh-pages/CONTRIBUTING.md#adding-a-license) for adding this license are met.

I'll merge this tomorrow if nobody points out any other problems with it.

PR #526, opened 2017-07-03
Add a reminder for updating year information and name of the copyright holder

lvscar:
Remind user update year information and name of the copyright holder while adopting Apache license.


mlinksva:
As mentioned in #417 the boilerplate at the end of the Apache-2.0 license text is not intended to be modified in the `LICENSE` file itself. See the repos under https://github.com/apache for as authoritative examples as possible, eg https://github.com/apache/hadoop/blob/trunk/LICENSE.txt -- the appendix is untouched.

lvscar:
Thanks for your explanation

PR #527, opened 2017-07-04
Added CeCILL Licenses v2.1

ljoubert:
Hello,

I propose to add the French CeCILL v2.1 license.

Best regards,

Laurent


_licenses/cecill-2.1.txt
```diff
@@ -0,0 +1,281 @@
+---
+title: CeCILL Free Software License Agreement v2.1
+spdx-id: CECILL-2.1
+source: http://www.cecill.info/licences/Licence_CeCILL_V2.1-fr.html
+
+description: French open-source license made by the CEA, INRIA and CNRS, compatible with the GNU GPL v2.0 and v3.0
+
+how: Create a text file (typically named LICENSE or LICENSE.txt) in the root of your source code and copy the text of the license into the file. 
+
+using:
```
mlinksva, line 9: These aren't ideal examples. The first is a license file, but due to multiple licenses and license files, wouldn't be automatically recognized. The second and third just state that the projects are licensed on the project websites. See https://github.com/github/choosealicense.com/blob/gh-pages/CONTRIBUTING.md#adding-a-license

> Identification of 3 notable projects using the license with straightforward LICENSE files which serve as examples newcomers can follow and that could be detected by licensee if it knew about the license

Are there more straightforward examples with single LICENSE files that new adopters could copy and have their license detected?

mlinksva:
@ljoubert thanks for this pull request.

I'm only getting 78 results for https://github.com/search?utf8=%E2%9C%93&q=filename%3ALICENSE+%22Version+2.1+du+2013-06-21%22&type= (1000 is the minimum criteria for adding a license here). I realize this is probably complicated by some projects using an English translation, and perhaps other things I'm missing. Feel free to point those things out. :smile:

ljoubert:
Hello,

License in French is written licence and sometimes the license is under COPYING. Also to match the 1000 threshold, can it be on the various versions of the CeCILL license (there is explicit forward compatibility since version 1: http://www.cecill.info/licences/Licence_CeCILL_V1-fr.txt

We would match:
- 789 results: https://github.com/search?utf8=%E2%9C%93&q=filename%3ALICENCE+%22CONTRAT+DE+LICENCE+DE+LOGICIEL+LIBRE+CeCILL%22&type=Code
- 242 https://github.com/search?utf8=%E2%9C%93&q=filename%3ALICENSE+%22CONTRAT+DE+LICENCE+DE+LOGICIEL+LIBRE+CeCILL%22&type=Code
 - 75  results : https://github.com/search?utf8=%E2%9C%93&q=filename%3ACOPYING+%22CONTRAT+DE+LICENCE+DE+LOGICIEL+LIBRE+CeCILL%22&type=Code
 
with an incredible total of 1106 in total on github only.

Thank you very much

_licenses/cecill-2.1.txt
ljoubert: Hello, thank you for the explanation, I updated the file accordingly.

_licenses/cecill-2.1.txt
```diff
@@ -0,0 +1,580 @@
+---
+title: CeCILL Free Software License Agreement v2.1
+spdx-id: CECILL-2.1
+source: http://www.cecill.info/licences/Licence_CeCILL_V2.1-fr.txt
+
+description: French open-source license made by the CEA, INRIA and CNRS, compatible with the GNU GPL v2.0 and v3.0
+
+how: Create a text file (typically named LICENSE or LICENCE) in the root of your source code and copy the text of the license into the file. 
+
+using:
```
mlinksva, line 9: BMC-Tools is the only one of these that would be recognized by licensee -- single license file in the repository root with the canonical license text only. Are there other examples like that?

mlinksva:
@ljoubert OK, that's adequate since the count is fuzzy. I would like the examples to be ones that would be detected, see my review comment.

_licenses/cecill-2.1.txt
```diff
@@ -0,0 +1,580 @@
...
+
+  * d'une part, le respect des principes de diffusion des logiciels
+    libres: accès au code source, droits étendus conférés aux utilisateurs,
+  * d'autre part, la désignation d'un droit applicable, le droit
+    français, auquel elle est conforme, tant au regard du droit de la
+    responsabilité civile que du droit de la propriété intellectuelle et
+    de la protection qu'il offre aux auteurs et titulaires des droits
+    patrimoniaux sur un logiciel.
+
+Les auteurs de la licence CeCILL (Ce[a] C[nrs] I[nria] L[ogiciel] L[ibre])
```
mlinksva, line 52: This obtains the following failure:

```
 1) license fillable fields The CeCILL Free Software License Agreement v2.1 license should only contain supported fillable fields
     Failure/Error: expect(extra_fields).to be_empty
       expected `["a", "nrs", "nria", "ogiciel", "ibre"].empty?` to return true, got false
     # ./spec/license_fields_spec.rb:11:in `block (4 levels) in <top (required)>'
```

I'm not sure what to do with this:

- change the license text, perhaps replace square brackets with angle brackets? a bit ugly
- remove test: does it really matter that there are things that look like fields with keys we don't know about? nothing will be done with them, right?

💭 @benbalter ?

ljoubert: @benbalter can I escape the brackets? At the same time I don't want to change the license text or make sure it will be recognized. Thank you very much for your help.

benbalter: Ooph. What if we change the regex in 

https://github.com/github/choosealicense.com/blob/c229dccb6d356729eecfe4ac20cdf26906c1e29e/spec/license_fields_spec.rb#L9

to 

```ruby
/\s+\[([a-z]+)\]\s+/
```

to require the field be surrounded by whitespace?

The reason for the test was because some licenses had `name` and others had `author`, meaning in GitHub.com the inconsistency would cause the fields not to be interpolated.

ljoubert: Hello, are you sure that for the others licenses' text, the fields would always be surrounded by space? For the CeCILL, that would do the trick, since the description has always a letter before.

mlinksva: @benbalter I made your suggested change at d01065b09a3bbf9918f0b46f38e38a3d4aa7eb85 with you as the author. Please scream if you hate that.

@ljoubert I was, but OFL-1.1 has `([email])`. I'm not sure that _should_ be included -- it isn't in https://spdx.org/licenses/OFL-1.1.html while http://scripts.sil.org/cms/scripts/page.php?site_id=nrsi&id=OFL and https://opensource.org/licenses/OFL-1.1 include more headers/fields. Anyway that's a separate issue.

Also, various fields in various licenses are at the beginning or end of a line.

So I suppose the regex should be loosened a bit, and there should be a test to make sure there are no expected fields being excluded.

mlinksva: Attempt at 286357a54c87b3adaf7ae5c5314275626f31b977

amotl:
I would like to see the english version of the CeCILL-2.1 license included. Are there any possibilities to achieve this?

mlinksva:
@amotl we don't have a mechanism here to handle licenses in multiple languages. I'd like to follow SPDX practice as close as possible ... and it seems they last month switched from only having the French versions to only the English ones at https://github.com/spdx/license-list-XML/pull/437

I guess they'll have to figure out how to handle both eventually as I imagine people will want to match both texts; I see there's a placeholder at https://github.com/spdx/license-list-XML/issues/438

amotl:
This sounds promising, thanks for investigating!

bzg:
Hi all, I'm following up @ljoubert suggestion. Is there any news on how SPDX could expose multiple translations?

mlinksva:
@bzg not that I'm aware of. Try asking on https://github.com/spdx/license-list-XML/issues/438 or one of their mailing lists, more likely to be noticed by someone following SPDX work closely.

bzg:
Thanks @mlinksva, I'll do this and report here any progress.

_licenses/cecill-2.1.txt
mlinksva: This (3 notable projects which could be detected by licensee) is the sole remaining thing needed to merge.

bzg: Here are three projects from the french administration using one of the CeCILL licenses:

- https://github.com/etalab/taxe-habitation
- https://github.com/etalab/calculette-impots-m-source-code
- https://github.com/InseeFr/Mesange

These projects are published by the french ministry of economy and finance - they may not not be "notable" in terms of Github fame, but they are important in terms of opengov/transparency.

I let @ljoubert update his PR, I hope this will be enough for it to be merged.

bzg: Sorry, replying to myself: the examples already provided by @ljoubert are throwing errors in the CI.

@mlinksva can you guide us in what could be wrong? Shall we modify the URLs or the text of the licenses to pass the tests in the CI?

mlinksva: @bzg those examples look reasonable to me. The first one is even fairly popular it seems. Of the existing examples I think only the BMC-tools one is workable. It was passing. It looks like the reason it might not be now is encoding, but I'm not sure what changed.

bzg: Okay, thanks again. I'll discuss this with @ljoubert and follow up later on.

bzg: I've spent some time digging into this issue again.

My understanding is that the CI tests fail because the [Licensee gem](https://github.com/licensee/licensee) does not detect the CeCILL v2.1 - which seems obvious because Licensee relies on... choosealicense.com licenses.

In general, I think CeCILL licenses may be hard to detect, so I'd suggest to add an exception in [https://github.com/github/choosealicense.com/blob/gh-pages/spec/license_meta_spec.rb](license_meta_spec.rb), along with the ncsa and postgresql exceptions.

WDYT?

mlinksva: The tests here [monkey patch licensee](https://github.com/github/choosealicense.com/blob/791e3fe05accab0e941a9086c8086c7f42152cc7/spec/spec_helper.rb#L135-L144) to use the license texts available in the branch being tested rather than the ones that licensee vendors. 

I'm not sure why CeCILL licenses should be hard to detect; they aren't templated. Is there something about the text that I'm missing?

bzg: I ruled out the encoding issue, I get the same [CI test failure](https://travis-ci.org/github/choosealicense.com/builds/520862523?utm_source=github_status&utm_medium=notification) with a correct encoding.  I'm out of ideas on how to interpret the failure right now.

mlinksva: Thanks for researching @bzg. I'll try to dig in more when I have some free time.

bzg: Hi @mlinksva, did you have time to find out what causes the CI test failure?

mlinksva: I didn't @bzg thanks for the ping. The bmc-tools and Test-Compta-Demat examples are https://en.wikipedia.org/wiki/ISO/IEC_8859 encoded. I suspect that may be a problem for licensee. But I'm not certain as the vitam example is UTF-8 and an exact match for the text in this PR, but still failing in the test. Which, re-reading your comments above, is exactly the conclusion you came to. 😆In other words, still some debugging to do in the guts of licensee, apologies for this non-update. I'll try to dig some more tonight, feel free to bug me again if I don't say more soon.

mlinksva: Alright, I confirmed that ISO-8859 encoding does cause problems, so bcm-tools and Test-Compta-Demat examples need to change (different repos, or change encoding of those files).

However, the file in vitam does get detected if I add CECILL and (and it as a test fixture) to licensee locally. So there's still something else going on in this repository's tests.

mlinksva: Alright again, the test here wasn't retrieving the vitam license file correctly since the example is on a branch with numbers in its name. Fixed the test in 4b67edf. Now the vitam example will pass and there will still be 2 test failures for the other 2 examples.

bzg: Thanks for the careful and useful tests! I've opened two PRs, fixing encoding for the license files for [Test-Compta-Demat](https://github.com/DGFiP/Test-Compta-Demat/pull/22) and [bmc-tools](https://github.com/ANSSI-FR/bmc-tools/pull/6). If this does not get merge soon enough, we can update the examples URLs and use the github.com/DISIC/ repositories. Thanks again!

bzg: I've waited a bit but I don't think it is reasonable to wait for a merge of [this PR](https://github.com/DGFiP/Test-Compta-Demat/pull/22). I suggest that @ljoubert updates his PR using https://github.com/DISIC/Test-Compta-Demat instead, which does not have the encoding problem.

mlinksva: https://github.com/DISIC/Test-Compta-Demat is just a fork of https://github.com/DGFiP/Test-Compta-Demat rather than another bona fide example. There's got to be another use on github.com or elsewhere publicly accessible.

bzg:
There used to be a problem with the non-UTF-8 encoding of the license file for one of the three examples for the CeCILL 2.1 (that of https://github.com/DGFiP/Test-Compta-Demat).

I ruled this exception out by using another example (https://github.com/etalab/taxe-fonciere).

Now there is [this](https://travis-ci.org/github/choosealicense.com/builds/573261110?utm_source=github_status&utm_medium=notification) failure: 

> 1) license meta The CeCILL Free Software License Agreement v2.1 license licensee detects using examples the https://github.com/DGFiP/Test-Compta-Demat/blob/master/src/testeur/Licence_CeCILL_V2.1-fr.txt URL is a cecill-2.1 license

I'm short of ideas here. Can anyone see if the regexp can be modified in a way that recognize the three examples as valid ones?

Thanks!

mlinksva:
@bzg I'm not sure how you're testing but https://github.com/etalab/taxe-fonciere works for me locally, we'll see what CI says.

mlinksva:
@bzg I think this is finally ready to merge! Let me know if you agree.

ljoubert:
Looks good to me. Thank you very much for your help to both of you !

bzg:
Looks good to me too! Thank you very much for your patience.

PR #530, opened 2017-07-12
Add .htaccess

soccerJoshNumberNine:
The .htaccess file adds a rewrite script to html files that removes the .html from the url seen on the bookmarks bar. A fairly simple change, **no changes occurred except for the addition of this file.** No page content was changed, added, or removed.


soccerJoshNumberNine:
My method was derived from [this page](https://alexcican.com/post/how-to-remove-php-html-htm-extensions-with-htaccess/).

mlinksva:
> the .html from the url seen on the bookmarks bar

Hi, @soccerJoshNumberNine when/where are you seeing this?

All of the links are configured from https://github.com/github/choosealicense.com/blob/04f260afb29790c08e8403b8ad004267a8fa89e6/_config.yml#L11 and in the source for each page to be rendered in a directory with a `index.html` file, accessed by URLs with trailing slashes, so `.html` should never appear.

The project is mainly designed to be deployed on GitHub Pages which doesn't use `.htaccess` or similar configuration files. So closing this pull request, but happy to get fixes for anything on the site not following pattern in previous paragraph that has been missed.

soccerJoshNumberNine:
Ok! I noticed that, and I just wanted to submit this pull request to suggest the use of a rewrite engine, but I did know that it was possible my request would be considered obsolete

PR #535, opened 2017-08-21
Removal of the ambiguous "and/or" in ISC

j8r:
More informations at: https://www.gnu.org/licenses/license-list.en.html#ISC
Original license source: https://cvsweb.openbsd.org/cgi-bin/cvsweb/src/share/misc/license.template?rev=HEAD

### DIFF
--- a/_licenses/isc.txt
+++ b/_licenses/isc.txt
@@ -31,7 +31,7 @@ ISC License
 
 Copyright (c) [year], [fullname]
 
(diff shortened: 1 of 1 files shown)


mlinksva:
'and/or' is in the text collected at https://opensource.org/licenses/ISC https://www.isc.org/downloads/software-support-policy/isc-license/ and https://spdx.org/licenses/ISC.html

We're not in the business of editing existing licenses here, so closing.

PR #539, opened 2017-10-01
Use HTTPS for licences if available

TuxCoding:
I checked if the links in the licenses are reachable with HTTPS (including if they have valid certificate). If they are I changed the url to use HTTPS, but kept the other ones to plain HTTP like: `http://www.wtfpl.net/`

### DIFF
--- a/_licenses/afl-3.0.txt
+++ b/_licenses/afl-3.0.txt
@@ -1,7 +1,7 @@
 ---
 title: Academic Free License v3.0
 spdx-id: AFL-3.0
-source: http://opensource.org/licenses/afl-3.0
+source: https://opensource.org/licenses/afl-3.0
 
 description: The Academic Free License is a variant of the Open Software License that does not require that the source code of derivative works be disclosed. It contains explicit copyright and patent grants and reserves trademark rights in the author.
 
--- a/_licenses/agpl-3.0.txt
+++ b/_licenses/agpl-3.0.txt
@@ -3,7 +3,7 @@ title: GNU Affero General Public License v3.0
 spdx-id: AGPL-3.0
 nickname: GNU AGPLv3
 redirect_from: /licenses/agpl/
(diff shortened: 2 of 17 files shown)


mlinksva:
@games647 good idea for the links in frontmatter, thanks. But I'd rather not change license texts at all. Could you update this to only change frontmatter links?

TuxCoding:
>But I'd rather not change license texts at all. 

Should I then change the text of the unlicense too, because the original version contains only a HTTP link?

https://github.com/github/choosealicense.com/blob/c229dccb6d356729eecfe4ac20cdf26906c1e29e/_licenses/unlicense.txt#L53

vs

https://unlicense.org/UNLICENSE (without HTTPS)

mlinksva:
@games647 yes and thank you!

I shouldn't have merged https://github.com/github/choosealicense.com/pull/520 -- I remember thinking the change was to the frontmatter, but obviously was only seeing what I thought I should see. 😳

TuxCoding:
Done

mlinksva:
Thanks again. Annoying that [this](https://travis-ci.org/github/choosealicense.com/builds/281991806#L899-L903) is consistently happening in CI-run tests:

> - ./_site/existing/index.html
>  *  External link https://www.openbsd.org/policy.html failed: response code 0 means something's wrong.
>             It's possible libcurl couldn't connect to the server or perhaps the request timed out.
>            Sometimes, making too many requests at once also breaks things.
>             Either way, the return message (if any) from the server is: SSL connect error

Though https://www.ssllabs.com/ssltest/analyze.html?d=www.openbsd.org obtains an 'A' rating.

I'll go ahead and merge as it seems that if there's a problem, it's with the test environment.

TuxCoding:
It could be because it's a let's encrypt certificate and the certificate storage on travis isn't up to date to support let's encrypt.

Issue #542, opened 2017-10-18
Literal brackets in license text


dolmen:
The text of the Apache 2.0 license on choosealicense.com has three lines different from the text at http://apache.org/licenses/LICENSE-2.0.txt

```diff
--- LICENSE.choosealicense.com	2017-10-19 00:33:45.000000000 +0200
+++ LICENSE.appache.org	2017-10-11 18:28:21.000000000 +0200
@@ -0,0 +1 @@
+
@@ -181 +182 @@
-      boilerplate notice, with the fields enclosed by brackets "{}"
+      boilerplate notice, with the fields enclosed by brackets "[]"
@@ -189 +190 @@
-   Copyright {yyyy} {name of copyright owner}
+   Copyright [yyyy] [name of copyright owner]
```

Note: The same problem appears also when initialising a new repo on GitHub, but I don't know where to report this issue.

mlinksva:
Hi, I'm pretty sure this has been reported before, but I can't find a reference now. The reason is that the fields in the example notice in the trailer of Apache-2.0 (and *GPL) are not filled out in a deployed `LICENSE` file (they're intended to be used elsewhere, eg in source headers), but GitHub (which indirectly sources license texts from this repo) uses `[]` to fill in fields which are supposed to be replaced in a deployed `LICENSE` file (e.g., in MIT and other template licenses).

The difference you note ensures that GitHub won't modify the trailer material when people choose a license on GitHub. I agree it isn't ideal to modify license texts at all, but that's why, or my understanding anyway.

I'll look into whether there's anything that can be done to avoid the modification.

mlinksva:
> (and *GPL)

Hmm, now that I look again, those use angle not square brackets in the original texts, so there's no good reason to substitute curly brackets. I'll make a PR shortly to fix.

The only other license cataloged here with the difference reported here is ECL-2.0, which is an Apache-2.0 derivative.

wking:
> I'm pretty sure this has been reported before, but I can't find a reference now.

You may be thinking of #417 or #494.

benbalter:
I believe https://github.com/github/choosealicense.com/issues/54 and https://github.com/github/choosealicense.com/pull/55 should explain the technical limitations at play (if any) here.

wking:
Can we re-title this issue “Literal brackets in license text“ or some such?  It's not really Apache-specific.

[...]

PR #543, opened 2017-10-19
Bring gpl-2,3.0 and lgpl-2.1 texts closer to source versions

mlinksva:
As promised at https://github.com/github/choosealicense.com/issues/542#issuecomment-337755749

There are still a couple differences:

- https://www.gnu.org/licenses/lgpl-2.1.txt has a parenthetical in square brackets near the top, our version has parens; maybe could be changed later depending on outcome of #542 
- https://www.gnu.org/licenses/gpl-3.0.txt has gnu.org and fsf.org links with https as does https://www.gnu.org/licenses/agpl-3.0.txt ... that seems to have been introduced only last month between http://web.archive.org/web/20170908053542/http://www.gnu.org/licenses/gpl-3.0.txt and  http://web.archive.org/web/20170930072152/http://www.gnu.org/licenses/gpl-3.0.txt ... unfortunate in my view

cc @wking I imagine you might have comments.

### DIFF
--- a/_licenses/gpl-2.0.txt
+++ b/_licenses/gpl-2.0.txt
@@ -326,8 +326,8 @@ to attach them to the start of each source file to most effectively
 convey the exclusion of warranty; and each file should have at least
 the "copyright" line and a pointer to where the full notice is found.
 
-    {description}
-    Copyright (C) {year}  {fullname}
+    <description>
+    Copyright (C) <year>  <fullname>
 
     This program is free software; you can redistribute it and/or modify
     it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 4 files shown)


wking:
On Thu, Oct 19, 2017 at 12:49:30AM +0000, Mike Linksvayer wrote:
> - https://www.gnu.org/licenses/lgpl-2.1.txt has a parenthetical in
>   square brackets near the top, our version has parens; maybe could
>   be changed later depending on outcome of #542

This PR isn't touching that parenthetical, so I don't think difference
needs to block this PR.  We can circle back to it if #542 results in a
way to use literal square brackets.

> - https://www.gnu.org/licenses/gpl-3.0.txt has gnu.org and fsf.org
>   links with https… that seems to have been introduced only last
>   month…

I really wish folks would version-control these things :p.  I've filed
spdx/license-list-XML#450 to track this in SPDX.  And the LGPL-3.0 has
the same change.

I'd expect these changes are small enough that we can bump this repo
to use the newly-canonical https forms, and rely on Sørensen–Dice to
keep licensee matching [1] regardless of whether a target repository
uses the http or https form.  That way new GitHub repositories using
this data will be seeded with the newly-canonical https.  But http
vs. https seems insignificant enough in these cases that I'd be fine
sticking with http here as well.

[1]: https://github.com/benbalter/licensee/tree/v9.2.1#the-solution

Issue #74, opened 2013-08-06
Set Line Width to 78 Characters


derks:
It is pretty common for text editors, and IDE's to default line wrapping to 78 characters.  It can be annoying to have to modify every license because it breaks to 'early', messing with the format.  

It would be appreciated to require that the content of all licenses not exceed 78 characters.  I'll be more than happy to submit a pull request if this is something that will be accepted.

haacked:
Seems fine with me. I'm surprised we're not already at 78 chars. Are all of them wrong? Or just a small set?

derks:
I checked a few... seems like they are off by one or two... probably 80 characters causing the last word of most lines to wrap at 78 chars.

haacked:
Yeah, I always thought 80 characters was most common in terminals and such. How'd you pick 78 and not say, 76 or 79? I'm not trying to be confrontational here mind you. Just curious. I'm used to editors that don't line wrap so I'd love to be educated here. :smile:

I guess 78 would make sense if the 80 char count included CR and LF.

artob:
Back when we formulated the [Unlicense](http://unlicense.org), there was a discussion on whether the hard line wrap should be 78, 76, or 72 characters. We decided to eventually go with the most conservative 72 characters, since that e.g. facilitates copy-and-paste [use in email](http://stackoverflow.com/questions/459281/do-you-use-word-wrapping-within-emails). Just something to consider in case you're going to change them anyhow.

haacked:
So one thing I'd like to do is keep licenses consistent with what has traditionally been on GitHub. Also, when in doubt, I follow WWMD (What would @mojombo do?). https://github.com/mojombo/jekyll/blob/master/LICENSE is set at 78 chars.

Let's go with `78`. :)

rtomayko:
:+1: 78 chars and really appreciate the attention to detail here. Nailing the text formatting on these licenses is one of the main reasons I personally wanted this project to exist in the first place since it's almost always overlooked / not considered important on most license sites. :heart::heart:

haacked:
@derks: would you be interested in submitting a PR for those that need fixing? We recently updated a lot of them using the plain text version supplied to us by the source. Hopefully there aren't too many that need it.

haacked:
Oh wait, this was merged already, wasn't it?

haacked:
![Uploading thumbs-up-basketball.gif . . .]()

Thanks!

PR #551, opened 2017-11-18
isc.txt: wrap text of the license to 80 columns

waldyrious:
As a bonus, the line breaks are closer to the clause boundaries,
which hopefully makes the text slightly easier to read.


mlinksva:
There are no tests to enforce, but the standard [here](https://github.com/github/choosealicense.com/blob/d6c6f9c4af631a30375313249846ef6402426e99/CONTRIBUTING.md#adding-a-license) is 78 see #74 for explanation. Feel free to argue that is the wrong decision.

Does 78 obtain the same result for ISC?

Related, maybe worth additional issues but just noting here for now:

1. It'd be nice to have a test enforcing wrapping at 78, ideally comparison with output of (2), but even a stupid test that no line in license text is >78 characters would be a start.
2. Is it possible to automatically wrap all licenses at 78 and have output make sense? Or is manual intervention required? I suspect a fair amount of special casing would be needed, but haven't tried. In addition to being used for tests, could be part of script to ingest license texts from SPDX in #489.

mlinksva:
(re 2 above, it would be easier to get good output if wrapping as part of structured SPDX XML license to text, maybe in <https://github.com/spdx/tools/blob/master/src/org/spdx/licensexml/LicenseXmlHelper.java>, rather than attempting to wrap plain text.)

waldyrious:
> Feel free to argue that is the wrong decision.

Honestly, the resolution in that thread sounds more as an arbitrary choice than a pondered decision. That said, @bendiken's comment in that thread (and particularly the SO link within it) does provide some good points in favor of a max width of less than 80 chars, so I won't contest the resolution.

> Does 78 obtain the same result for ISC?

No. Several lines in this PR are 79 or 80 characters long.

> It'd be nice to have a test enforcing wrapping at 78

I agree that such tooling would be helpful, but I'm not able to contribute to that.

> Is it possible to automatically wrap all licenses at 78 and have output make sense?

I'm not sure what you mean by "make sense". Do you have any specific example in mind?

As for this PR, feel free to close it in light of the current guideline of 78 chars.

mlinksva:
First draft of test in #552.

By "make sense" I mean lines that shouldn't be wrapped, aren't, and wrapping maintains indentation appropriately. That requires some smarts about when lines that aren't separated by a newline shouldn't be potentially merged, and figuring out what the appropriate indentation is for each line. Licenses with lists or indented items (eg AFL-3.0, Apache-2.0) cause problems for the `fmt` command, so something smarter than that is required.

Closing this, thanks for the idea!

PR #553, opened 2017-11-20
WIP: Vendor license text from SPDX repo

mlinksva:
Take license texts from SPDX repo as HTML, convert to plain text.

Slightly different approach from #489, which takes license templates from the SPDX API, which is the same as taking license templates from the SPDX repo.

Yet another would be to take license texts from SPDX repo as plain text.

Each has its own problems. I need to dig into SPDX template and text generation to sort out, but for now want to keep each approach in own branch.


benbalter:
>  as HTML

Curious if plaintext or template (also in the repo) may be easier.

mlinksva:
Plaintext is easiest to deal with at this point. Closing this and continuing to add commits to #489.

PR #554, opened 2017-11-24
Add EPL-2.0

### LINKED ISSUE #536: Add Eclipse Public License version 2.0 (epl-2.0)
The Eclipse fundation has published a new version of the Eclipse Public License (epl).

Press release:
https://www.eclipse.org/org/press-release/20170829eplv2.php

Text:
https://www.eclipse.org/org/documents/epl-2.0/index.php

The list in choosealicense.com (and on GitHub) should be updated.

mlinksva:
Fixes #536 

Two things needed for this to be merged:

1. 2 more 'using' examples with unmodified license files (that would be detected by licensee, thus are easy to copy examples) -- help finding those wanted, thanks!
2. SPDX to publish an official ID for the license on its website (note the approval test is failing, but that's because the test looks up OSI approval from SPDX records; that will succeed when the SPDX publishes)

### DIFF
--- a/_licenses/epl-2.0.txt
+++ b/_licenses/epl-2.0.txt
@@ -10,7 +10,9 @@ description: This commercially-friendly copyleft license provides the ability to
 how: Create a text file (typically named LICENSE or LICENSE.txt) in the root of your source code and copy the text of the license into the file.
 
 using:
+  - Eclipse Ditto: https://github.com/eclipse/ditto/blob/master/LICENSE
   - Eclipse SmartHome: https://github.com/eclipse/smarthome/blob/master/LICENSE
+  - SUMO: https://github.com/DLR-TS/sumo/blob/master/COPYING
 
 permissions:
   - commercial-use
--- a/_licenses/epl-2.0.txt
+++ b/_licenses/epl-2.0.txt
@@ -1,0 +1,309 @@
+---
+title: Eclipse Public License 2.0
+spdx-id: EPL-2.0
+redirect_from: /licenses/eclipse/
+source: https://www.eclipse.org/org/documents/epl-2.0/EPL-2.0.txt
+hidden: false
+
+description: This commercially-friendly copyleft license provides the ability to commercially license binaries; a modern royalty-free patent license grant; and the ability for linked works to use other licenses, including commercial ones.
+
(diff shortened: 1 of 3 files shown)


lukateras:
Mind that [the original EPL-2.0 text file](https://www.eclipse.org/org/documents/epl-2.0/EPL-2.0.txt) doesn't end with a newline:

```
@@ -306,4 +274,4 @@
   file in a relevant directory) where a recipient would be likely to
   look for such a notice.
 
-  You may add additional accurate notices of copyright ownership.
+  You may add additional accurate notices of copyright ownership.
\ No newline at end of file
```

_licenses/epl-2.0.txt
```diff
@@ -0,0 +1,309 @@
...
+redirect_from: /licenses/eclipse/
+source: https://www.eclipse.org/org/documents/epl-2.0/EPL-2.0.txt
+hidden: false
+
+description: This commercially-friendly copyleft license provides the ability to commercially license binaries; a modern royalty-free patent license grant; and the ability for linked works to use other licenses, including commercial ones.
+
+how: Create a text file (typically named LICENSE or LICENSE.txt) in the root of your source code and copy the text of the license into the file.
+
+using:
+  - Eclipse SmartHome: https://github.com/eclipse/smarthome/blob/master/LICENSE
```
lukateras: This question applies to all license metadata: why not reference the exact revision, i.e. https://github.com/eclipse/smarthome/blob/36c9e2717e4f89cf60a0d885e7ecc6069b338f48/LICENSE? That way, if the project ever gets relicensed (say, updated to EPL-3.0), the link won't go stale.

lukateras:
Also, just in case if it matches the criteria of notability (probably not), here's my JSON encoder/decoder in Erlang licensed under EPL-2.0: https://github.com/yegortimoshenko/json

_licenses/epl-2.0.txt
wking: > … why not reference the exact revision…

+1 to that.  It would be even nicer to reference a tagged release (e.g. once SmartHome cuts 0.9.0).  As consumers are updated to use different licenses, I expect them to be removed from `using`.  But still, where stable links are easy, I think we should be using them instead of floating with `master`.

mlinksva: See #555 for a different idea.

PR #555, opened 2017-11-26
Test examples provided in 'using:' would be detected by licensee

mlinksva:
Followup to #514 "Note absolutely no testing is done to ensure that 3 examples could be detected (it'd be possible to rig up a test using licensee and the licenses cataloged in a branch) or even are URLs."

This adds such testing. It monkeypatches licensee to use the licenses in the choosealicense.com branch being tested, which might be updated from what has been vendored by licensee (not least when a PR is made to add a license to choosealicense.com).

Prompted to to this by comments at https://github.com/github/choosealicense.com/pull/554#discussion_r153034353 suggesting that examples link to a specific commit rather than master. I don't think that solves the problem as completely as actually testing whatever is linked there, and it covers up potentially confusing examples.

These tests point out some examples that wouldn't be detected by the current version of licensee, and ought be replaced.

First though there are a few non-github.com examples, and they probably also need some URL mangling in the test in order to retrieve just the license text in question.

### DIFF
--- a/spec/license_meta_spec.rb
+++ b/spec/license_meta_spec.rb
@@ -72,6 +72,7 @@ describe 'license meta' do
           content = open(example_url).read
           detected = Licensee::ProjectFiles::LicenseFile.new(content, 'LICENSE').license
           it example_url do
+            skip 'NCSA and PostgreSQL licenses hard to detect' if %(ncsa postgresql).include?(license['slug'])
             expect(detected.key).to eq(license['slug'])
           end
         end
--- a/spec/license_meta_spec.rb
+++ b/spec/license_meta_spec.rb
@@ -64,6 +64,8 @@ describe 'license meta' do
           example_url = example.values[0]
           if example_url.index('https://github.com/') == 0
             example_url.gsub!(%r{\Ahttps://github.com/([\w-]+/[\w-]+)/blob/([\w-]+/\S+)\z}, 'https://raw.githubusercontent.com/\1/\2')
+          elsif example_url.index('https://git.savannah.gnu.org/') == 0 || example_url.index('https://git.gnome.org/') == 0
+            example_url.gsub!(%r{/tree/}, '/plain/')
           end
           content = open(example_url).read
           detected = Licensee::ProjectFiles::LicenseFile.new(content, 'LICENSE').license
--- a/spec/license_meta_spec.rb
+++ b/spec/license_meta_spec.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require 'licensee'
 require 'spec_helper'
 
 describe 'license meta' do
@@ -18,6 +19,8 @@ describe 'license meta' do
         missing = required - raw_fields.keys
         expect(missing).to be_empty
       end
+
+      examples = raw_fields['using'] || []
 
       it 'using contains 3 examples' do
         legacy = [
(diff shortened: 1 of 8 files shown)


lukateras:
This is orthogonal to specifying revision in `using:` URLs. Licensee can still ensure that the linked license is indeed detected. Also, mind that if this gets merged, and metadata still points to branches rather than revisions, fixing `using` URLs will become a source of constant work.

wking:
> ... fixing `using` URLs will become a source of constant work.

I agree, but think that's a good thing.  We don't want to cite projects that used to use a license as current examples.  Ideally we'd have a stable link that worked reliably for consumers, and then a floating link with this logic so we know when to remove a project from `using`.  But two links for each project is more work, and may not be worth working up.

spec/license_meta_spec.rb
```diff
@@ -43,9 +45,38 @@
...
         expect(examples.length).to eq(3)
       end
+
+      context 'licensee detects using examples' do
+        module Licensee
+          class License
+            class << self
+              def license_dir
+                dir = ::File.dirname(__FILE__)
+                ::File.expand_path '../_licenses', dir
```
benbalter, line 57: 👍

spec/license_meta_spec.rb
```diff
@@ -43,9 +45,38 @@
...
+              end
+            end
+          end
+        end
+
+        examples.each do |example|
+          example_url = example.values[0]
```
benbalter: What does this line do?

spec/license_meta_spec.rb
```diff
@@ -43,9 +45,38 @@
...
+          if example_url.index('https://github.com/') == 0
```
benbalter: You may also be able to use `include?` or `start_with?` if you think it's easier to read.

spec/license_meta_spec.rb
```diff
@@ -43,9 +45,38 @@
...
+            example_url.gsub!(%r{\Ahttps://github.com/([\w-]+/[\w-]+)/blob/([\w-]+/\S+)\z}, 'https://raw.githubusercontent.com/\1/\2')
+          elsif example_url.index('https://git.savannah.gnu.org/') == 0 || example_url.index('https://git.gnome.org/') == 0
+            example_url.gsub!(%r{/tree/}, '/plain/')
+          elsif example_url.index('https://bitbucket.org/') == 0
+            example_url.gsub!(%r{/src/}, '/raw/')
+          end
+          content = open(example_url).read
+          detected = Licensee::ProjectFiles::LicenseFile.new(content, 'LICENSE').license
+          it example_url do
```
benbalter: I may make this human readable, to make the failure more clear, e.g., `"#{example_url} is a #{license["slug"]} license"`.

spec/license_meta_spec.rb
```diff
@@ -43,9 +45,38 @@
...
+          example_url = example.values[0]
+          if example_url.index('https://github.com/') == 0
+            example_url.gsub!(%r{\Ahttps://github.com/([\w-]+/[\w-]+)/blob/([\w-]+/\S+)\z}, 'https://raw.githubusercontent.com/\1/\2')
+          elsif example_url.index('https://git.savannah.gnu.org/') == 0 || example_url.index('https://git.gnome.org/') == 0
+            example_url.gsub!(%r{/tree/}, '/plain/')
+          elsif example_url.index('https://bitbucket.org/') == 0
+            example_url.gsub!(%r{/src/}, '/raw/')
+          end
+          content = open(example_url).read
+          detected = Licensee::ProjectFiles::LicenseFile.new(content, 'LICENSE').license
```
benbalter: It *may* be slightly easier / more maintainable to wrap the test in a URL-specifc context (e.g. `context "the #{example_url} URL" do`) and move the `example_url`, `content`, and `detected` variables to `let` statements, (e.g., `let(:content) { open(example_url).read }`, but it's mostly stylistic.

benbalter (approved):
Nice!

spec/license_meta_spec.rb
mlinksva: Each example is a dictionary with project name as key, URL as value. This is getting the URL.

PR #563, opened 2017-12-31
*GPL -or-later and -only in sidebar instructions

mlinksva:
SPDX license list 3.0 uses -or-later and -only over + and no modifier

Reflect that update in sidebar instructions for *GPL as seen e.g. at https://choosealicense.com/licenses/agpl-3.0/

### DIFF
--- a/_includes/sidebar.html
+++ b/_includes/sidebar.html
@@ -1,7 +1,7 @@
 <div class="sidebar">
 
   <a href="#" data-clipboard-target="#license-text" data-proofer-ignore="true" class="js-clipboard-button button">Copy license text to clipboard</a>
-    
+
   <div class="how-to-apply">
     <h3>How to apply this license</h3>
     <p>
@@ -14,7 +14,7 @@
     {% endif %}
     {% assign xgpl = false %}
     {% if page.spdx-id contains 'GPL' %}{% assign xgpl = true %}{% endif %}
-    <p class="note"><strong>Optional: </strong> Add <strong><code>{{ page.spdx-id }}</code>{% if xgpl %}+{% endif %}</strong>{% if xgpl %} (or <strong><code>{{ page.spdx-id }}</code></strong> to disallow future versions){% endif %} to your project's package description, if applicable (e.g., <a href="https://docs.npmjs.com/files/package.json#license">Node.js</a>, <a href="http://guides.rubygems.org/specification-reference/#license=">Ruby</a>, and <a href="http://doc.crates.io/manifest.html#package-metadata">Rust</a>). This will ensure the license is displayed in package directories.</p>
+    <p class="note"><strong>Optional: </strong> Add <strong><code>{{ page.spdx-id }}{% if xgpl %}-or-later{% endif %}</code></strong>{% if xgpl %} (or <strong><code>{{ page.spdx-id }}-only</code></strong> to disallow future versions){% endif %} to your project's package description, if applicable (e.g., <a href="https://docs.npmjs.com/files/package.json#license">Node.js</a>, <a href="http://guides.rubygems.org/specification-reference/#license=">Ruby</a>, and <a href="http://doc.crates.io/manifest.html#package-metadata">Rust</a>). This will ensure the license is displayed in package directories.</p>
   </div>
 
   {% if page.source %}
--- a/_licenses/apache-2.0.txt
+++ b/_licenses/apache-2.0.txt
@@ -13,8 +13,8 @@ how: Create a text file (typically named LICENSE or LICENSE.txt) in the root of 
 note: The Apache Foundation recommends taking the additional step of adding a boilerplate notice to the header of each source file. You can find the notice at the very end of the license in the appendix.
 
 using:
-  - Elasticsearch: https://github.com/elastic/elasticsearch/blob/master/LICENSE.txt
   - Kubernetes: https://github.com/kubernetes/kubernetes/blob/master/LICENSE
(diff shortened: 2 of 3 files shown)


_includes/sidebar.html
```diff
@@ -14,7 +14,7 @@ <h3>How to apply this license</h3>
...
+    <p class="note"><strong>Optional: </strong> Add <strong><code>{{ page.spdx-id }}{% if xgpl %}-or-later{% endif %}</code></strong>{% if xgpl %} (or <strong><code>{{ page.spdx-id }}-only</code></strong> to disallow future versions){% endif %} to your project's package description, if applicable (e.g., <a href="https://docs.npmjs.com/files/package.json#license">Node.js</a>, <a href="http://guides.rubygems.org/specification-reference/#license=">Ruby</a>, and <a href="http://doc.crates.io/manifest.html#package-metadata">Rust</a>). This will ensure the license is displayed in package directories.</p>
```
wking: I haven't checked all of these locations, but the [`package.json` docs][1] explicitly call for SPDX 2.0:

>  If your package is licensed under multiple common licenses, use an SPDX license expression syntax version 2.0 string, …

although it's not clear to me how strongly they mean that.  In any case, I think it's probably best to continue to recommend SPDX license expressions use the old identifiers, at least until consumering tools have had some time to update.  It's also not entirely clear to me how the license list is supposed to interact with the spec.  Does [the 2.1 spec's inclusion of the 2.5 license list][2] limit SPDX 2.1 license expressions to those IDs?  In spdx/spdx-spec#37, I'm suggesting [explicitly pinning the license-list version][3].  If that sounds like a possible interpretation of the current spec, you may want to wait at least until the next SPDX *spec* release before recommending the 3.0 license list identifiers.

[1]: https://docs.npmjs.com/files/package.json#license
[2]: https://spdx.org/spdx-specification-21-web-version#h.1jlao46
[3]: https://github.com/spdx/spdx-spec/pull/37/files#diff-f006efd0d473205cd5b1f2fee088e3f1R7

mlinksva: I assumed the there is no intent to have the spec version imply a license list version. Surely that intent would be clearly expressed places like:

> The following table contains the full names and short identifiers for the SPDX License List, v2.5 which was released July 2016. For the full and most up-to-date version of the SPDX License List as well as other related information, please see http://spdx.org/licenses/

Not to mention in the definitions of various license declared/concluded/data license fields.

I'll leave this open pending more information, and subscribe to the issue you point out. Thanks!

mlinksva: Meant to mention: there's an optional `LicenseListVersion` field in the SPDX spec, which is a pretty strong implication that a spec version isn't tied to a license list version, though a producer can explicitly specify one.

wking: > …there's an optional `LicenseListVersion` field in the SPDX spec…

That sounds useful when you have an SPDX file, although [the spec][1] doesn't seem to discuss the default value.  The [`package.json` docs][2] don't either, and there's no way to specify a license-list version there.  License-expression consumers would seem to need both an SPDX-spec version for the license-expression syntax and (if it is intended to float independently, as `LicenseListVersion` implies) a license list version.

[1]: https://spdx.org/spdx-specification-21-web-version#h.1y810tw
[2]: https://docs.npmjs.com/files/package.json#license

mlinksva: > until consumering tools have had some time to update

ex: https://github.com/rust-lang/cargo/issues/4888

mlinksva: @wking you've probably seen elsewhere, but I believe https://github.com/shinnn/spdx-license-ids/pull/8 is where what eventually is used by the NPM registry and tools will pick up support for SPDX license list verion 3.0.

Next time I'm procrastinating I'll track down the equivalent for Rubygems.

After all 3 are merged I'll merge this PR. Yes there are other package managers, they'll come along eventually.

wking: The Rust docs seem to have moved to https://doc.rust-lang.org/cargo/reference/manifest.html#package-metadata:

```console
$ curl -s http://doc.crates.io/manifest.html | grep moved
This page has moved. Click <a href="https://doc.rust-lang.org/cargo/reference/manifest.html">here</a> to go to the new page.
```

And those docs do not yet contain rust-lang/cargo#4898.  Perhaps they only update them after Cargo releases?

And the Ruby docs seem to have a similar issue to the old Rust/Cargo `/`, with [`licenses`][1] that does not distinguish between `AND` and `OR` and a lack of clarity about the SPDX (list) versions and whether [`license`][2] takes license expressions or just short identifiers.  I think pulling the Rust declarations forward is going to absorb my available free time, so I'll leave pulling Ruby forward to someone else ;).

[1]: http://guides.rubygems.org/specification-reference/#licenses=
[2]: http://guides.rubygems.org/specification-reference/#license=

mlinksva: @wking I think I found where for Ruby: https://github.com/rubygems/rubygems/pull/2152

mlinksva: (I mean where rubygems gets its list of licenses; lack of support for license expressions and/or use instead of ambiguous `licenses` I haven't looked into, and doesn't impact this PR.)

mlinksva: For my own memory, https://github.com/npm/npm/issues/19558 is probably what will be closed when npm uses the updated license list.

mlinksva: npm 6, released today, includes updated license list support.

7koma95 (approved):

PR #565, opened 2018-01-03
project_url -> projecturl

mlinksva:
Followup to #564 which documented `project_url`. This removes the underscore from the field name.

Not having to deal with underscores in field names makes things a bit simpler for other tooling that may or may not preprocess license texts -- I'm thinking particularly of https://github.com/benbalter/licensee/blob/4e20608c7252c93fde3972019465ab705762cf4a/lib/licensee/content_helper.rb#L14 which treats underscore as markdown markup to be stripped.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ The licenses on choosealicense.com are regularly imported to GitHub.com to be us
 * `project` - The repository name
 * `description` - The description of the repository
 * `year` - The current year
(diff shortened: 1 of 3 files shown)


BLUEOO73 (approved):

sozoops:
> # - [ ] - []()@github-cla ~~````

````~~****bjmss

PR #424, opened 2016-05-31
make tests not overwrite non-mutable document 'id'

mlinksva:
WIth Jekyll 3.1.6 CI is now failing. Skimming sources I'm not sure when introduced, but now the class that represents a document has a 'id' method and is considered immutable, so our tests which add an 'id' property to each license to set up the tests now fail when adding that property (see stracktrace below). So I changed the property added and used from 'id' to 'spdx-lcase' which is descriptive of the property value. Now tests/CI will pass.

cc @parkr as likely to be able to tell me if I'm missing something obvious or otherwise doing it wrong.

```
/home/travis/build/github/choosealicense.com/vendor/bundle/ruby/2.0.0/gems/jekyll-3.1.6/lib/jekyll/drops/drop.rb:78:in `[]=': Key id cannot be set in the drop. (Jekyll::Errors::DropMutationException)
    from /home/travis/build/github/choosealicense.com/vendor/bundle/ruby/2.0.0/gems/jekyll-3.1.6/lib/jekyll/drops/drop.rb:194:in `block in merge!'
    from /home/travis/build/github/choosealicense.com/vendor/bundle/ruby/2.0.0/gems/jekyll-3.1.6/lib/jekyll/drops/drop.rb:185:in `each_key'
    from /home/travis/build/github/choosealicense.com/vendor/bundle/ruby/2.0.0/gems/jekyll-3.1.6/lib/jekyll/drops/drop.rb:185:in `merge!'
    from /home/travis/build/github/choosealicense.com/vendor/bundle/ruby/2.0.0/gems/jekyll-3.1.6/lib/jekyll/drops/drop.rb:177:in `block in merge'
    from /home/travis/build/github/choosealicense.com/vendor/bundle/ruby/2.0.0/gems/jekyll-3.1.6/lib/jekyll/drops/drop.rb:175:in `tap'
    from /home/travis/build/github/choosealicense.com/vendor/bundle/ruby/2.0.0/gems/jekyll-3.1.6/lib/jekyll/drops/drop.rb:175:in `merge'
    from /home/travis/build/github/choosealicense.com/spec/spec_helper.rb:38:in `block in licenses'
    from /home/travis/build/github/choosealicense.com/spec/spec_helper.rb:36:in `map'
    from /home/travis/build/github/choosealicense.com/spec/spec_helper.rb:36:in `licenses'
    from /home/travis/build/github/choosealicense.com/spec/license_fields_spec.rb:4:in `block in <top (required)>'
[...]

PR #318, opened 2016-01-18
Add script to check approval of a license.

benbalter:
A follow up to https://github.com/github/choosealicense.com/pull/316, this adds a quick script to check the approval of a given license to help with responding to issues proposing new licenses be added.

Example output:

```
➜  choosealicense.com git:(approval-script) script/check-approval mit
+--------------+-------------+
|        License: mit        |
+--------------+-------------+
| SPDX ID      | MIT         |
| SPDX Name    | MIT License |
| OSI approved | Yes         |
| FSF approved | No          |
| OD approved  | No          |
+--------------+-------------+
| Eligible     | Yes         |
+--------------+-------------+

Code search: https://github.com/search?q=mit+filename%3ALICENSE&type=Code
```

```
➜  choosealicense.com git:(approval-script) script/check-approval foobar
+--------------+---------+
|    License: foobar     |
+--------------+---------+
| SPDX ID      | Invalid |
| SPDX Name    | None    |
| OSI approved | No      |
| FSF approved | No      |
| OD approved  | No      |
+--------------+---------+
| Eligible     | No      |
+--------------+---------+

Code search: https://github.com/search?q=foobar+filename%3ALICENSE&type=Code
```

PR #567, opened 2018-01-03
check-approval: Fix missing license_ids variable

wking:
Avoid:

```console
$ ./script/check-approval MIT
./script/check-approval:55:in `<main>': undefined local variable or method `license_ids' for main:Object (NameError)
```

The `license_ids` line was added in e505eb8f (#318), but `license_ids` was removed from the helper in b99e7ab0 (#424).  This commit restores the old code locally, since this script is the only consumer.

### DIFF
--- a/script/check-approval
+++ b/script/check-approval
@@ -52,6 +52,7 @@ approvals.each do |approver, licenses|
   rows << ["#{approver} approved", licenses.include?(license)]
 end
 
+license_ids = licenses.map { |l| l['id'] }
 current = license_ids.include?(license)
 rows << ['Current license', current]
 


wking:
I've pushed f845adf → d2ed77d to pick up #568.

wking:
And I've pushed d2ed77d → 45369ac to pick up #569.

mlinksva:
I personally don't think it's worth the effort and potential disruption of anyone else working on branch to force push when picking up updates. But it seems to be part of your workflow, so go ahead. Thanks for the contributions! Here and to the various license meta projects you are so steadfastly improving. 👍

wking:
> I personally don't think it's worth the effort and potential disruption of anyone else working on branch to force push when picking up updates.

Ah, sorry.  I thought it would help folks in the future by avoiding backwards merges, when we could rebase to accomplish the same thing.  I'll let the backwards merges go in the future ;).

PR #488, opened 2017-03-04
Use CC0-1.0 license text from source

mlinksva:
The existing license text was introduced in https://github.com/github/choosealicense.com/pull/126 presumably by copying and pasting from https://creativecommons.org/publicdomain/zero/1.0/legalcode though I'm not sure how to explain the link at the end. Probably should be updated with https://creativecommons.org/publicdomain/zero/1.0/legalcode.txt or https://spdx.org/licenses/CC0-1.0.html ... latter marks disclaimer as optional and doesn't have first line "Creative Commons Legal Code". This commit uses former.

Making due to https://github.com/benbalter/licensee/issues/172 but holding off on merge pending more discussion there.

Fixes #518

PR #574, opened 2018-03-06
Use https for CC0 link

Rotzbua:
Use https for CC0 link


mlinksva:
I'd rather not introduce another variation on the license text. That link will go away when #488 switches to using the text provided by the license publisher. Thanks!

PR #580, opened 2018-04-10
Adding the Universal Permissive License 1.0.

Djelibeybi:
Oracle would like to propose the addition of the Universal Permissive License (UPL) 1.0.

Signed-off-by: Avi Miller <avi.miller@oracle.com>

### DIFF
--- a/_licenses/upl-1.0.txt
+++ b/_licenses/upl-1.0.txt
@@ -1,0 +1,65 @@
+---
+title: The Universal Permissive License (UPL), Version 1.0
+spdx-id: UPL-1.0
+source: https://oss.oracle.com/licenses/upl/
+featured: true
+hidden: false
+
+description: A permissive, OSI and FSF approved, GPL compatible license, expressly allowing attribution with just a copyright notice and a short form link rather than the full text of the license.  Includes an express grant of patent rights.  Licensed works and modifications may be distributed under different terms and without source code, and the patent grant may also optionally be expanded to larger works to permit use as a contributor license agreement.
+
+how: Insert the license or a link to it along with a copyright notice into your source file(s), and/or create a text file (typically named LICENSE or LICENSE.txt) in the root of your source code and copy the text of the license and your copyright notice into the file.
+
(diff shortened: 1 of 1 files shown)


Djelibeybi:
I've fixed the Travis CI failures that I can, but I'm not sure how to fix the rest of them. Some guidance here would be appreciated.

mlinksva:
@Djelibeybi thanks for adding this. The main thing needing fixing is the examples. None of the 3 have LICENSE files in their roots containing only UPL-1.0, so we can't automatically detect that the examples are UPL-1.0 licensed. We're looking for examples that are trivial to follow for new projects wishing to use the license.

Djelibeybi:
@mlinksva thanks for that feedback. I'll find some projects that are only licensed under the UPL and update accordingly.

Djelibeybi:
@mlinksva I've updated the using section, but it's still failing. Is this because the word wrap is different on those repos or something else? I can update the LICENSE files if necessary, I just want to make sure I make the right changes.

mlinksva:
@Djelibeybi I don't see an obvious reason those new examples you've added aren't being detected. I'll investigate, but probably not until tomorrow.

Djelibeybi:
@mlinksva all good. :)

mlinksva:
The problem was that examples should link to `LICENSE` files. Tests pass now.

For anyone following along https://github.com/search?utf8=%E2%9C%93&q=%22Universal+Permissive+License%22+filename%3ALICENSE&ref=simplesearch obtains 1280 results.

Ideally all of the examples wouldn't be under the same org (oracle) but that's not a requirement.

I'll leave this open before merging a couple days to give others a chance to review the license metadata and description -- looks fine to me.

Djelibeybi:
Thanks for fixing that up. :) FYI, all of the non-Oracle examples I could find are dual-licensed under the UPL and Apache 2.0 licenses, which breaks the CI testing.

mlinksva:
Hmm, I'd think that using the UPL would decrease the instinct to dual license as Apache-2.0, where that instinct exists. I guess as people get used to UPL, maybe. Tiny increment on that, merging this PR. :smile:

PR #581, opened 2018-04-10
Updated GPL licenses from official site

ale5000-git:
Ref: https://github.com/benbalter/licensee/pull/291

### DIFF
--- a/_licenses/agpl-3.0.txt
+++ b/_licenses/agpl-3.0.txt
@@ -37,7 +37,7 @@ limitations:
                     GNU AFFERO GENERAL PUBLIC LICENSE
                        Version 3, 19 November 2007
 
- Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
  Everyone is permitted to copy and distribute verbatim copies
  of this license document, but changing it is not allowed.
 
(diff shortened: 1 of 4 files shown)


wking:
There's some brief previous discussion of this change in #543.  SPDX has explicitly allowed both forms in the GPL-3 family since spdx/license-list-XML#450, and is considering blanket support for HTTP(S) (spdx/license-list-XML#633).

ale5000-git:
@wking
It would be nice to support both but also it would be nice to display the updated one [here](https://choosealicense.com/licenses/gpl-3.0/) and in the other places.

mlinksva:
Not planning on changing line wrapping now, but https update in in #588

ale5000-git:
@mlinksva: Did you notice that `www.gnu.org/philosophy/why-not-lgpl.html` is changed to `www.gnu.org/licenses/why-not-lgpl.html` inside gpl-3?

What about this?

mlinksva:
@ale5000-git I didn't notice that, thanks for pointing it out again. Looks like that change was made at the same time as http->https, looking at the wayback machine links I posted at https://github.com/github/choosealicense.com/pull/543#issue-147444584

Perhaps nobody other than you has noticed until now as I see SPDX doesn't account for matching both paths yet. I think that's the first place to address it.

ale5000-git:
Thanks.

ale5000-git:
@mlinksva

Hi,
the above-mentioned link isn't updated here: https://choosealicense.com/licenses/gpl-3.0/

mlinksva:
@ale5000-git this repo doesn't directly consume material from spdx/license-list-XML, though I hope to [eventually](https://github.com/github/choosealicense.com/pull/489). Getting the change reflected there is a way of getting additional review before we make changes here.

If you want to submit a new PR that only changes `philosophy` to `licenses` inside the `why-not-lgpl` URL here I'll accept it now.

ale5000-git:
Done.

PR #583, opened 2018-04-18
Fix some licenses substitutions

orsinium:
Github use some substitutions in license text:

- `[year]` -- current year.
- `[fullname]` -- name of repository owner.

Apache 2.0 and ECL 2.0 licenses contains this wrong substitutions: `[yyyy] [name of copyright owner]`. Many people forget fix it manually. For example, [yapf](https://github.com/google/yapf/blob/master/LICENSE#L190).

### DIFF
--- a/_licenses/apache-2.0.txt
+++ b/_licenses/apache-2.0.txt
@@ -223,7 +223,7 @@ limitations:
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright [year] [fullname]
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
--- a/_licenses/cc-by-4.0.txt
+++ b/_licenses/cc-by-4.0.txt
@@ -1,5 +1,5 @@
 ---
-title: Creative Commons Attribution 4.0
+title: Creative Commons Attribution 4.0 International
 spdx-id: CC-BY-4.0
 source: https://creativecommons.org/licenses/by/4.0/legalcode.txt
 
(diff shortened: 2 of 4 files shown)


orsinium:
I don't know why Travis CI build failed. All fails in "cc-by-4.0", not in edited files.

orsinium:
I'm actualize CC-BY licenses names. Now Travis CI Build is passed

mlinksva:
The text in the Apache 2.0 appendix is not intended to be substituted in `LICENSE` -- only when deployed in file headers. Related info, see #558 

The CC names were corrected in SPDX license list 3.1 which was recently released, see https://github.com/spdx/license-list-XML/issues/618 ... thanks for updating so that tests pass. I'll cherry-pick your fix in a new PR.

waldyrious:
> The text in the Apache 2.0 appendix is not intended to be substituted in LICENSE -- only when deployed in file headers. Related info, see #558

@mlinksva do you think it would be helpful to include that information in the usage notes / how to apply section for the Apache 2.0 license?

In particular, is it acceptable to omit the appendix from the license file? (That would be the opposite of #558, btw.) I'm asking because PDF.js, [one of the example projects](https://github.com/github/choosealicense.com/blob/6642fd15cb9e32d19785b8276c0136d401c90539/_licenses/apache-2.0.txt#L16) using this license, does so in their [LICENSE file](https://github.com/mozilla/pdf.js/blob/master/LICENSE).

mlinksva:
@waldyrious yes it is acceptable (and licensee accounts for that by stripping text after "end of terms and conditions"; a [test](https://github.com/licensee/licensee/blob/06cfa8ee1046604df0ddc300ffcb9e28a31ab159/spec/fixtures/gpl3-without-instructions/LICENSE)).

I'm ambivalent about adding to the usage notes as I'd rather not cause people to think about modifying the license texts at all, but I could see doing it for Apache 2.0 given the occasional misunderstanding about use of the appendix as a license file. Feel free to make a PR.

waldyrious:
Thanks for the clarification. I'll give it a shot.

PR #598, opened 2018-08-01
Update licenses.html to reflect the change of the home page

NutshellySima:
Because the Apache 2.0 license has been removed from the home page, thus update licenses.html to reflect the change.


NutshellySima:
The travis build failure is caused by Eclipse che switched to EPL2.0; Fixed in another patch (https://github.com/github/choosealicense.com/pull/599).

mlinksva:
Thanks!

PR #601, opened 2018-08-04
rm source metadata, defer to SPDX record

mlinksva:
I think the SPDX page for each license is superior to identifying a single source URL for each license in this project, especially given that the single source URL is sometimes squirrelly or context-free (latter: text files) while the SPDX page contains the source and relevant other links.

What about consumers?

Licensee does expose license metadata, including source: https://github.com/benbalter/licensee/blob/0d87ecc0f59189a670597d001641abf775433362/docs/usage.md#license-ruby-api

It could manufacture source metadata, as [done in this PR](https://github.com/github/choosealicense.com/pull/601/files#diff-88116a8ad5517e8b9bfd0e9911642382R22).

The GitHub licenses API does not expose source metadata https://developer.github.com/v3/licenses/#response-1

### DIFF
--- a/_includes/sidebar.html
+++ b/_includes/sidebar.html
@@ -19,7 +19,7 @@
 
   {% if page.source %}
   <div class="source">
-    <a href="{{ page.source }}">
+    <a href="https://spdx.org/licenses/{{ page.spdx-id }}.html">
       <span class="license-sprite"></span>
       Source
     </a>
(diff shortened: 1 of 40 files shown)


benbalter (approved):

_includes/sidebar.html
```diff
@@ -19,7 +19,7 @@ <h3>How to apply this license</h3>
 
   {% if page.source %}
```
benbalter, line 20: @mlinksva this will now never be true. We'll need to remove this conditional if we want the source link to display.

PR #605, opened 2018-08-19
Change phrasing of no-license text.

wolfgang42:
Someone asked me for advice on picking an open source license, so I pointed them to choosealicense.com. They read all of the options on the page, got down to “You don’t have to” and said “oh good, I won’t bother then” and closed the tab. This rather missed the point (they really did want users to be able to use their program), so here’s a PR that clarifies the wording to make it clearer that there are consequences to this decision.

### DIFF
--- a/index.html
+++ b/index.html
@@ -72,7 +72,7 @@ permalink: /
   <li>
     <h3>I don’t want to choose a license.</h3>
     <p>
-      <a href="no-permission/">You don’t have to</a>.
+      <a href="no-permission/">Here’s what happens if you don’t</a>.
     </p>
   </li>
 </ul>


mlinksva:
Seems like a great idea. Thanks for following up based on experience!

waldyrious:
I agree with the spirit of this change, but isn't the new wording a little too clickbaity? As in "You won't believe what happens next", etc. Maybe we could come up with a better wording. I can't think of something concise OTTOMH, but I was thinking of something along the lines of "You don't have to, but that may stifle collaboration".

mlinksva:
It is in the stereotypical clickbait headline style. Happy to take further edits in a different style though I'd prefer to keep the wordcount very short so as to stay on one line and match the other two things on that line.

PR #610, opened 2018-10-09
feat: suggest the license to a repository from the license page

felubra:
This PR implements #401 by adding a new "Suggest this license" feature in the sidebar of the license page. The user can input a repository URL and, upon pressing enter, being redirected to a page where he can open a PR to add the license to the project. 

This PR is based on the [idea](https://github.com/github/choosealicense.com/issues/401#issuecomment-397908544) and valuable feedback by @mlinksva.

_Edit in 2018-10-19: this explains the behavior after 4f1eab6._


mlinksva:
Thanks @felubra! Glad to see this basically works.

The project selection/autocomplete seems a little finicky to me. I think maybe because it is only searching repo name, not the user/owner. I guess it would be possible to search by name and/or owner depending on what is typed in, and combine results when searching both, but it's probably hard to get exactly right.

Another way might be to not do any realtime lookup, but just take what user enters, do something with it upon hitting enter (initially only supporting full URL even), where something might be check that it is not licensed (glad you thought to filter for that, BTW) before sending on to the github.com license chooser?

felubra:
@mlinksva thank you for the feedback!

> Another way might be to not do any realtime lookup, but just take what user enters, do something with it upon hitting enter (initially only supporting full URL even), where something might be check that it is not licensed (glad you thought to filter for that, BTW) before sending on to the github.com license chooser?

I think this workflow is a better option, I will do something based on this. Thank you!

felubra:
Following up on this, here is my implementation of your second suggestion @mlinksva. Thanks!

assets/js/app.coffee
```diff
@@ -70,6 +71,109 @@ class Choosealicense
...
+        @setStatus "Fetching"        
+        @fetchInfoFromGithubAPI repositoryFullName, (err, repositoryInfo=null) =>
+          if (err)
+            @setStatus "Error", err.message
+            return
+          if repositoryInfo.license # The repository already has a license
+            license = repositoryInfo.license
+            @setStatus "Error", @repositoryLicense repositoryFullName, license
+          else # The repository is unlicensed
+            window.open "https://github.com/#{repositoryFullName}/community/license/new?template=#{@licenseId}"
```
mlinksva: This seems to trigger both Firefox and Chrome popup blockers. Would `location.href` work instead?

mlinksva:
This is great @felubra! If it can avoid a popup (see my review comment) I think it just needs a tiny bit of documentation, which I can suggest.

assets/js/app.coffee
felubra: Good catch!

felubra:
@mlinksva after making the change you requested, I've noticed that the URL to suggest a license to a third-party repo is restricted to logged-in Github users. Any logged-out attempts to access it would result in a 404 error. I've prefixed it with the `https://github.com/login?return_to=`, so the user can log-in before going to the restricted page. Not sure if this is the right way to do this 😊

mlinksva:
Hmm, 404 is definitely not a great experience. I don't think there's any way to detect if a visitor to choosealicense.com is logged into github.com, so guess the solution will have to be more up-front context, warning that one has to be logged into github.com?

Maybe in the case of a valid URL for a non-licensed repo, present a click-through that explains you need to be logged in -- in place of an error if you provide a non-valid URL or a licensed repo?

felubra:
>Hmm, 404 is definitely not a great experience. I don't think there's any way to detect if a visitor to choosealicense.com is logged into github.com, so guess the solution will have to be more up-front context, warning that one has to be logged into github.com?

After 2d3c449 the user will never get a 404 because he will be redirected to the login page of Github, that is, after pressing enter and if the URL typed is of a valid unlicensed repository, the browser will be redirected to `https://github.com/login?return_to=%URL_to_suggest_the_license%`. If the user is logged in, Github will redirected the user transparently to the `%URL_to_suggest_the_license%`; if not, a login form appears, the user has the opportunity to log in and, after doing so, will be redirected to `%URL_to_suggest_the_license%`.

>Maybe in the case of a valid URL for a non-licensed repo, present a click-through that explains you need to be logged in -- in place of an error if you provide a non-valid URL or a licensed repo?

Sorry if I din't understood, but checking for a invalid URL or a already licensed repository does not require a logged in user. This is done by validating the URL against a regex and fetching the public Github API, respectively.

I agree with the need for more explanation. Maybe a paragraph above the input with something like:

> You can suggest this license to a Github repository to make it easier for other people to contribute - this is done by opening a pull request, therefore you need to authenticate with your Github account to do this.

Sounds good?

mlinksva:
@felubra apologies, I failed to notice 2d3c449 or realize that return_to is a solution, I obviously did not read your comment closely, or think. 😳 

Your suggestion for an explanation looks good to me. Thanks!

felubra:
No problem 😊
I've added the explanation. Let me know if you need anything else.
Thanks!

mlinksva:
Great, thanks @felubra! I'm pretty happy with this now. I'm going to leave it open a few days to give others a chance to give feedback, since this is in a way a pretty big feature for choosealicense.com, though it didn't take much code or UI real estate.

kfogel:
Just want to say thanks for taking action on this, @felubra!

felubra:
@kfogel I'm having a great experience and learning a lot about contributing to OSS, thank you for the providing me this opportunity 😊

kfogel:
Well I look forward to working with you more @felubra!  Actually you and @mlinksva did the real work here.  What I did was more "cheerleading" :-).

Issue #441, opened 2016-07-06
Annotating license texts with license rules


mlinksva:
[Comment/question today](https://github.com/github/choosealicense.com/pull/320#issuecomment-230746990) about whether EUPL-1.1 is accurately described reminded me to file this enhancement idea.

For each license, license rules could be annotated with ranges of text in the license pertinent to the rule. Highlighting of ranges could be turned on/off on individual license pages by selecting in the license rules (permissions/conditions/limitations) table. Very crude mockup taking a very simple case (the one condition of MIT).

![mit-highlight-condition](https://cloud.githubusercontent.com/assets/40415/16633043/25735510-437c-11e6-84f8-1e504d48f345.png)

Obviously this is not a big help for MIT, but for longer licenses, it can be tricky to figure out what bits of the license are pertinent for a particular rule, at least if you only want to read once, which is more already than I suspect most people do.

Such annotations _might_ be very helpful to "demystify" licenses, and creating them might be good QA for choosealicense.com license descriptions.

Probably the simplest way to to annotate would be to have one or more full text extracts for each license for each rule, though there are other ways to do it, inline or [out of band](https://github.com/hypothesis/h/wiki/robust-anchors).

Very low priority, may even be a bad idea, just recording it for now.

waldyrious:
This would be great! :+1:

DigiBytes:
Cool

okenwa103:
Okenwa103

PR #614, opened 2018-10-22
[WIP] feat: click on a license rule to highlight his associated text ranges

felubra:
This implements #441:

license-annotation-demo

For this to work, the ranges are defined inside the license file as arrays of `start-end` string positions, inside each rule, that can now be a object too. The purposed new data format does not break the old. **Example**: 
```
permissions:
  - tag: commercial-use
    ranges:
    - 249-393
    - 401-523
  - modifications
  - distribution
  - patent-use
  - private-use
```
In this example, only the `commercial-use` rule is annotated.
@mlinksva could you please review this implementation? This is not complete yet, as I need to deal with overlapping ranges and I think the UI need some adjustments, especially to indicate the presence of the annotated rules - any suggestion is welcome.

If this gets traction, I think we should make some tool, maybe inside the license page itself, to aid the creation of these annotated ranges of text.

### DIFF
--- a/_layouts/license.html
+++ b/_layouts/license.html
@@ -31,10 +31,19 @@
                         {% for rule_obj in rules %}
                           {% assign req = rule_obj.tag %}
                           {% if page[type] contains req %}
-                            <li class="{{ req }}">
+                            <li class="{{type}} {{ req }}">
                               <span class="license-sprite"></span>
                               {{ rule_obj.label }}
                             </li>
+                          {% else %}
+                            {% for rule in page[type] %}
+                              {% if rule.tag == req %}
+                                <li class="{{type}} {{ req }} annotated-rule" data-rule-ranges="{{ rule.ranges | join: "," }}">
+                                  <span class="license-sprite"></span>
+                                  {{ rule_obj.label }}
+                                </li>
+                              {% endif %}
+                            {% endfor %}
                           {% endif %}
                         {% endfor %}
                       </ul>
--- a/assets/css/application.scss
+++ b/assets/css/application.scss
@@ -240,6 +240,10 @@ strong {
   margin-bottom: 5px;
 }
 
+.license-rules li.annotated-rule {
+  cursor: pointer;
+}
+
 .license-rules li:hover {
   color: #444;
 }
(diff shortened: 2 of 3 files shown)


mlinksva:
🆒 @felubra! Couple quick comments:

I'm not thrilled with absolute character ranges. If e.g., a text were re-wrapped, they would have to be redone, right? Happy to be convinced this is still the right approach. When I filed #441 I had in mind reviewing https://web.hypothes.is/blog/fuzzy-anchoring/ though I'm not sure I see the most obvious more-robust-but-still-simple option: record unique strings that mark beginning/end of ranges.

Because it's hard to know until it's done how challenging (in a few dimensions) identifying all of the portions of each license associated with each property will be, until doing it, I think it might be best to do in a parallel collection rather than in `_licenses`. But if we were to try to put all of the annotations in `_licenses`, I'm not sure I follow "The purposed new data format does not break the old" because it looks to me that software expecting the old data format would need to be updated.

felubra:
@mlinksva 

> If e.g., a text were re-wrapped, they would have to be redone, right?

I don't think so, since the legal text is inside a `<pre>` tag. I think the annotations would only need to be redone if the license text changes. 

After listening to you and now facing a bug that I thought I've fixed (a `&gt;` entity being parsed as a `>`, creating a new node and breaking the code), I will abandon this unreliable approach and research more about web annotations - a tricky topic, now I realize, after reading the hypothesis blog post.

> I'm not sure I follow "The purposed new data format does not break the old" because it looks to me that software expecting the old data format would need to be updated

You are right. I wrote that with only the perspective of the website consuming the data from the license file.

> I think it might be best to do in a parallel collection rather than in _licenses

I think one crucial aspect of this problem is to define where this annotations would came from.

I will research more on the topic of web annotations and propose a new PR when I feel confident with a new solution again. Thank you! 😊

mlinksva:
@felubra license texts don't change often, but they do. Not substantively of course, but formatting, http->https, wrapping/indentation due to bringing closer to the SPDX cataloged versions. I really appreciate your prototyping and investigating this!

Issue #314, opened 2016-01-02
Add Free Art License


xuv:
[Free Art License 1.3](http://artlibre.org/licence/lal/en/)

strugee:
@xuv what does this license cover that the licenses already on ChooseALicense don't? Note that we already suggest Creative Commons for non-code works.

xuv:
In terms of author's right, there is a difference between roman based code and english based code. This license was written in France and as such integrates a roman based version of the author's right such as the non-transferable moral rights while still being copyleft and compatible with a CC-BY-SA for example. Because of its European origin and particularity, it is sometimes the preferred choice for creators living in Europe over licenses originating from the US for example.

So although [compatible with the Creative Commons Attribution Sharealike](https://creativecommons.org/weblog/2014/10/21/44030), they are different in philosophy and motivation for their own existence. And as such, this difference matters a lot for some creators.

There is also significant usage of it on Github ([french](https://github.com/search?q=%22Licence+Art+Libre%22) and [english](https://github.com/search?utf8=✓&q="Free+Art+License") version) for it to be included.

benbalter:
@xuv thanks for starting the discussion. How would you explain to a developer releasing their first open source project (with no background in copyright or open source licensing) why they should prefer the free art license ver CC-BY-SA?

xuv:
@benbalter I would say it like this:
Compared to the CC-BY-SA, the Free Art License is much shorter text to read. It is written in a form that is easier to understand by anyone, without most of the usual lawyer vocabulary, while still having the same properties and benefits. The text of the license also ends with a user guide. All these benefits come because it was mostly written by artists and creators for themselves and their audience.

People might also prefer it because there is only one Free Art License and it does not require to understand the whole feature naming scheme of Creative Commons. Without any background in open source licensing, there is no doubt with the Free Art License about it's intent and use.

strugee:
The immediate concern I see is that it doesn't have any of the "usual lawyer vocabulary"... yes, that makes licenses harder to read, but it's there for a reason and I worry that the Free Art License would be harder to enforce in court. (That being said, I'm certainly not an expert.)

xuv:
@strugee "usual lawyer vocabulary" was maybe not the best comment from me regarding this license. It has been reviewed by lawyers and they have not made any remarks regarding this. It was a way for me to emphasize its easier reading. To compare, the MIT license is also easier to read than similar ones.

[...]

PR #619, opened 2018-10-31
Add Free Art License 1.3

blue-atom:
Add Free Art License 1.3

Related Issue: #314 

Free Art License 1.3 is the English translation of Licence Art Libre 1.3 ( spdx LAL-1.3)

**Help Wanted**

I need help for filling fields :- conditions, limitations and using

also please provide opinion on using english translation or original french version

### DIFF
--- a/_licenses/fal-1.3.txt
+++ b/_licenses/fal-1.3.txt
@@ -1,0 +1,167 @@
+---
+title : Free Art License 1.3
+spdx-id : FAL 1.3 (unofficial) , english translation of LAL 1.3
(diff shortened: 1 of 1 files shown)


mlinksva:
I'm going to close this because almost none of the bits in the file are what they should be, but feel free to open a fresh PR. A few things that should definitely be different in new PR:

- SPDX ID is LAL-1.3 and filename should be lal-1.3.txt
- Canonical text is French, use that https://github.com/github/choosealicense.com/issues/314#issuecomment-207038914 not an English translation
- If the text in https://github.com/spdx/license-list-data/blob/cbf30eced842598b0b951739a7ab7adcf762462f/text/LAL-1.3.txt is correct, copy that text, if not, submit correction to https://github.com/spdx/license-list-XML (we're slowly moving this repo to be aligned with SPDX sources, eventually will consume those directly)
- Copy permissions/conditions/limitations from https://github.com/github/choosealicense.com/blob/7eebfb905a78616838ab7ebc6eb8ab90d269f5c0/_licenses/cc-by-sa-4.0.txt to start with as it is a similar license

PR #627, opened 2018-12-09
Include Rust in community.md

JadedBlueEyes:
It's growing, and there are more and more people that are going to be using it. It's probably a good idea to let people know that most rusticians use the dual MIT/Apache-2.0 licence.

> The Apache license includes important protection against patent aggression, but it is not compatible with the GPL, version 2. To avoid problems using Rust with GPL2, it is alternately MIT licensed.
<https://doc.rust-lang.org/1.4.0/complement-project-faq.html#why-dual-mit/asl2-license?>


mlinksva:
Good idea @derpmarine168, thanks!

PR #628, opened 2018-12-27
Gitignored cname

RDIL:
Makes 100% sure it won’t change unless needed.


mlinksva:
Not sure I see the utility of this. Feel free to reopen if you can point to an explanation of why it is needed.

PR #632, opened 2019-01-06
Patch 1

AMDphreak:
Added context to the page about Licenses for noobs to licensing. Most software enthusiasts don't want to know much about licensing until they absolutely need it. This makes your page a resource for newcomers.

Added some clarification to terminology that is unfamiliar outside of the art industry ("provenance").

Made the description of AGPL actual English. The new one is a description and an English sentence whereas the old one was neither.


Tikhon88 (approved):

mlinksva:
I'm interested in less text not more. One can get a history of FLOSS many places.

PR #639, opened 2019-01-30
update title match actual name of license

ghost:
you might not like the title but that doesn't mean you can change the name


jhabdas:
p.s. your build broke

mlinksva:
We use the names cataloged at SPDX, indeed we have a test to ensure we're using those names. Feel free to take it up with them. Thanks!

jhabdas:
Nice hand wave.

ghost:
Thanks @mlinksva I [sent in a pull](https://github.com/spdx/license-list-XML/pull/764) to try and address the issue in a way I hope others will find acceptable while raising the visibility of this excellent license.

PR #669, opened 2019-06-28
Add the BSD-4-Clause license

alistair23:
This adds the BSD-4-clause license type.

1. SPDX Identifier: https://spdx.org/licenses/BSD-4-Clause.html
2. GNU's list of free licenses: https://www.gnu.org/licenses/license-list.en.html#OriginalBSD and FSF approved
3. Over 1000 projects: https://github.com/search?q=BSD+4-Clause+filename%3ALICENSE&type=Code
4. Yosemite Blockchain, mono/mcs/jay and LibreSSl use the BSD-4-clause license

This is based on the previous work of: https://github.com/github/choosealicense.com/pull/511

It appears that the matches fails to match any licenses as the BSD
licenses end up being different for every project. If this is merged we
can then work on updating projects to the the SPDX wording which should
fix the problem.

Signed-off-by: Alistair Francis <alistair@alistair23.me>

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,7 +16,7 @@ end
 
 group :test do
   gem 'html-proofer', '~> 3.0'
-  gem 'licensee'
+  gem 'licensee', git: 'https://github.com/licensee/licensee.git', branch: 'master'
   gem 'rake'
   gem 'rspec'
   gem 'rubocop'
--- a/_licenses/bsd-4-clause.txt
+++ b/_licenses/bsd-4-clause.txt
@@ -1,0 +1,62 @@
+---
+title: BSD 4-Clause "Original" or "Old" License
+spdx-id: BSD-4-Clause
+hidden: true
+
+description: A permissive license similar to the <a href="/licenses/bsd-3-clause/">BSD 3-Clause License</a>, but with a 4th clause (known as the "advertising clause") that requires an acknowledgment of the original source in all advertising material.
+
+how: Create a text file (typically named LICENSE or LICENSE.txt) in the root of your source code and copy the text of the license into the file. Replace [year] with the current year and [fullname] with the name (or names) of the copyright holders. Replace [project] with the project organization, if any, that sponsors this work.
+
+using:
+  - Yosemite Blockchain: https://github.com/YosemiteLabs/yosemite-public-blockchain/blob/master/LICENSE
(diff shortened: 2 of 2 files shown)


mlinksva:
Thanks! We need to find examples that can be detected before merging. Feel free to substitute others that would work or get the ones now to match.

alistair23:
It seems like the matching is really picky. I managed to get it to work with my [own license](https://github.com/alistair23/DashSight-Mezzanine-Card/blob/master/LICENSE), but it breaks as soon as there is a space in `[fullname]` or `[project]`. It looks like the [BSD 3-Clause](https://github.com/quilljs/quill/blob/develop/LICENSE) isn't so picky which I don't understand.

Am I missing something in the implementation?

It would be a lot easier to find projects that match if the matcher wasn't so strict.

mlinksva:
Multiple copyright lines (as in the BSD-3-Clause example you link to) are easy to parse out, templated text in the middle of the license is not, at least for the current implementation, which you can find in https://github.com/licensee/licensee.

alistair23:
Ah, it looks like licensee has special cases for each license. Without that it looks like we will have to fall back to more exact matching, hence the strictness I am seeing.

It looks like Licensee has to be updated first before the matcher will work here. They use the list of licenses from here though, so this seems like a chicken and egg problem.

mlinksva:
The tests here monkeypatch licensee to use the license texts in the branch being tested.

mlinksva:
Nevermind, I misunderstood the sequencing problem you were pointing out.

In any case licensee's special casing for some licenses is quite limited. This one presents a more general problem of fill-in fields in the middle of the license text. Handling that robustly requires much deeper work.

alistair23:
Ah ok. In which case it might be easier to patch current projects to use a more standard license. The problem then is finding projects that can have single names inside the license text.

I'm stating to see why this license wasn't submitted previously

alistair23:
With a WIP commit updating to use the master branch of licensee this now passes the tests :)

Can this be reviewed again? I assume you don't want to use the master branch of licensee so once the next release is out I'll update this and it should be ok to merge then.

mlinksva:
> use the master branch of licensee

Actually that's a reasonable idea to make permanent to reduce round trip between the two repos. I'll probably keep that.

PR #675, opened 2019-08-13
Sufiyang99

sufiyangk7:
ผมคนเดียว


mlinksva:
Hi, this looks like a Thai translation of GPL-3.0? We don't translate license texts in this repo. You might want to get in touch with the license's publisher, which has a process for unofficial translations, see https://www.gnu.org/licenses/translations.en.html

PR #679, opened 2019-08-26
Remove optional line in Unlicense

corollari:


mlinksva:
I don't think the optional line should be removed from this repo. AFAIK the optional line is usually included in actual uses of Unlicense, and in general the texts cataloged here should reflect the most common/standard usage. Feel free to add evidence to the contrary!

corollari:
You are right, but I would like to point that the fact that most current versions of the unlicense include the website link could be a result of:
- Github not detecting the unlicense without link properly.
- Github providing a version of the unlicense that includes the link in the drop-down that appears every time a repo is created
So it could be argued that including the link in the default version leads to a self-fulfilling prophecy in which the link version ends up dominating.

Still, it's true that currently most versions of the unlicense use the link so I agree that keeping the link follows the objective of this repo better and this PR should remain closed, just wanted to point out these problems.

PR #685, opened 2019-09-25
HTTPS is great

annevk:


nschonni:
Looks like the actual license text on https://unlicense.org/ uses HTTP in the URL, so it should probably stay as-is

annevk:
https://github.com/unlicense/unlicense.org/pull/55 got merged. @artob is there some other file that needs updating for the website text?

mlinksva:
The site needs to be deployed manually if I understand, it isn't using GitHub Pages, see https://github.com/unlicense/unlicense.org/issues/69#issuecomment-528602059

I'm OK with merging this PR here. Thanks!

PR #687, opened 2019-10-04
community breadcrumbs

bgr8:
breadcrumb added at community page


mlinksva:
Thanks but this is not needed -- the additional breadcrumb is needed on individual license pages because there's an intermediary licenses page. There isn't an intermediary for any other page on the site.

PR #695, opened 2019-11-12
Update licenses.html

Wulfheart:
There are some licenses missing like EUPL.


mlinksva:
Which fall within in the range represented.

PR #696, opened 2019-11-17
Add "Tor" to list of software using BSD-3-Clause

TheDcoder:


mlinksva:
Thanks for the suggestion, but the examples should be ones [that](https://github.com/github/choosealicense.com/blob/gh-pages/CONTRIBUTING.md#adding-a-license) "have straightforward LICENSE files which serve as examples newcomers can follow and that could be detected by licensee if it knew about the license." The example added in this PR is complicated. This is sometimes the case for large projects, nothing wrong with it, just doesn't serve as an example someone choosing a license for a new project can easily copy from.

TheDcoder:
@mlinksva I missed the part about having a straightforward LICENSE file as it was under the "Adding a license" section, thanks for explaning why Tor can't be added to the list.

PR #700, opened 2019-11-23
BSD-4-Clause tweaks

mlinksva:
Following up on #669 

- hidden is not needed as it is the default
- shorten description, don't refer to "4th clause" as the advertising clause is actually the 3rd clause

cc @alistair23


alistair23 (approved):

antdog1982:
Closed  #32

PR #702, opened 2019-12-03
Make rules with -- eg same-license--file appear lighter in appendix table

mlinksva:
Not needed elsewhere eg license view because actual rule is spelled out and no distinction with other licenses with non-attenuated versions of the rules is useful for eyeballing.

Some visual cue other than .5 opacity might be more useful, eg smaller or not filled in circle, but this is easy first cut given current sprite setup.

Here's what it looks like
image

(Boost in the license and copyright notice column, LGPL* in the same license column.)

Compare with live version of https://choosealicense.com/appendix/

### DIFF
--- a/appendix.md
+++ b/appendix.md
@@ -42,7 +42,12 @@ If you're here to choose a license, **[start from the home page](/)** to see a f
         {% for r in license[t] %}
           {% if r contains req %}
             <td class="license-{{ t }}" style="text-align:center">
-              <span class="{{ r }}">
+              {% if r contains "--" %}
+                {% assign lite = " lite" %}
+              {% else %}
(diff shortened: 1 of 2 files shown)


mtrackeros:
> Not needed elsewhere eg license view because actual rule is spelled out and no distinction with other licenses with non-attenuated versions of the rules is useful for eyeballing.
> 
> Some visual cue other than .5 opacity might be more useful, eg smaller or not filled in circle, but this is easy first cut given current sprite setup.
> 
> Here's what it looks like image
> 
> (Boost in the license and copyright notice column, LGPL* in the same license column.)
> 
> Compare with live version of https://choosealicense.com/appendix/

PR #709, opened 2019-12-24
Amend [L]GPL-3.0 how to apply and "hide" LGPL-3.0 from PR initiator

### LINKED ISSUE #607: Create clear standard instructions for adding LGPL to a GitHub repository
It would be ideal if all instructions and flows provided by GitHub matched each other and were what was created when you generate a new repository with a license template included or at least if choosealicense made it clear that you should ignore the GNU instrucitons and the basic license flows in the case of the LGPL, where a few extra steps should be taken.

1. GNU says to put the GPL in COPYING and LGPL in COPYING.LESSER
2. choosealicense says to put the GPL in LICENSE and LGPL in COPYING.LESSER -- Is this a mistake or did choosealicense intend for two different base file names?
3. Choosing the LGPL as the license template at repository creation generates only LICENSE with the LGPL (no GPL)
4. Following the GitHub blog instructions for adding a license to an existing repository generates only LICENSE with the LGPL (no GPL)

Quickly examining a handful of LGPL repositories, most didn't include the GPL anywhere.

I've seen users suggest using LICENSE and LICENSE.LESSER as these are a broader standard outside of the linux community where the standard is COPYING for the license file.

So, what is the standard for GitHub?  If I follow the instructions on choosealicense and use LICENSE and COPYING.LESSER, will the LGPL be properly detected as the repository license, or will it think the repository is under the GPL?

mlinksva:
As it requires multiple steps not supported by GitHub license picker, should be carried out by people who know exactly what they're doing.

Fixes #607 to extent it is an issue that can be addressed in this repository.

### DIFF
--- a/_includes/sidebar.html
+++ b/_includes/sidebar.html
@@ -1,6 +1,7 @@
 <div class="sidebar">
 
   <a href="#" data-clipboard-target="#license-text" data-proofer-ignore="true" class="js-clipboard-button button">Copy license text to clipboard</a>
+  {% unless page.spdx-id == 'LGPL-3.0' %}
   <h3 id="suggest-this-license">Suggest this license</h3>
   <div class="repository-suggestion">    
     <p>Make a pull request to suggest this license for a project that is <a href="/no-permission/">not licensed</a>. Please be polite: see if a license has already been suggested, try to suggest a license fitting for the project's <a href="/community/">community</a>, and keep your communication with project maintainers friendly.</p>
@@ -9,6 +10,7 @@
       <div class="status-indicator"></div>
     </div>
   </div>
+  {% endunless %}
 
   <div class="how-to-apply">
     <h3 id="how-to-apply">How to apply this license</h3>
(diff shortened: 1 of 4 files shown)


_includes/sidebar.html
```diff
@@ -1,6 +1,7 @@
 <div class="sidebar">
 
   <a href="#" data-clipboard-target="#license-text" data-proofer-ignore="true" class="js-clipboard-button button">Copy license text to clipboard</a>
+  {% unless page.spdx-id == 'LGPL-3.0' %}
   <h3 id="suggest-this-license">Suggest this license</h3>
   <div class="repository-suggestion">
```
waldyrious, line 6: Unrelated to this PR, but the whitespace at the end of this line could be removed.

jlumbroso:
@mlinksva As mentioned in https://github.com/github/choosealicense.com/issues/607#issuecomment-663137607, I think removing LPGPv3 from options is a paternalistic decision that is unnecessary and should be reverted. Thank you.

Florifulgurator:
"COPYING.lesser" seems the right thing. Why not add that to the help article.
It is UTTERLY BIZARRE that you need to study some source code if you want to correctly add LGPLv3

PR #735, opened 2020-05-16
Remove unnecessary (c) from MIT license

xtqqczze:
The copyright symbol `(c)` is unnecessary in the MIT license, also the Open Source Initiative does not use it in their copy: https://opensource.org/licenses/MIT

PR #737, opened 2020-05-19
Remove unnecessary (c) from MIT license

xtqqczze:
**Supersedes #735**

The copyright symbol (c) is unnecessary in the MIT license, also the Open Source Initiative does not use it in their copy: https://opensource.org/licenses/MIT

PR #738, opened 2020-05-21
Remove unnecessary (c) symbol from licenses

xtqqczze:
**Supersedes #735**

The symbol (c) is redundant when compbined with the word "Copyright"

### DIFF
--- a/_licenses/0bsd.txt
+++ b/_licenses/0bsd.txt
@@ -25,7 +25,7 @@ limitations:
 
 ---
 
-Copyright (c) [year] [fullname]
+Copyright [year] [fullname]
 
 Permission to use, copy, modify, and/or distribute this software for any
 purpose with or without fee is hereby granted.
--- a/_licenses/bsd-2-clause.txt
+++ b/_licenses/bsd-2-clause.txt
@@ -30,7 +30,7 @@ limitations:
 
 BSD 2-Clause License
 
-Copyright (c) [year], [fullname]
+Copyright [year], [fullname]
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
--- a/_licenses/bsd-3-clause-clear.txt
+++ b/_licenses/bsd-3-clause-clear.txt
@@ -27,7 +27,7 @@ limitations:
 
 The Clear BSD License
 
-Copyright (c) [year] [fullname]
+Copyright [year] [fullname]
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
--- a/_licenses/bsd-3-clause.txt
+++ b/_licenses/bsd-3-clause.txt
@@ -29,7 +29,7 @@ limitations:
 
 BSD 3-Clause License
 
-Copyright (c) [year], [fullname]
+Copyright [year], [fullname]
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
--- a/_licenses/bsd-4-clause.txt
+++ b/_licenses/bsd-4-clause.txt
@@ -28,7 +28,7 @@ limitations:
 
 BSD 4-Clause License
 
-Copyright (c) [year], [fullname]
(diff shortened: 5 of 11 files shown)


waldyrious:
> Follow-up to #735

And #737, for completeness.

xtqqczze:
> I agree it is redundant, since it plays the same role as the word "Copyright". Quoting Wikipedia's [Copyright notice § Form of notice](https://en.wikipedia.org/wiki/Copyright_notice#Form_of_notice):
> 
> > The notice for visually perceptible copies should contain all three elements described below. They should appear together or in close proximity on the copies.
> >
> > 1. The symbol © (letter C in a circle); the word “Copyright”; or the abbreviation “Copr.”
> > 2. The year of first publication (...)
> > 3. The name of the copyright owner (...)
> >
> > Example: © 2012 Jane Doe 
> 
> Note the "or" in the first item of the list. The `(c)` is one of three possible representations of the same concept. So unless the license requires usage of the "Copyright (c)" format, I agree we should generally normalize (all licenses, not just MIT) towards just "Copyright".

_Originally posted by @waldyrious in https://github.com/github/choosealicense.com/pull/737#issuecomment-632008293_

mlinksva:
I agree that (c) is redundant in the presence of Copyright. However, seeing all of these change (aside: is Zlib omitted intentionally?) I'm a little reluctant because:

- The redundant usage is nearly omnipresent in copyright notices with these licenses in the wild; people are used to that
- I don't consider it important to educate anyone on this redundancy
- The license texts in this repo aren't where reform happens, it's relatively trailing edge, reflecting what is in the wild

Happy to read about why this is an important change to make, here. Otherwise, I'd suggest taking the proposal to https://github.com/spdx/license-list-XML ... if it is accepted there, I'd take it onboard here immediately.

Another question: why not just "©" instead of just "Copyright"?

xtqqczze:
@mlinksva Thank you for pointing towards the [spdx/license-list-XML](https://github.com/spdx/license-list-XML) repo, I was not aware that it was on Github. I agree that would be the best place to make any changes.

waldyrious:
> Another question: why not just "©" instead of just "Copyright"?

I can think of a few reasons, off the top of my head:
- Ease of input
- Searchability
- Explicitness (spelled out)
- Readability ("copyright &lt;year&gt; &lt;author&gt;" can sort of be sounded out as a sentence)
- Precedence

Granted, none of these are particularly strong arguments, but perhaps cumulatively they command sufficient weight to make spelled-out "Copyright" a more reasonable choice? I don't actually have a strong opinion on this.

On a separate, but related note: if you would instead suggest using © rather than (c) or (C), then maybe that would be a more straightforward stance to adopt. The reason to use the latter was just a technical limitation; it doesn't bring any readability benefits, and even though it makes writing slighly more inconvenient, writing copyright notices is not something that one does very often anyway.

mlinksva:
@waldyrious I don't see explicitness, readability, or precedence as advantages for "copyright" over "©" -- how does one read the latter other than "copyright"? I'm not sure what precedence means -- in what sense does the word have precedence over the symbol. You're right about ease of input and search. I wasn't suggesting © rather than (c) though that probably makes sense as a tiny improvement.

I opened a dozen books near my desk and all but one had "Copyright © YYYY", and that one had "© YYYY".

I'm going to close this as I do think it'd be nice if OSS licenses skipped the word and/or replaced "(c)" with "©" but again, I'd rather consume that change from a consensus arrived at by SPDX rather than push it.

waldyrious:
I won't argue for "copyright" over "©" — as I said, my arguments were all pretty weak ones individually. I might explain my reasoning in more detail if you're curious, though.

> I wasn't suggesting © rather than (c) though that probably makes sense as a tiny improvement.

Is this also something you'd prefer SPDX to take the lead on, or would it be OK to do in choosealicense.com?

PR #750, opened 2020-06-28
Adding Hippocratic License Version Number: 2.1

jottenlips:
This license is from https://firstdonoharm.dev/ and I think it would make a great addition. 

This license is on the SPD list https://spdx.org/licenses/
It is included on the GNU list as well https://www.gnu.org/licenses/license-list.en.html.

It is used 426 times on Github. I know this isn't 1,000, but I think it is used enough for consideration and it is becoming more relevant as ethics in software become more of a concern in open-source. Thank you.

https://github.com/search?q=HIPPOCRATIC+filename%3ALICENSE&type=Code


jottenlips:
It is used by We transfer (https://github.com/wetransfer/zip_tricks), Ruby Gems,
(https://rubygems.org/gems/rack-read_only), and Gryphon (https://github.com/vinivendra/Gryphon).

nschonni:
It shows up on the "NonFree Software License" section https://www.gnu.org/licenses/license-list.en.html#NonFreeSoftwareLicenses, not as an endorsement

mlinksva:
As @nschonni says and see the note in https://github.com/github/choosealicense.com/blob/gh-pages/CONTRIBUTING.md#adding-a-license

> GNU's list of free licenses _(note: the license must be listed in one of the three "free" categories)_

There is a test for this, but I it's not reached in this PR because the added file doesn't have any metadata -- from the same link (for future reference):

> Each license has both [required and optional metadata](https://github.com/github/choosealicense.com#license-metadata) that should be included.

PR #751, opened 2020-07-08
Symfony and Sylius are using MIT license

jacquesbh:
🙃


mlinksva:
Good, so are millions of other projects, eg https://github.com/search?q=license%3Amit

MIT is not a license lacking good use examples 😄 and I'd rather keep the examples here to 3. If there's something lacking about one of the current examples (feel free to point out), happy to consider substituting.

jacquesbh:
It was worth a try!

I would have kept Symfony but it doesn't really matter :).

PR #759, opened 2020-07-24
Generalize software->licensed material in permission/condition/limitation descriptions

### LINKED ISSUE #743: The term "software" is used for cc-by-sa-4.0
The explanations for Permissions, Conditions and Limitations when the mouse hovers the items under each section use the word "software" whereas material should be used, here: https://choosealicense.com/licenses/cc-by-sa-4.0/

mlinksva:
Fixes #743

### DIFF
--- a/_data/rules.yml
+++ b/_data/rules.yml
@@ -1,14 +1,14 @@
 permissions:
-- description: Licensed material and derivatives may be used for commercial purposes.
+- description: The licensed material and derivatives may be used for commercial purposes.
   label: Commercial use
   tag: commercial-use
-- description: Licensed material may be modified.
+- description: The licensed material may be modified.
   label: Modification
   tag: modifications
-- description: Licensed material may be distributed.
+- description: The licensed material may be distributed.
   label: Distribution
   tag: distribution
-- description: Licensed material may be used and modified in private.
+- description: The licensed material may be used and modified in private.
   label: Private use
   tag: private-use
 - description: This license provides an express grant of patent rights from contributors.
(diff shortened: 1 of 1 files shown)


_data/rules.yml
```diff
@@ -1,43 +1,43 @@
 permissions:
-- description: This software and derivatives may be used for commercial purposes.
+- description: This licensed material and derivatives may be used for commercial purposes.
```
waldyrious: I'd say "licensed" is somewhat redundant (or rather, not very natural-sounding) when preceded by "this". How about omitting it in these cases, and leaving it for "the licensed material" only?

waldyrious: Alternatively, replace "this" with "the", which would make these sentences better match the structure of the ones in the "conditions" section.

mlinksva: Hmm, how about dropping "This"? It doesn't really make sense on choosealicense.com, and I'm not sure it helps when there is licensed software/material present either.

mlinksva: I made the above and a couple other changes. How does it look to you now @waldyrious?

waldyrious: Looks better, but personally I wouldn't say the "the"s were unhelpful. Without them the text feels more vague and legalese-like. Having the articles makes the text more concrete, and in a sense, more down-to-earth and approachable. Hope this makes sense.

mlinksva: I prefer it, but yes makes sense -- I re-added the "the"s in e94c302

waldyrious (approved):
LGTM, FWIW!

Issue #693, opened 2019-10-31
mention missing packages


peterducai:
please mention requirements before running bootstrap... on my Fedora 31 i had to install

> sudo dnf install zlib ruby-devel zlib-devel

thanks

peterducai:
also missing gems

```
gem install commonmarker -v '0.17.13' --source 'https://rubygems.org/'
gem install commonmarker -v '0.17.13' --source 'https://rubygems.org/'
gem install nokogiri -v '1.10.4' --source 'https://rubygems.org/'
sudo gem pristine eventmachine --version 1.2.7
sudo gem pristine ffi --version 1.11.1
gem install github-pages
gem pristine http_parser.rb --version 0.6.0
sudo gem pristine http_parser.rb --version 0.6.0
gem install github-pages
gem install html-proofer
gem install licensee
```

peterducai:
but even after that i'm stuck with licensee gem on

```
find_executable: checking for gmake... -------------------- yes

--------------------

find_executable: checking for cmake... -------------------- yes

--------------------

find_executable: checking for pkg-config... -------------------- yes

--------------------

/usr/bin/gmake
gmake: *** No targets specified and no makefile found.  Stop.
```

nschonni:
> please mention requirements before running bootstrap... on my Fedora 31 i had to install
> 
> > sudo dnf install zlib ruby-devel zlib-devel

That is a little buried but is a general Jekyll thing https://jekyllrb.com/docs/installation/other-linux/

A bunch of those look like they should have been covered by the Gemfile https://github.com/github/choosealicense.com/blob/gh-pages/Gemfile

mlinksva:
`script/bootstrap` as stated in https://github.com/github/choosealicense.com/tree/8a290cffad311dbb45db1074a0a42ffecfb2964f#run-it-on-your-machine should install the needed gems.

I don't think detailed platform specific requirements should be spelled out in this repo but I can imagine a link to https://jekyllrb.com/docs/installation/ could be helpful; feel free to make a PR.

mlinksva:
I think this was addressed by #765. If further documentation is needed, feel free to make a PR.

PR #765, opened 2020-09-03
Added documentation for installing dependencies and edge case with 'sudo' access in MacOS

sohitmiglani:
Adding this documentation because I had to figure out the dependencies and the 'sudo' access request myself in my MacOS system so I think this would be beneficial to other developers working on their local MacOS system.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,20 +16,20 @@ We catalog [select](CONTRIBUTING.md#adding-a-license) open source licenses with 
 
 ### Managing Dependencies
 
-It may be the case that your system doesn't have the required dependencies. You will need 'cmake' and 'make' installed on your computer. 
+It may be the case that your system doesn't have the required dependencies. You will need `cmake` and `make` installed on your computer. 
 
-* For MacOS, use Homebrew to update your dependencies (install Homebrew from <https://brew.sh/>):
+For MacOS, use Homebrew to update your dependencies (install Homebrew from <https://brew.sh/>):
 ```bash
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,6 +14,23 @@ We catalog [select](CONTRIBUTING.md#adding-a-license) open source licenses with
...
+brew install make cmake
+```
+* For Linux/Ubuntu, use the apt-get tool to install the dependencies:
+```bash
+sudo apt-get install make cmake
+```
+
+### Installing and Running the tool
+
+Note that the third command (./script/bootstrap) may require you to use 'sudo' access for installation. Add 'sudo' right before the third command, if need be, but don't use root access.
```
mlinksva: When would sudo be required to run the bootstrap script? It seems like if someone is changing file permissions or setting GEM_HOME, they know what they are doing.

mlinksva (commented):
Looks like a useful addition, only one comment.

Perhaps this should close #693?

README.md
mlinksva: Oh I see that you wrote:

>  I had to figure out the dependencies and the 'sudo' access request myself in my MacOS system

Curious why that was, but I'll take your word for it. 😄

sohitmiglani: I also had a similar issue like the one on #693. Regarding the sudo access, it's needed for the gems installation on Mac if you're not an admin user because gems get installed in /Library/ where sudo permissions are needed.

PR #779, opened 2020-10-03
Convert license `using` array to map

### LINKED ISSUE #758: license yaml format issue
The `using` part of the license.txt looks weird to me. It's an array of single-element map like this
```yaml
using:
  - youtube-dl: https://github.com/rg3/youtube-dl/blob/master/LICENSE
  - kakoune: https://github.com/mawww/kakoune/blob/master/UNLICENSE
  - RDF.rb: https://github.com/ruby-rdf/rdf/blob/master/UNLICENSE
```
why not just use a single map like this
```yaml
using:
  youtube-dl: https://github.com/rg3/youtube-dl/blob/master/LICENSE
  kakoune: https://github.com/mawww/kakoune/blob/master/UNLICENSE
  RDF.rb: https://github.com/ruby-rdf/rdf/blob/master/UNLICENSE
```
or an array of structered map
```yaml
using:
  - name: youtube-dl
    url: https://github.com/rg3/youtube-dl/blob/master/LICENSE
  - name: kakoune
    url: https://github.com/mawww/kakoune/blob/master/UNLICENSE
  - name: RDF.rb
    url: https://github.com/ruby-rdf/rdf/blob/master/UNLICENSE
```
It makes the deserialise code boilerplate, for e.g in Rust I have to declare this field like 
```Rust
using: Option<Vec<HashMap<String, String>>>,
```
When using this `using` field, I have to iterate the vector, during which I have to iterate the single-element map!
In other formats, it would be
```Rust
using: Option<HashMap<String, String>>,
```
or
```Rust
using: Option<Vec<SomeStruct>>,
```

ObserverOfTime:
Closes #758

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ Licenses sit in the `/_licenses` folder. Each license has YAML front matter desc
 * `spdx-id` - Short identifier specified by https://spdx.org/licenses/
 * `description` - A human-readable description of the license
 * `how` - Instructions on how to implement the license
-* `using` - A list of 3 notable projects using the license with straightforward LICENSE files which serve as examples newcomers can follow and that can be detected by [licensee](https://github.com/licensee/licensee) in the form of `project_name: license_file_url`
+* `using` - A map of 3 notable projects using the license with straightforward LICENSE files which serve as examples newcomers can follow and that can be detected by [licensee](https://github.com/licensee/licensee) in the form of `project_name: license_file_url`
 * `permissions` - Bulleted list of permission rules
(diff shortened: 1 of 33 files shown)


mlinksva:
Thanks @ObserverOfTime this LGTM but before I merge it I want to make sure it doesn't cause problems elsewhere, so I'm trying it out at https://github.com/licensee/licensee/pull/463 which vendors licenses from this repo.

jackromo888 (approved):

PR #780, opened 2020-10-06
Added Beerware.

jirian:
Added Beerware license: https://en.wikipedia.org/wiki/Beerware

### DIFF
--- a/_licenses/beerware.txt
+++ b/_licenses/beerware.txt
@@ -1,0 +1,31 @@
+---
+title: "THE BEER-WARE LICENSE"
+spdx-id: Beerware
+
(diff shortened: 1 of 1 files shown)


nicc938 (commented):
ser

mlinksva:
I don't think there's much chance this will ever meet our adoption criteria for cataloging here: https://github.com/search?q=filename%3ALICENSE+%22THE+BEER-WARE+LICENSE%22&type=code gets 0 hits.

It's also not reusable as-is, with an email address hardcoded. If you really want to promote it I'd suggest first taking getting replaceable bits annotated in https://github.com/spdx/license-list-XML/blob/master/src/Beerware.xml

PR #784, opened 2020-10-13
Capitalization, punctuation and spelling fixes

waldyrious:
- Harmonize capitalization in section headings:
  Use "Sentence case" instead of "Title Case", since the former style was more prevalent.
- Harmonize quotation marks and apostrophes:
  Use straight marks in source files (txt, md) and curly marks in rendered files (html).
- Fix a typo in the terms of service: "merchantibility" → "merchantability".


mlinksva:
> Harmonize quotation marks and apostrophes:
> Use straight marks in source files (txt, md) and curly marks in rendered files (html).

What is the rationale for harmonizing on straight marks in source files? I don't feel strongly about that outside of the license texts themselves, but I'd rather not touch those.

The other two changes look good, thanks!

waldyrious:
It's common to limit source files to ASCII-only characters, where possible, to facilitate editing, search/replace operations, avoid potential ambiguities in diffs, etc. I understand that the license texts themselves are not supposed to be edited, but in practice things like typo fixes (case in point!), line wrapping, indentation adjustments, etc. tend to happen, so it's best to keep them text-editor friendly. This is also recommended by style guides such as [Google's](https://developers.google.com/style/quotation-marks#straight-and-curly-quotation-marks) and [Wikipedia's](https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style#Quotation_characters). (As an aside, [this article](https://www.theatlantic.com/technology/archive/2016/12/quotation-mark-wars/511766/) may be an interesting read on this topic.)

By the way: these were pretty much the only non-ASCII characters (`[^\x00-\xFF]`) in the `*.txt` and `*.md` files of this repo. The other exception is dashes, which I'm happy to replace with `-` or `--` as well, since that's another source of inconsistency and ambiguity. For example, currently the only license text that uses dashes is `odbl-1.0.txt`, which uses `–` (that's an en-dash, by the way); all others use plain hyphens as `-` or `--`; and there are a couple instances of the en-dash in markdown files, but there's a single occurrence of an em-dash (`—`) in `about.md`. Everything else is plain ASCII. IMO it would be worth normalizing as well.

mlinksva:
The .html and .md files here each get rendered as HTML. I pushed another commit to use curly throughout, and reverted the license text changes. I'd prefer license text changes get made in https://github.com/spdx/license-list-XML first, and copy what happens there here.

UniLateralUSA (approved):

PR #787, opened 2020-11-01
Improve Language Comprehension

Apollo-Apostolos:
These changes help non-English speakers to read along.

Changing "but" to "and" suggests a correct continual pathos whilst adding clarity.
"But" is used to suggest a hard shift in mindset. It was misused.

Changing "them" to "those" clarified that the sentence is about fonts and not about people setting the license.
"Them" is used as shorthand for humans. It will confuse the reader.

### DIFF
--- a/non-software.md
+++ b/non-software.md
@@ -16,7 +16,7 @@ Any open source software license or open license for media (see [above](#data-me
 
 ### Fonts
 
-The [SIL Open Font License 1.1](/licenses/ofl-1.1/) keeps fonts open but allows them to be freely used in other works.
+The [SIL Open Font License 1.1](/licenses/ofl-1.1/) keeps fonts open and allows those to be freely used in other works.
 
 ### Mixed Projects
 


mlinksva:
The first change is great, thanks.

`Those` sounds unnatural to me, to the extent if I were reading this sentence for the first time, I'd wonder what `those` refers to. I don't see any confusion when reading `them` -- licensors aren't mentioned in the sentence. How about `licensed fonts`? Would love another opinion -- I am a native English speaker and may well be missing how most people would read the sentence.

Apollo-Apostolos:
You're Welcome : )

   What do you think about:
 "allows fonts to be freely used in other works."
      or
 "opens fonts for free use in other works."

These imply one concept (openness), with fewer words.

waldyrious:
> Changing "them" to "those" clarified that the sentence is about fonts and not about people setting the license.

While I see what you mean in the general sense, I don't think this confusion would happen in the specific sentence in question (emphasis mine):

> The [SIL license] keeps fonts open and allows **them** to be freely used in other works.

There are a couple reasons why the confusion wouldn't be likely:

1. The only subjects mentioned in the sentence are the _license_ and the _fonts_, and only one of them is plural, so "them" could only refer to the fonts;
2. Even if we were to assume people setting the license are an implied subject in the sentence (via the passive construction), the fact that "them" appears in the context of "allows them to be freely used in other works" means that for one to assume it was about people, they would have to consider the sentence to be about "using people in works", which wouldn't make sense.

IMO keeping "them" is the most natural way to phrase this sentence.

Maybe a more direct way to join the two clauses could help make this clearer — how about this?

> The [SIL license] keeps fonts open, allowing them to be freely used in other works.

Apollo-Apostolos:
I love it!
"Them" is good for comfort. Comfort is important.
Your final suggestion implies a consequence to a action. 
This reduced two narratives into one. It reduces the readers' cognitive load.
Wow, thank you for allowing me to engage on this.

mlinksva:
I'm fine with the final suggestion as well, thanks both!

mlinksva:
@Apollo-Apostolos would you like to push another commit?

mlinksva:
Incorporated this in #790, thanks!

PR #791, opened 2020-11-20
turn on bundler cache

mlinksva:
Try turning on gem caching for CI, noticed capability in https://github.com/github/government.github.com/pull/851

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,8 +20,5 @@ jobs:
         ruby-version: 2.7.1
         bundler-cache: true
 
-    - name: Install dependencies
-      run: ./script/bootstrap
-
     - name: Run tests
       run: ./script/cibuild
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,6 +18,7 @@ jobs:
       uses: ruby/setup-ruby@v1
       with:
         ruby-version: 2.7.1
+        bundler-cache: true
 
     - name: Install dependencies
       run: ./script/bootstrap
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -18,6 +18,7 @@ jobs:
       uses: ruby/setup-ruby@v1
       with:
         ruby-version: 2.7.1
+        bundler-cache: true
 
     - name: Install dependencies
```
nschonni: Think you can remove this call when the `bundler-cache` is on, since it also handles the install

mlinksva: Oh, thanks for catching that!

.github/workflows/test.yml
```diff
@@ -18,9 +18,7 @@ jobs:
       uses: ruby/setup-ruby@v1
       with:
         ruby-version: 2.7.1
```
nschonni, line 20: From that other PR, if you pull this into a `.ruby-version` this line can also be skipped

mlinksva: You're right, merged before I saw. `spec/ci_ruby_version_spec.rb` would also need to be updated. If you'd like to make a PR, go ahead, otherwise I'll get to it, probably another day.

UniLateralUSA: Let me know if you guys need me to do anything on my end. I believe it just needs a license so I'm learning as I go. Thank you so much guys!

mlinksva: On second thought, I'd rather not use `.ruby_version` because it introduces unnecessary (IMO) friction when developing locally. Not having the precise version installed has never caused me any problems. Limiting it to CI with ruby-version seems like a good compromise.

UniLateralUSA (approved):

PR #523, opened 2017-06-22
Add EUPL v1.2

MattiSG:
- Date of adoption: 18/05/2017.
- Date of effect: 08/06/2017.
- No end.

Mostly updates the list of compatible licenses.

PR #797, opened 2020-12-22
Update format of eupl-1.2.txt

bbenno:
Use the official UTF-8 format version from
<https://joinup.ec.europa.eu/sites/default/files/custom-page/attachment/2020-03/EUPL-1.2%20EN.txt>
that is linked on
<https://joinup.ec.europa.eu/collection/eupl/eupl-text-eupl-12>

Advantage: better readability (especially header and "_copyright notice for the Work: ..._" line)

### DIFF
--- a/_licenses/eupl-1.2.txt
+++ b/_licenses/eupl-1.2.txt
@@ -32,47 +32,56 @@ limitations:
 
 ---
 
-European Union Public Licence
-V. 1.2
+                      EUROPEAN UNION PUBLIC LICENCE v. 1.2
+                      EUPL © the European Union 2007, 2016
 
-EUPL © the European Union 2007, 2016
-
-This European Union Public Licence (the ‘EUPL’) applies to the Work (as
-defined below) which is provided under the terms of this Licence. Any use of
-the Work, other than as authorised under this Licence is prohibited (to the
-extent such use is covered by a right of the copyright holder of the Work).
+This European Union Public Licence (the ‘EUPL’) applies to the Work (as defined
+below) which is provided under the terms of this Licence. Any use of the Work,
+other than as authorised under this Licence is prohibited (to the extent such
+use is covered by a right of the copyright holder of the Work).
 
 The Work is provided under the terms of this Licence when the Licensor (as
 defined below) has placed the following notice immediately following the
-copyright notice for the Work: “Licensed under the EUPL”, or has expressed by
-any other means his willingness to license under the EUPL.
+copyright notice for the Work:
+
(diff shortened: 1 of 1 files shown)


mlinksva:
Thanks @bbenno it appears that there are some extra ‘ characters in our copy, not sure how we missed that before, appears to be in #523 which added EUPL-1.2.

Because of this I'll go ahead and merge, though I [usually](https://github.com/github/choosealicense.com/pull/784#issuecomment-727276600) prefer that changes are made first in https://github.com/spdx/license-list-XML/ if applicable and we align with that. I'd still encourage making sure https://github.com/spdx/license-list-XML/blob/master/src/EUPL-1.2.xml is updated -- at the least a crossref to the new source URL you mention seems appropriate.

bbenno:
> Thanks @bbenno it appears that there are some extra ‘ characters in our copy, not sure how we missed that before, appears to be in #523 which added EUPL-1.2.
> 
> Because of this I'll go ahead and merge, though I [usually](https://github.com/github/choosealicense.com/pull/784#issuecomment-727276600) prefer that changes are made first in https://github.com/spdx/license-list-XML/ if applicable and we align with that.

Oh, didn't know that. Sorry...
It seems like I skipped the corresponding section in [CONTRIBUTE](https://github.com/github/choosealicense.com/blob/gh-pages/CONTRIBUTING.md#adding-a-license) and only read the `Making changes` section.

> I'd still encourage making sure https://github.com/spdx/license-list-XML/blob/master/src/EUPL-1.2.xml is updated -- at the least a crossref to the new source URL you mention seems appropriate.

I will add the new crossref as suggested.

bbenno:
@mlinksva 
I am very uncertain about further changes in [spdx/license-list-XML](https://github.com/spdx/license-list-XML) as the changes of this #797 mainly introduced some blank lines and changed the bullet sign of the unordered lists.

The only that I could imagine applying to [spdx/license-list-XML](https://github.com/spdx/license-list-XML) would be the header centering :one: as well as the removal of list syntax at the bottom most lines :two:.

:one:: I think (but definitly not sure) the left align of the header might come from the `<p>` enclosure...

mlinksva:
Looking at the wdiff between the EC version now also in this repo and https://github.com/spdx/license-list-data/blob/2e20899c0504ff6c0acfcc1b0994d7163ce46939/text/EUPL-1.2.txt which is rendered from https://github.com/spdx/license-list-XML/blob/master/src/EUPL-1.2.xml it looks like the only change representable in the XML version would be replacing — with - for bullets.

PR #800, opened 2021-01-26
a bit of update

ghost:


mlinksva:
We don't modify license texts collected here. Thanks!

ghost:
ah i see.

PR #814, opened 2021-03-20
Make the { } on home page more visible

TrainerGamer:
Currently, the { } on the home page surrounding the h2 titles are barely visible. I have changed the colour to something that increases the contrast and makes them more visible.

### DIFF
--- a/assets/css/application.scss
+++ b/assets/css/application.scss
@@ -93,7 +93,7 @@ strong {
 }
 
 .home h2 span {
-  color: #f1f1f1;
+  color: #bfbfbf;
   font-size: 70px;
   padding: 0 5px;
   position: relative;


mlinksva:
I don't think they need to be more visible, they don't mean anything.

I'd be OK with dropping the {}'s entirely, but I'm not sure what design thing they were supposed to fulfill initially.

Feel free to comment with any further insight, anyone, but for now I'm closing this.

TrainerGamer:
I personally think they should either be removed or made at least a little more visible. At the moment they are kinda sitting on the fence and they kinda look out of place once you notice them.

MichaelAgarkov:
It's just a barely noticeable part of the design, it doesn't make it not be a design, and it's still somewhat visible, so there is no big point in this

waldyrious:
I tend to agree that we might as well remove them, since they have no function or meaning.

(Btw, cross-linking the [related discussion](https://github.com/github/choosealicense.com/pull/833#issuecomment-898684123) that took place recently in #833.)

PR #815, opened 2021-03-23
BSD-3-Clause description: copyright holder rather than project

mlinksva:
As [pointed out](https://twitter.com/thadguidry/status/1374422231606980609) by @thadguidry 🙇


mlinksva:
Test failures due to submodule seemingly not available in CI environment. I'll debug that separately.

Note it'll take awhile for this change to be vendored in a new licensee version and for that new version to be used on github.com.

Issue #761, opened 2020-08-08
<!-- removed -->


uqsme:

mlinksva:
MulanPSL-2.0 likely does not yet meet our usage requirements -- https://github.com/search?q=%22Mulan+PSL+v2%22+filename%3ALICENSE obtains 89 hits -- but should otherwise qualify. See https://github.com/github/choosealicense.com/blob/21b959fc2858835884a9fa2966a34f60502b0eeb/CONTRIBUTING.md#adding-a-license for details on the requirements.

Feel free to open a PR for adding this license, and we'll merge when all requirements are met.

ianharrd69:
I own it

TheInterestingSoul:
now 2289 hits. all [conditions](https://github.com/github/choosealicense.com/blob/21b959fc2858835884a9fa2966a34f60502b0eeb/CONTRIBUTING.md#adding-a-license) listed are met.

3 notable projects:
https://github.com/opengauss-mirror/openGauss-server
https://github.com/OSCPU/NutShell
https://github.com/gzu-liyujiang/Android_CN_OAID

mlinksva:
Fixed by #818

PR #818, opened 2021-04-04
Add the Mulan-PSL v2 license

uqsme:
1. SPDX Identifier: https://spdx.org/licenses/MulanPSL-2.0.html
2. OSI approved: https://opensource.org/licenses/MulanPSL-2.0
3. Over 1000 projects: https://hub.fastgit.org/search?q=%22Mulan+PSL+v2%22+filename%3ALICENSE
4. 3 notable projects founded, see the file

Chinese characters are wrapped to 50 characters each line width since they're much wider than English letters.

previous issue: #761


_licenses/mulanpsl-2.0.txt
```diff
@@ -0,0 +1,225 @@
+---
+title: Mulan Permissive Software License，Version 2
+spdx-id: MulanPSL-2.0
+hidden: true
```
mlinksva: This line is not needed, it's the default.

_licenses/mulanpsl-2.0.txt
```diff
@@ -0,0 +1,225 @@
...
+description: A permissive license similar to the <a href="/licenses/apache-2.0/">Apache License</a>, but explicitly states that it does not grant trademark rights. Mulan Permissive Software License is the first open source license in both Chinese and English approved by OSI.
+
+how: Create a text file (typically named LICENSE or LICENSE.txt) in the root of your source code and copy the text of the license into the file.
+
+note: It's suggested to take the additional step of adding a boilerplate notice to the top of each file. The boilerplate can be found at the end of the license.
+
+using:
+  - openGauss Server: https://github.com/opengauss-mirror/openGauss-server/blob/master/LICENSE
+  - OSCPU NutShell: https://github.com/OSCPU/NutShell/blob/master/LICENSE
+  - Android CN OAID: https://github.com/gzu-liyujiang/Android_CN_OAID/blob/master/LICENSE
```
mlinksva: Drop the `- ` before each example, see https://github.com/github/choosealicense.com/pull/779

This is what is causing CI to fail.

mlinksva (changes requested):
Thanks! A couple small changes in the metadata noted.

uqsme:
Oh Sorry, I'm not familiar with YAML, I just copied the code from #669 . I'm correcting it right now.

_licenses/mulanpsl-2.0.txt
```diff
@@ -0,0 +1,226 @@
+---
+title: Mulan Permissive Software License，Version 2
```
nschonni: Regular comma
```suggestion
title: Mulan Permissive Software License, Version 2
```

uqsme:
Comma form has changed.
There is still one failed example:
```
rspec './spec/license_meta_spec.rb[1:31:4:2:1]' # license meta The Mulan Permissive Software License，Version 2 license licensee detects using examples the https://github.com/OSCPU/NutShell/blob/master/LICENSE URL is a mulanpsl-2.0 license
```

the `LICENSE` file from this repo uses a incomplete license. The beginning part of `MulanPSL v2` (from [SPDX](https://spdx.org/licenses/MulanPSL-2.0.html)) is like this:
```
木兰宽松许可证, 第2版

木兰宽松许可证， 第2版
```
but they kept only one of the two lines. In this case, I can do nothing but remind repo maintainers. @wakafa1

_licenses/mulanpsl-2.0.txt
```diff
@@ -0,0 +1,226 @@
...
+
+Copyright (c) [Year] [name of copyright holder]
+[Software Name] is licensed under Mulan PSL v2.
+You can use this software according to the terms and conditions of the Mulan 
+PSL v2.
+You may obtain a copy of Mulan PSL v2 at:
+         http://license.coscl.org.cn/MulanPSL2
+THIS SOFTWARE IS PROVIDED ON AN "AS IS" BASIS, WITHOUT WARRANTIES OF ANY 
+KIND, EITHER EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO 
+NON-INFRINGEMENT
```
mlinksva: There's a missing  coma here, compare to https://license.coscl.org.cn/MulanPSL2/ and https://github.com/spdx/license-list-data/blob/38ecfa995d4d472090e6730e4edf0296c6a12ed8/text/MulanPSL-2.0.txt

I also think this line could be folded with the next one and still be 78 characters or less.

mlinksva (changes requested):

mlinksva:
Re the https://github.com/OSCPU/NutShell/blob/master/LICENSE example, the first two lines aren't the issue. That file has both, the first is just indented, which shouldn't cause a problem for detection. There are a few other minor differences such as no periods after some sub-bullets.

Most likely what is causing non-detection though is that the copyright holder and project are replaced in the license notice templates contained in the license, which increases the word difference. I'm not certain whether this replacement is intended by the license or not. For Apache-2.0 LICENSE files, the boilerplate notice is typically left as-is, with values only substituted when copied into source files. The instructions for MulanPSL-2.0 seem more ambiguous:

> How to Apply the Mulan Permissive Software License，Version 2 
> (Mulan PSL v2) to Your Software
> 
> To apply the Mulan PSL v2 to your work, for easy identification by 
> recipients, you are suggested to complete following three steps:
> 
> i. Fill in the blanks in following statement, including insert your software 
> name, the year of the first publication of your software, and your name 
> identified as the copyright owner;
> ii. Create a file named "LICENSE" which contains the whole context of this 
> License in the first directory of your software package;
> iii. Attach the statement to the appropriate annotated syntax at the 
> beginning of each source file.

The notice with substituted values is clearly intended to be put in source files, but the order may indicate that the substituted values (i) should be put in the LICENSE file (ii) as well. If that is the intended practice, then a few changes should be made to this PR:

- replace `[Year]` with `[year]`
- replace `[name of copyright holder]` with `[fullname]`
- replace `[Software Name]` with `[project]`

If it's not 100% certain, I'd suggest not making the above changes, and instead choosing a different example (which may be necessary anyway due to other small changes mentioned above). I can later consider treating the above fields as equivalent in the license detection program (licensee) to make detection of MulanPSL-2.0 a bit more forgiving.

mlinksva:
Apologies, forget everything I wrote above. https://github.com/spdx/license-list-XML/blob/master/src/MulanPSL-2.0.xml does not treat those fields as substitutable, and I'm trying to align with SPDX consistently. If those fields should be substitutable, they should be annotated as such in the SPDX XML representation.

So, please add a different example for now. Thanks!

_licenses/mulanpsl-2.0.txt
nschonni: Looks like it is the full width comma in the source, and may be intentional https://en.wikipedia.org/wiki/Comma#Languages_other_than_Western_European

mlinksva: The title here needs to match the name in https://spdx.org/licenses/licenses.json -- if that value is incorrect it should be fixed in https://github.com/spdx/license-list-XML/blob/master/src/MulanPSL-2.0.xml

uqsme:
Here I chose [Huawei LiteOS security function library](https://github.com/LiteOS/LiteOS/blob/master/lib/libsec/LICENSE) —— **not a whole project but a submodule**. However, I still think it's very representative.

_licenses/mulanpsl-2.0.txt
```diff
@@ -0,0 +1,226 @@
...
+Copyright (c) [Year] [name of copyright holder]
+[Software Name] is licensed under Mulan PSL v2.
+You can use this software according to the terms and conditions of the Mulan 
+PSL v2.
+You may obtain a copy of Mulan PSL v2 at:
+         http://license.coscl.org.cn/MulanPSL2
+THIS SOFTWARE IS PROVIDED ON AN "AS IS" BASIS, WITHOUT WARRANTIES OF ANY 
+KIND, EITHER EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO 
+NON-INFRINGEMENT,
+MERCHANTABILITY OR FIT FOR A PARTICULAR PURPOSE.
```
mlinksva: Thanks for adding the missing comma. I'd still suggest folding these two lines together, as they'll be under 78 chars, and the very short line with only `NON-INFRINGEMENT,` looks odd.

mlinksva (changes requested):

mlinksva:
Thanks @TheInterestingSoul, that does make the tests pass. I'd prefer an example though where the example LICENSE is the one used at the repo level, so that someone following the example doesn't get confused when they see that the detected license for the repo is something else (BSD-3-Clause in this case; licensing is confusing enough already, and choosealicense.com caters to newcomers). I also left one more formatting comment.

_licenses/mulanpsl-2.0.txt
uqsme: Sorry, I simply ignored it. I've corrected it.

uqsme: Sorry, I simply ignored it. I've corrected it.

mlinksva: That's not what I meant though. Now there's an even shorter line. 😄 

I meant 

```
KIND, EITHER EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO
NON-INFRINGEMENT, MERCHANTABILITY OR FIT FOR A PARTICULAR PURPOSE.
```

uqsme:
Well done. Thanks for constantly correcting mistakes.

_licenses/mulanpsl-2.0.txt
uqsme: 😅never mind

uqsme: wait

uqsme: Now works. thx

mlinksva (approved):

mlinksva:
🎉 @TheInterestingSoul thanks for your persistence. It's up at https://choosealicense.com/licenses/mulanpsl-2.0/

Issue #822, opened 2021-04-05
ruby-3.0 - no error bundle update, but now can't serve


ahmadrasyidsalim:
```
$ bundle exec jekyll serve
Configuration file: /home/ahmadrasyidsalim/choosealicense.com/_config.yml
            Source: /home/ahmadrasyidsalim/choosealicense.com
       Destination: /home/ahmadrasyidsalim/choosealicense.com/_site
 Incremental build: disabled. Enable with --incremental
      Generating...
                    done in 2.406 seconds.
jekyll 3.9.0 | Error:  no implicit conversion of Hash into Integer
/home/ahmadrasyidsalim/.rvm/gems/ruby-3.0.0/gems/pathutil-0.16.2/lib/pathutil.rb:502:in `read': no implicit conversion of Hash into Integer (TypeError)
```

mlinksva:
I can't replicate that, but I do get https://github.com/jekyll/jekyll/issues/8523 which I'll make a PR to fix.

ahmadrasyidsalim:
Even after webrick is installed, I still got the same
```
$ bundle exec jekyll serve
Configuration file: /home/ahmadrasyidsalim/choosealicense.com/_config.yml
            Source: /home/ahmadrasyidsalim/choosealicense.com
       Destination: /home/ahmadrasyidsalim/choosealicense.com/_site
 Incremental build: disabled. Enable with --incremental
      Generating...
                    done in 2.467 seconds.
jekyll 3.9.0 | Error:  no implicit conversion of Hash into Integer
/home/ahmadrasyidsalim/.rvm/gems/ruby-3.0.0/gems/pathutil-0.16.2/lib/pathutil.rb:502:in `read': no implicit conversion of Hash into Integer (TypeError)
```

R-BLIP885:
☆

mlinksva:
Ok, I was able to reproduce on Debian. Someone else had the same [problem](https://stackoverflow.com/questions/65539326/is-the-pathutil-ruby-gem-compatible-with-jekyll-v3-9-0-and-ruby-v3-0-0). Full stack trace below. The cause isn't obvious to me, but presumably a jekyll issue that can't be fixed in this repo.

```
jekyll 3.9.0 | Error:  no implicit conversion of Hash into Integer
[...]

PR #823, opened 2021-04-05
Add webrick to Gemfile so jekyll serve works under ruby 3

mlinksva:
As noted in https://github.com/github/choosealicense.com/issues/822


UniLateralUSA (commented):
Got it

PR #831, opened 2021-05-29
Improve sidebar reference links

ghost:
https://docs.npmjs.com/files/package.json#license no longer points to the correct destination and https://doc.rust-lang.org/cargo/reference/manifest.html#package-metadata doesn't link directly to the section where it talks about licenses.


mlinksva:
Thanks!

PR #838, opened 2021-07-18
Better consistency between MIT and derivatives

vaporoxx:
This PR improves the consistency between the MIT, MIT-0 and ISC license texts. The following changes were made:
- ISC: Remove the comma between the copyright year and name
- ISC: Increase the max line length to 80 (The no-warranty clause is now equivalent to the one of 0BSD)
- MIT-0: Add a (c) before the copyright year


mlinksva:
I'm OK with the ISC rewrap and removing the comma from the copyright line, which isn't specified either way in https://github.com/spdx/license-list-XML/blob/e0b157f45e17ff1a7b6aa8ad46fbfab05912a014/src/ISC.xml#L13 though I would like @waldyrious review before making this change.

I don't want to add a title to 0BSD unless and until it is done in SPDX, perhaps by https://github.com/spdx/license-list-XML/pull/1297

https://github.com/spdx/license-list-XML/blob/e0b157f45e17ff1a7b6aa8ad46fbfab05912a014/src/MIT-0.xml#L19 does specify and excludes (c) so I'd rather not add it. Plus, I'd rather move away from including (c) generally as it is redundant.

@nschonni I see you are 👎 feel free to expand on that here, particularly if you think the ISC change shouldn't be made.

Pending feedback from mentions above, my suggestion would be a new PR with only the ISC change.

vaporoxx:
> I don't want to add a title to 0BSD unless and until it is done in SPDX, perhaps by spdx/license-list-XML#1297

That makes sense, it's also a really lucky coincidence that there was a PR opened to add the title just a few days ago 😄 Maybe there could be a follow-up PR once the mentioned PR is merged or do you want it to be in an actual SPDX release first?

> https://github.com/spdx/license-list-XML/blob/e0b157f45e17ff1a7b6aa8ad46fbfab05912a014/src/MIT-0.xml#L19 does specify and excludes (c) so I'd rather not add it. Plus, I'd rather move away from including (c) generally as it is redundant.

I'm neutral on whether (c) should be added or not, as long as it is consistent across related licenses. If you think that it is redundant, maybe someone could work on removing them in other licenses at some point (at least if possible).

But apart from that, reverting everything except the ISC changes (or opening a new PR if preferred) would be fine with me after the requested reviews.

waldyrious:
FWIW, I have no objections to the proposed changes to the ISC license's text. Both 72 (the current wrapping limit) and 80 (the proposed one) are [pretty standard](https://en.wikipedia.org/wiki/Characters_per_line) line lengths in computing, and neither affects the legibility of the disclaimer text (unlike e.g. [semantic line-breaking](https://rhodesmill.org/brandon/2012/one-sentence-per-line/) would).

That said, you might want to propose the same change to the [ISC license text in the OpenSourceOrg/licences repo](https://github.com/OpenSourceOrg/licenses/blob/master/texts/plain/ISC) as well. It was initially wrapped in https://github.com/OpenSourceOrg/licenses/pull/43, probably following the wrapping used in ISC's own repositories, e.g. [here](https://gitlab.isc.org/isc-projects/dhcp/-/blob/5d0ff7ea7cc183d004752c08e002e965921d4d48/LICENSE), though (1) the original wrapping doesn't seem to have been strict in adhering to a particular line length (e.g. the "WITH" in line 9 or the "ANY" in line 11 would have fit in their preceding lines without affecting the maximum line length), and (2) the text "ISC" is replaced with "THE AUTHOR" in the OSI template, which changes the line length and thus would have produced a different wrapping anyway; so there's no special significance to the current wrapping at opensource.org.

pakarwebdesignku (approved):

pakarwebdesignku (approved):

mlinksva:
#845 has only the ISC change.

Issue #871, opened 2021-10-04
Extending the instructions for applying Apache 2.0 license


char0n:
Hello,

First of all, thank you for the project! I've been wondering if you'd be interested in PR which extends the instructions for application of Apache 2.0 license. I went through a research and I've written an article which clarifies couple of things around Apache 2.0 license. The PR would contain the information from the article.

https://vladimirgorej.com/blog/how-to-apply-apache2-license-to-your-open-source-software-project/

Have a great day!

mlinksva:
Depends on what you have in mind. If you have a _concise_ patch for the `how:` or `note:` fields I'm happy to consider it, though no promises.

Thanks for pointing out https://infra.apache.org/licensing-howto.html and the end of your post. Might be worth considering linking that.

char0n:
Hi @mlinksva,

> Depends on what you have in mind.

Well linking the article would be best. It describes the common patterns of Apache 2.0 miss-application a provides a guideline how to apply the Apache 2.0 license properly

 > If you have a concise patch for the how: or note: fields I'm happy to consider it, though no promises.

At least adding https://infra.apache.org/licensing-howto.html, to `how` field would be great. Not many people are aware of this document and a lot of wrong assumptions and made because of this. I can issue a PR if agreed.

mlinksva:
@char0n OK such a PR would be welcome.

char0n:
@mlinksva PR has been issued https://github.com/github/choosealicense.com/pull/905

mlinksva:
https://github.com/github/choosealicense.com/pull/905#issuecomment-986088515

PR #905, opened 2021-11-23
Provide more info for Apache-2.0 application

char0n:
Ref #871

### DIFF
--- a/_licenses/apache-2.0.txt
+++ b/_licenses/apache-2.0.txt
@@ -7,7 +7,7 @@ hidden: false
 
 description: A permissive license whose main conditions require preservation of copyright and license notices. Contributors provide an express grant of patent rights. Licensed works, modifications, and larger works may be distributed under different terms and without source code.
 
-how: Create a text file (typically named LICENSE or LICENSE.txt) in the root of your source code and copy the text of the license into the file. Consider adding a NOTICE text file with explicit Copyright Notice as well. More information about the license application is available in Apache Foundations [Assembling LICENSE and NOTICE files](https://infra.apache.org/licensing-howto.html) document.
+how: Create a text file (typically named LICENSE or LICENSE.txt) in the root of your source code and copy the text of the license into the file. Consider adding a NOTICE text file with an explicit copyright notice as well. More information about the license application is available in Apache Foundation's [Assembling LICENSE and NOTICE files](https://infra.apache.org/licensing-howto.html) document.
 
 note: The Apache Foundation recommends taking the additional step of adding a boilerplate notice to the header of each source file. You can find the notice in the appendix at the very end of the license text.
 
--- a/_licenses/apache-2.0.txt
+++ b/_licenses/apache-2.0.txt
@@ -7,7 +7,7 @@ hidden: false
 
 description: A permissive license whose main conditions require preservation of copyright and license notices. Contributors provide an express grant of patent rights. Licensed works, modifications, and larger works may be distributed under different terms and without source code.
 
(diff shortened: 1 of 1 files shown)


mlinksva:
Thanks. @ShaneCurcuru has occasionally commented in this repo. Shane, if you have any thoughts on/review of adding this link to ASF policy around LICENSE and NOTICE files, would be much appreciated. Not asking for an official statement. 😄

_licenses/apache-2.0.txt
```diff
@@ -7,7 +7,7 @@ hidden: false
 
 description: A permissive license whose main conditions require preservation of copyright and license notices. Contributors provide an express grant of patent rights. Licensed works, modifications, and larger works may be distributed under different terms and without source code.
 
-how: Create a text file (typically named LICENSE or LICENSE.txt) in the root of your source code and copy the text of the license into the file.
+how: Create a text file (typically named LICENSE or LICENSE.txt) in the root of your source code and copy the text of the license into the file. Consider adding a NOTICE text file with explicit Copyright Notice as well. More information about the license application is available in Apache Foundations [Assembling LICENSE and NOTICE files](https://infra.apache.org/licensing-howto.html) document.
```
waldyrious:
```suggestion
how: Create a text file (typically named LICENSE or LICENSE.txt) in the root of your source code and copy the text of the license into the file. Consider adding a NOTICE text file with an explicit copyright notice as well. More information about the license application is available in Apache Foundation's [Assembling LICENSE and NOTICE files](https://infra.apache.org/licensing-howto.html) document.
```

char0n:
Suggestion from @waldyrious accepted.

mlinksva:
Looking at this again, I've decided against merging. The doc doesn't correspond entirely with the added text, and is a bit confusing even for me:

- There's no instruction to add a copyright notice; it's just in the example at the end
-  It's unclear whether notices for bundled code are supposed to be added to LICENSE or NOTICE
- Requests for help are directed to an ASF committee, which doesn't seem appropriate for general consumption

A NOTICE file is not essential and I don't know that these instructions are helpful for people who don't already know OSS licenses well and are choosing one, ie audience for the site.

Thanks for the idea and feel free to tell me why I should reconsider.

ShaneCurcuru:
Apologies for the delay in review.

NOTICE files are only required in section 4.4 if the work you're then redistributing already contains a NOTICE file, so I agree this is a somewhat esoteric point to include for the targeted audience, who you're just trying to get to use **a** license, any one.

Also the various pages on apache.org about NOTICE are primarily very specific ASF policy for Apache projects, and is pretty heavy reading for someone outside the ASF just trying to use the license.

I agree this is not needed, given the audience.

Bigger picture (separately from this issue), I have two suggestions (that I'd help with if I weren't doing a new startup!):

- Having more resource links to _understanding_ software licenses would be a useful addition, beyond what's in the About - which is also a  little buried - having a top-level nav to "Understanding Licenses" or the like would be cool.
- Adding a FAQ section to each license might be useful.  It couldn't be definitive, but there are some common questions about some licenses that you could link to, even just to get users aware of the question.  For example, a common question for Apache and many copyleft licenses are: "Which other licenses is this compatible with?" Where you could link to the relevant [ASF](https://www.apache.org/legal/resolved.html#criteria) and [FSF](https://www.gnu.org/licenses/license-list.html#SoftwareLicenses) pages.

joseliaon:
Vladimir?

char0n:
Sure, sounds like a start. Original problem that is was trying to solve is - where to put Copyright Notice when using Apache 2.0 license. In majority cases, people tend to put the copyright notice in the text of the Apache 2.0 license, which is incorrect application of the license and breaks  its "Reusable without rewording" principle.

PR #910, opened 2021-12-23
Trademark Use

ghost:
Add an update about the licenses includes an implicit trademark restriction without permission, for those, not MPLv2 or Apache-licensed. Because for example Filezilla is under the GPL, yet it still can enforce its trademark restrictions without the GPL mentioning it.

### DIFF
--- a/licenses.html
+++ b/licenses.html
@@ -5,6 +5,8 @@ class: license-types
 title: Licenses
 ---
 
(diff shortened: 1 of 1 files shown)


mlinksva:
This is covered in the description of the trademark use field https://choosealicense.com/appendix/#trademark-use

PR #912, opened 2021-12-24
Remove space between "Share" and "Alike"

Tyler887:


mlinksva:
Our titles here match the name given in SPDX. If you think this change should be made please suggest it for https://github.com/spdx/license-list-XML/blob/master/src/CC-BY-SA-4.0.xml

PR #916, opened 2021-12-30
Add Hippocratic License 3.0

Tyler887:


mlinksva:
Does not meet requirements in https://github.com/github/choosealicense.com/blob/gh-pages/CONTRIBUTING.md#adding-a-license

PR #934, opened 2022-01-22
Remove "all rights reserved" from BSD-[23]-Clause reflecting recent SPDX update

Tyler887:
https://github.com/spdx/license-list-XML/commit/1c213e99b55163b388c8889ec5df3dfd2da1d5ca


mlinksva:
Fair enough. I see that https://github.com/spdx/license-list-XML/pull/1363/files also removes from BSD-3-Clause; so added a commit to reflect that here too.

Licensee will catch/has tests to ensure it will normalize away "all rights reserved" even if it isn't in the license text vendored from here, so good to go on that front.

PR #936, opened 2022-01-23
Modify Rules to match licenses

Tyler887:
Does not update the YAML tags, but the descriptions were changed, and some names were changed. The tags are left where they are.


mlinksva:
I'm not inclined to accept this. Apart from the editorialization/warning added, which I'd rather not have, unclear that the tweaks are improvements.

PR #937, opened 2022-01-23
Update rules.yml

Tyler887:
Adding the `claims` limitation. Will add to licenses that state it soon.

### DIFF
--- a/_data/rules.yml
+++ b/_data/rules.yml
@@ -54,3 +54,6 @@ limitations:
 - description: This license explicitly states that it does NOT provide any warranty.
   label: Warranty
   tag: warranty
+- description: This license explicitly states that you may not claim you own the licensed material.
+  label: Copyright claims against the holder
+  tag: claims


mlinksva:
It's not clear to me that this feature rises to level worth cataloging here. What licenses do you have in mind?

Tyler887:
> It's not clear to me that this feature rises to level worth cataloging here. What licenses do you have in mind?

zlib has this limitation, like how most other licenses catalogued on the site do.

mlinksva:
Thanks for explaining. I don't see the benefit of including this. We aren't attempting to do a fine grained description of every license feature. There are other projects that do that such as https://www.osadl.org/Open-Source-License-Checklists.oss-compliance-lists.0.html

mlinksva:
A better link is https://www.osadl.org/Access-to-raw-data.oss-compliance-raw-data-access.0.html

I would be interested in a crosswalk and/or tests to make sure this repo's descriptions of licenses aren't at odds with OSADL's though.

PR #943, opened 2022-02-23
Replace font-size absolute units with relative units

### LINKED ISSUE #904: Use relative units for fonts (rem, em)
Hello,

I see that this website uses pixel values on "application.scss" file. I suggest to convert pixel values used for fonts and spacing to relative fonts like "rem'" or "em". This will lead to accessibility of the site and rank higher on Google.

Thank you

aisgbnok:
## Summary
I noticed #904 and thought I would attempt to resolve it. I believe I translated all `font-size` CSS properties from `px` to `rem`, in the `application.css` file. 

If your browser is at the default font settings you should see no change in font size, and the site will look identical. If you modify your browser's font size you will now see the site's font size scale accordingly.

- Resolves #904 

### REM
Web browsers calculate REM from the root element size. By default, modern web browsers use  `16px = 1rem`. Therefore I performed `px/16 = rem`. 

Rem is a relative unit meaning it is adjustable by the browser or by modifying the root element in CSS. As an example, if a user changes their browser's default font size then rem will increase/decrease accordingly. Therefore this change improves user accessibility.

*[Mozzila Documentation on CSS values and units.](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Values_and_units)*
*[WC3/WAI - Draft Tutorial](https://www.w3.org/WAI/tutorials/page-structure/styling/)*

### DIFF
--- a/assets/css/application.scss
+++ b/assets/css/application.scss
@@ -7,7 +7,7 @@
 body {
   background: #fafafa;
   color: #5c5855;
-  font: 14px/1.4 "Helvetica Neue", Helvetica, Arial, sans-serif;
+  font: 0.875rem/1.4 "Helvetica Neue", Helvetica, Arial, sans-serif;
 }
 
 @media (prefers-color-scheme: dark) {
@@ -68,7 +68,7 @@ h1, h2, h3, h5 {
 }
 
 h1 {
-  font-size: 35px;
+  font-size: 2.1875rem;
   margin-bottom: 25px;
 }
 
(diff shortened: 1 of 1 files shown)


mlinksva:
Thanks @aisgbnok! At a glance the rendered version of this looks good to me. Further improvements or fixes welcome if needed.

mlinksva:
(The CI failure is due to an unrelated broken link, will address separately.)

PR #945, opened 2022-03-16
.NET Core > .NET

Tyler887:
According to https://en.wikipedia.org/wiki/.NET, the .NET foundation dropped the "Core" from .NET's name. .NET Framework remains using it's current name.

### DIFF
--- a/_licenses/mit.txt
+++ b/_licenses/mit.txt
@@ -10,7 +10,7 @@ how: Create a text file (typically named LICENSE or LICENSE.txt) in the root of 
 
 using:
   Babel: https://github.com/babel/babel/blob/master/LICENSE
-  .NET Core: https://github.com/dotnet/runtime/blob/master/LICENSE.TXT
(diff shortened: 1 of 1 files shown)


_licenses/mit.txt
```diff
@@ -10,7 +10,7 @@ how: Create a text file (typically named LICENSE or LICENSE.txt) in the root of
 
 using:
   Babel: https://github.com/babel/babel/blob/master/LICENSE
-  .NET Core: https://github.com/dotnet/runtime/blob/master/LICENSE.TXT
+  .NET: https://github.com/dotnet/runtime/blob/master/LICENSE.TXT
```
mlinksva: Additionally the branch name has changed.

```suggestion
  .NET: https://github.com/dotnet/runtime/blob/main/LICENSE.TXT
```

mlinksva (approved):
Thanks for catching this. Also confirmed by https://github.com/dotnet/runtime/pull/33694

PR #984, opened 2022-04-27
chore: remove git.io

renbaoshuo:
All links on git.io will stop redirecting after April 29, 2022.

- https://github.blog/changelog/2022-04-25-git-io-deprecation/

### DIFF
--- a/spec/license_spec.rb
+++ b/spec/license_spec.rb
@@ -28,7 +28,7 @@ describe 'licenses' do
 
       context 'industry approval' do
         it 'should be approved by OSI or FSF or OD' do
-          expect(approved_licenses).to include(spdx_lcase), 'See https://github.com/github/choosealicense.com/blob/master/CONTRIBUTING.md#adding-new-licenses.'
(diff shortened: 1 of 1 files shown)


spec/license_spec.rb
```diff
@@ -28,7 +28,7 @@
 
       context 'industry approval' do
         it 'should be approved by OSI or FSF or OD' do
-          expect(approved_licenses).to include(spdx_lcase), 'See https://git.io/vzCTV.'
+          expect(approved_licenses).to include(spdx_lcase), 'See https://github.com/github/choosealicense.com/blob/master/CONTRIBUTING.md#adding-new-licenses.'
```
mlinksva: Avoid another redirect:

```suggestion
          expect(approved_licenses).to include(spdx_lcase), 'See https://github.com/github/choosealicense.com/blob/gh-pages/CONTRIBUTING.md#adding-new-licenses.'
```

mlinksva: avoid master->gh-pages redirect

PR #1005, opened 2022-06-17
make the text visible on yellow :target background

ZacharyCrespin:
The white text was not visible on the yellow `:target` background so I made it darker.
(this is only a problem with dark mode)


mlinksva:
Thanks for the bug report and fix!

PR #1009, opened 2022-07-07
Add SerenityOS to BSD-2-Clause

detherminal:
Added the well-known repository [SerenityOS](https://github.com/SerenityOS/serenity) to BSD-2-Clause page.


mlinksva:
SerenityOS is a cool project but we only need 3 examples and the existing ones seem fine.

PR #1025, opened 2022-08-25
Add Grafana as AGPL-3.0 licensor

StEvUgnIn:
https://grafana.com/blog/2021/04/20/grafana-loki-tempo-relicensing-to-agplv3/


mlinksva:
Good idea, thanks! Added 2 more and removed test exception for AGPL-3.0 examples since now we have 3.

PR #1028, opened 2022-09-02
Add License: FPA General Code License

William-McGonagle:
Reference to the Fairfield Programming Association General Code License can be found here: https://about.fairfieldprogramming.org/licenses/code. 

The main benefit is that it separates the conditions between educational/ personal use and commercial use. The condition included in the commercial use section is significant modification- to prevent asset rips which are a problem in certain open source communities.

### DIFF
--- a/_licenses/fpa.txt
+++ b/_licenses/fpa.txt
@@ -1,0 +1,44 @@
+---
+title: Fairfield Programming Association
+spdx-id: FPA
+featured: false
+hidden: false
+
+description: A permissive license whose main conditions require preservation of copyright and significant modification for commercial use. All rights within educational and personal contexts are in the permissively worded MIT-style of license. 
+
+how: Create a text file (typically named LICENSE or LICENSE.txt) in the root of your source code and copy the text of the license into the file.
+
+note: The Fairfield Programming Association <a href="https://about.fairfieldprogramming.org/licenses/code/">states</a> that modifications to the license are allowed but should be done by legal professionals. 
+
+using:
+  Kubernetes: https://github.com/kubernetes/kubernetes/blob/master/LICENSE
+  PDF.js: https://github.com/mozilla/pdf.js/blob/master/LICENSE
+  Swift: https://github.com/apple/swift/blob/main/LICENSE.txt
+
+permissions:
+  - modifications
+  - distribution
+  - private-use
+
+conditions:
+  - include-copyright
+  - document-changes
+
+limitations:
+  - trademark-use
+  - liability
+  - warranty
+
+---
+
+Copyright <YEAR> <COPYRIGHT HOLDER>
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without limitation in the rights to use, copy, modify, merge, publish, and/ or distribute copies of the Software in an educational or personal context, subject to the following conditions: 
+
+- The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+
+Permission is granted to sell and/ or distribute copies of the Software in a commercial context, subject to the following conditions:
+
+- Substantial changes: adding, removing, or modifying large parts, shall be developed in the Software. Reorganizing logic in the software does not warrant a substantial change. 
+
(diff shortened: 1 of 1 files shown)


_licenses/fpa.txt
```diff
@@ -0,0 +1,44 @@
...
+hidden: false
+
+description: A permissive license whose main conditions require preservation of copyright and significant modification for commercial use. All rights within educational and personal contexts are in the permissively worded MIT-style of license. 
+
+how: Create a text file (typically named LICENSE or LICENSE.txt) in the root of your source code and copy the text of the license into the file.
+
+note: The Fairfield Programming Association <a href="https://about.fairfieldprogramming.org/licenses/code/">states</a> that modifications to the license are allowed but should be done by legal professionals. 
+
+using:
+  Kubernetes: https://github.com/kubernetes/kubernetes/blob/master/LICENSE
```
nschonni, line 14: These are all Apache 2.0 projects

nschonni (commented):
Does this actually meet the criteria for inclusion https://github.com/github/choosealicense.com/blob/gh-pages/CONTRIBUTING.md#adding-a-license

mlinksva:
It does not meet the criteria for inclusion here as @nschonni points out.

Just a personal comment, the license seems very limited and or unclear. Is it `(personal OR educational) AND noncommercial-unless-substantial-changes`? What are "asset rips"?

William-McGonagle:
Sorry- did not see the criteria about a 1,000 repository minimum. 

On another note @mlinksva, many open-source projects in areas like Game Development or Tooling want to provide their code while also not closing themselves off to revenue steam. This MIT-style license allows community contribution and uses in personal and educational settings while also not allowing corporations from taking the project and putting it onto stores themselves (a very common situation in open-source asset store projects). In other words, yes, it is `(personal OR educational) AND noncommerical-unless-substantial-changes`.

For instance, the FPA uses this on our open projects for games because it lets us create games with the community while still being able to put the game on steam and other platforms later without other people 'beating us to the punch' (because stores do not let people post the same project twice).

nschonni:
Limiting a particular group is a violation of 5, so it actually wouldn't be considered an open source license https://opensource.org/osd
If you've also modified an existing license without using a lawyer to validate it, you may not be protecting yourself as the resulting changes @mlinksva found confusing might also be found unenforcable by a court

William-McGonagle:
@nschonni Yes- the license is legally binding as it was written by the legal professionals at the Fairfield Programming Association. As per not fitting into one organization's open-source definition, it is disappointing to hear about the lack of innovation that these rather closed organization have. 

Also, discrimination against groups or persons (5) does not refer to contexts, it is referring to blocking of use based on who the people are, not by what they are using it for. On top of this, the MIT license does not fit the second clause as it does not require source code to be included.

mlinksva:
@William-McGonagle, Nick probably meant (6) fields of endeavor rather than (5). (2) pertains to a program, not its license. The OSD does not mandate a license require source code, but a program without source code is not OSS per that definition even it is under an open license.

OSD aside (clearly you have no intention for the license to conform to it, which is of course totally up to you), I'm idly curious about whether I didn't misread and it is not actually `(personal OR educational) OR noncommerical-unless-substantial-changes`. It isn't super clear to me which it is, but if it were `AND` I'm not sure what the commercial term is supposed to permit -- a for-profit educational institution to substantially change and use? I guess from your described use case above, you want to allow any commercial use if substantially changed, such as making a new game and selling it on an app store? (Just a question of idle curiosity/learning.)

mlinksva:
ps not that you shouldn't make your own license, and indeed it's not quite the same, but noting it's a bit reminiscent of https://creativecommons.org/licenses/sampling/1.0/ and https://polyformproject.org/licenses/perimeter/1.0.0/

William-McGonagle:
@mlinksva - Yes! I'm delighted to hear about your curiosity, though it may be idly 😉. The license is designed so that users can support their projects commercially while still making their code and knowledge public to the open-source world. The reason that the 'educational context' is included is because the Fairfield Programming Association is a non-profit focused on education, so we wanted to let educational institutions (schools, universities, etc.) use the projects without running into problems with the 'commerical' section of the license. We told this to legal, and they gave us the 'educational context' wording. From what they explained, the legal precedent would allow 501(c)(3) educational institutions and for-profit educational institutions to "use, copy, modify, merge, publish, and/ or distribute copies" but not "sell and/ or distribute". 

One example of this is the MongoDB project, which has always been in collaboration with the community and switched its license to SSPL in 2018. They ran into a bit of a problem when it came to Amazon because Amazon simply forked the project for its own commercial gain. If say, MongoDB choose a semi-permissive, non-commercial license (I'm not really sure how else to word it) to start the project, they would have been able to keep this (FPA) license and pair it with a their own commercial license to sustain the project. Amazon would also be able to then fork the project and license it as its own if it significantly modified the code (Adding new features, making it lightweight, changing around the underlying systems, etc.) to make it its own type of MongoDB server. 

I know it doesn't fit the case of MongoDB quite perfectly, but this would be how licensing with something like this would work. Also if a school or person wanted to host their own MongoDB server, they would fully be able to– outside of a commercial context (financially gaining).

Victorcorreiaaraujo (changes requested):
H

PR #1066, opened 2022-12-27
Add "Modified" nickname to BSD-3-Clause

mlinksva:
Note it has a number of nicknames https://en.wikipedia.org/wiki/BSD_licenses#3-clause_license_(%22BSD_License_2.0%22,_%22Revised_BSD_License%22,_%22New_BSD_License%22,_or_%22Modified_BSD_License%22) the first two of which are captured in the current title.

This change will facilitate matching a semi-common markup version, see https://github.com/licensee/licensee/pull/578


MustardseedX:
MustardseedX 

SigmaXIK

PR #1130, opened 2023-07-19
Update sudo usage instructions in README.md

nikohoffren:
This pull request aims to clarify the instructions related to using sudo in the README.md file.

The original instruction reads: "Add sudo right before the third command, if need be, but don't use root access." This can potentially confuse readers as sudo essentially provides root access.

To improve clarity, I propose changing this sentence to: "Add sudo right before the third command if needed, but don't run the script as root." This new phrasing emphasizes that the script should not be run as root while maintaining the information about when to use sudo.


mlinksva:
Seems like a reasonable change, but I wonder if we shouldn't instead strike the note about sudo entirely. It strikes me as likely bad advice in 2023. I was skeptical in 2020 when the text was added https://github.com/github/choosealicense.com/pull/765#pullrequestreview-482208362 but recall seeing pleadings to not use `sudo` with `gem` go past my eyes more recently.

nikohoffren:
Yeah, while some MacOS and Linux users might find the `sudo` instruction useful, i agree that it might be better to remove the note about `sudo` entirely and maybe provide more detailed guidance for different operating systems and situations. Linux users too sometimes need to use `sudo` for tasks like global installations, but it's generally recommended to avoid it when possible and there are works around that. Maybe we could perhaps include a note about this in the documentation, specifying that `sudo` should be used sparingly and wisely, and providing links to resources on managing permissions and environments correctly to avoid the need for `sudo`? Or maybe just scrap the whole `sudo` part all together.

I'm open to updating the pull request to reflect this approach if needed.

mlinksva:
Think I'd prefer to scrap the `sudo` part. If there's some guidance on how to manage environments provided ~upstream (say Jekyll or Ruby docs) that could be ultra concisely linked to, I'd be OK with that.

nikohoffren:
> Think I'd prefer to scrap the `sudo` part. If there's some guidance on how to manage environments provided ~upstream (say Jekyll or Ruby docs) that could be ultra concisely linked to, I'd be OK with that.

Deleted these lines from Installing and running the tool:
- Note that the third command (`./script/bootstrap`) may require you to use `sudo` access for installation. Add `sudo` right before the third command if needed, but don't run the script as root.

And added:
- If you encounter any issues with the above steps, please refer to the official [Jekyll](https://jekyllrb.com/docs/) documentation and this [guide on running Jekyll as a non-superuser](https://jekyllrb.com/docs/troubleshooting/#no-sudo) for more detailed installation instructions.

This approach follows the principle of providing minimum necessary instructions in your README, and linking to official documentation for more detailed guidance.

mlinksva:
Thanks!

PR #1133, opened 2023-08-11
Fix qtip viewport overflow.

canklot:
The qtips under the permissons was overflowing left of the screen and only half of the text was readable. To activate position.adjust.method plugin, I added the viewport option. Also changed default adjust method to shift instead of flip.


canklot:
Here are the screen shots of before and after:

before
after

assets/vendor/qtip2/jquery.qtip.js
```diff
@@ -1924,7 +1924,7 @@ QTIP.defaults = {
 			mouse: TRUE,
 			scroll: TRUE,
 			resize: TRUE,
-			method: 'flipinvert flipinvert'
+			method: 'shift shift'
```
mlinksva: Is changing the defaults in the vendored code necessary? Why not add to `app.coffee` as done with viewport?

canklot: I reverted my last commit and did the changes only on `app.coffe`

PR #1134, opened 2023-08-25
Update footer.html

scyilk:
It makes more sense to include the list of contributors instead of the repository, because "You!" entails the idea of people not their work.


mlinksva:
`You!` is an invitation to contribute, not just a credits list -- "you" wouldn't apply to anyone but contributors so we'd want a different noun, like "contributors" in that case. But I'd prefer to keep it as an invitation to contribute.

PR #1136, opened 2023-08-26
Drop special support for IE < 9

mlinksva:
Had reason to look at the js for this site, realized a chunk of it only exists to support Internet Explorer versions < 9. According to https://caniuse.com/usage-table IE8 has 0.03% market share, and below that, 0.

Seems reasonable to me to drop support for these ancient browsers. Would very mildly enjoy hearing reasons why not, so will leave this open for the time being.


mlinksva:
Not specifically for IE<9, nor js, but it seems maybe `normalize.css` could be tossed out as well? Many random comments along the lines of https://www.reddit.com/r/webdev/comments/10kqgju/how_relevant_is_normalize_css_in_2023/

Issue #1140, opened 2023-08-26
replace qtip2


mlinksva:
https://github.com/qTip2/qTip2 not maintained in a long time, and functionality probably no longer requires javascript. Might https://github.com/chinchang/hint.css be a replacement?

Thoughts or PRs welcome.

chahmedejaz:
Hi @mlinksva! I hope you are doing great. I did a spike on the issue and here are my findings:
- We **can** replace `qTips2` with `hint.css`
- However, we **cannot** remove JS from the functionality entirely:
  - We need to conditionally show the error qTip on the sidebar input here:
![image](https://github.com/github/choosealicense.com/assets/59338032/7f28c6ce-310e-4cc4-ae01-960fbab55a64)
- We might need to compromise the design of qTips as hint.css just render text in qTip, whereas we are rendering a **heading** (_Error_) and **text** (_Invalid url_) for the qTips. 

Overall, we can do this. Should I move forward with the issue?

mlinksva:
Thanks @chahmedejaz for doing the research/spike (I now see pure CSS [limitations]( https://github.com/chinchang/hint.css/wiki/Frequently-Asked-Questions))! The heading/message in two colors is used also or license condition tooltips eg on https://choosealicense.com/licenses/. I'd be OK with changing the design to only be one line/one color, perhaps with the current heading, colon, then the current message. So yes, please do move forward with the issue!

chahmedejaz:
Hey @mlinksva, the PR is ready for review. Would love to hear your thoughts on it. Thanks :)

mlinksva:
Fixed by #1141

PR #1141, opened 2023-08-27
1140, Replace QTip2 with Hint.css

chahmedejaz:
Hi there!
This PR replaces the QTips2 package functionality with the Hint.css as per issue #1140. 
**Here are the results after the replacement:**
image

image

**Here's the point to be noted:**
Previously we were showing the formatted error message for the repository URL input field error:
image

However, the hint.css package doesn't support the html tags in the tooltip body. Due to this limitation, we are now showing the error message in plain text:
image


mlinksva:
Thanks @chahmedejaz! I tried it out and have a few comments.

First, something changed with the /appendix page; the green/blue/red circles are now flush to the left of their table cells, rather than centered:

image

Second, in a very narrow browser window, the tooltips get cut off; this is not a huge deal as it was the case for the existing site until very recently https://github.com/github/choosealicense.com/pull/1133 but it'd be nice not to regress:

Screenshot 2023-08-28 at 6 29 47 PM

Third and least importantly, light text on darker color background (taken from the current heading?) strikes me as harder to read than darker color text on light color background (like the current message). Added: I might be wrong, guess it's a matter of taste.

chahmedejaz:
Thank you @mlinksva for the review. Let me address the comments and get back to you :smile:

chahmedejaz:
Hey @mlinksva - I've addressed your comments. Please review the fixes.
- [first comment](https://github.com/github/choosealicense.com/pull/1141/commits/94ec385ac93f3e383045645e4b41b70ccc883fb9)
- [second comment](https://github.com/github/choosealicense.com/pull/1141/commits/d4b0fadeb245b6e566d359a435cddf176ea90da6)
- [third comment](https://github.com/github/choosealicense.com/pull/1141/commits/5c0df8383145d9481f192d3bcbe64a727787693d)

**Results:**
Now the tooltip is responsive:
image

Tooltip colors:
image
image
image

The appendix page is as per the original site now:
image

Please let me know if I'm missing anything. Thanks.

mlinksva:
The screenshots look good, but when I run it, the tooltips have odd shadow or doubled text:

Screenshot 2023-08-30 at 5 19 41 PM

This is on a Mac FWIW, but I get the same on both Firefox and Chrome.

Any idea what could be causing that?

assets/js/app.coffee
```diff
@@ -16,18 +16,17 @@ class Choosealicense
...
-    adjust:
-      method: 'shift shift'
-  # Annotation rule types as defined in `_config.yml`
-  ruletypes:
-    permissions: "Permission"
-    conditions: "Condition"
-    limitations: "Limitation"
+  tooltipAttributesMapperByRuleType:
+    permissions:
+      heading: 'Permissions'
```
mlinksva:
```suggestion
      heading: 'Permission'
```

This should be singular, same for the two below.

assets/js/app.coffee
```diff
@@ -16,18 +16,17 @@ class Choosealicense
...
+      color: 'hint--success'
```
mlinksva: It looks like success/info/error are built into hint.css and while the colors basically match what's needed, I'd probably prefer to separately define permission/condition/limitation colors as success/info/error aren't semantically correct, even if it is just code the end user won't see.

chahmedejaz:
@mlinksva  - 

- Actually this effect is from the hint.css
- It is putting the shadow on the tool-tip body text and causing the doubled-text effect
- I've [overridden this shadow to none](https://github.com/github/choosealicense.com/pull/1141/commits/56122d32885e833c66702b3016e6f7700c126747) for our use and hopefully it will do the fix
- It's putting the shadow at my end as well but apparently it's more prominent at your end :sweat_smile: 

**Result:**
image

assets/js/app.coffee
chahmedejaz: @mlinksva  - Updated [here](https://github.com/github/choosealicense.com/pull/1141/commits/591dd6a388bbb05953dddce6979eac45af9b5b2b).

assets/js/app.coffee
chahmedejaz: @mlinksva  - Updated [here](https://github.com/github/choosealicense.com/pull/1141/commits/c71d2749bab23b306550b1066aa0541a81031c1a).

chahmedejaz:
Hey @mlinksva - I've addressed your comments, please review. Thanks.

chahmedejaz:
I noticed that the tooltip arrow colors were lost due to the new color class names. Updated the arrow color for the new classes [here](https://github.com/github/choosealicense.com/pull/1141/commits/0987bb0156a6144608de8fc99d5478d31f0c1c02)

.gitignore
```diff
@@ -5,7 +5,6 @@
 assets/vendor/clipboard/src
 assets/vendor/clipboard/test
 assets/vendor/selectivizr/tests
-assets/vendor/qtip2/basic
```
mlinksva, line 8: Seems like we could ignore hint.css/src?

mlinksva (changes requested):
Excellent, all looks like it's working. One final comment: think we could avoid vendoring hint.css source

.gitignore
chahmedejaz: Yes @mlinksva , we can. Sorry, It skipped my mind before :sweat_smile: 
Updated and removed the ignored source files from the branch as well.

mlinksva (approved):
🎉 @chahmedejaz, well done and thanks

PR #1143, opened 2023-08-31
Update unlicense.txt

HGStyle:
It justs adds Youtube DL as an example of Unlicensed software, since its a well-known software that is released under Unlicense.


mlinksva:
It is a well-known project, but we only need 3 examples for each license here; it's meant to be illustrative, examples to learn/copy from, rather than a larger list or the most popular projects for each license. Thanks for the suggestion!

waldyrious:
@mlinksva perhaps we could replace RDF.rb with youtube-dl? The former seems to be significantly less well-known than the others (by at least an order of magnitude in terms of GitHub stars).

PR #1153, opened 2023-09-15
Eclipse hawkBit is now licensed under EPL-2.0

shogo82148:
https://github.com/eclipse/hawkbit/commit/9c86729a68af97a23f56e4fbe9be38b863895981

### DIFF
--- a/_licenses/epl-1.0.txt
+++ b/_licenses/epl-1.0.txt
@@ -7,7 +7,6 @@ description: This commercially-friendly copyleft license provides the ability to
 how: Create a text file (typically named LICENSE or LICENSE.txt) in the root of your source code and copy the text of the license into the file.
 
 using:
-  Eclipse hawkBit: https://github.com/eclipse/hawkbit/blob/master/LICENSE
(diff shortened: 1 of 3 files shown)


mlinksva:
Thanks for noticing. We don't need another 2.0 example, but do need to replace hawkbit as a 1.0 example. https://github.com/clojure/clojurescript/blob/master/LICENSE seems like an obvious pick.

shogo82148:
> We don't need another 2.0 example

Right. We don't need 4 examples.
But there are other examples that should be removed.
I removed [Eclipse SmartHome](https://github.com/eclipse-archived/smarthome) because it is archived.

> https://github.com/clojure/clojurescript/blob/master/LICENSE seems like an obvious pick.

I agree. I added ClojureScript on https://github.com/github/choosealicense.com/pull/1153/commits/d9f9de4c6368f7d2f59faa26ef55b96bf722c6fe.

mlinksva:
Oh I did not see #1154 @shogo82148 glad we chose the same project.

PR #1162, opened 2023-10-30
Fix Code Search link in CONTRIBUTING

vorburger:
2.8M instead of 4 files under MIT license is more like it? ;-)


spboy777 (commented):
Texr

spboy777 (commented):
Test

PR #1170, opened 2023-12-16
Fixing indents for some licenses

ghost:


mlinksva:
The existing indents seem to match those at https://www.gnu.org/licenses/gpl-3.0.txt https://www.gnu.org/licenses/lgpl-3.0.txt https://www.gnu.org/licenses/lgpl-3.0.txt

PR #1175, opened 2023-12-28
Improve dark mode contrast on licenses page

Smankusors:
I decided to use the footer color because they uses the same color on light mode

before:
image

after:
image

also [remove unnecessary spaces](https://github.com/github/choosealicense.com/commit/0f4b11105153fdd0a76f60b0ef2715bc9f1bbb1b)
(the spaces was auto removed from my VSCode)


mlinksva:
Thanks, that does look better!

PR #1177, opened 2024-01-01
Implement static URIs for licenses

### LINKED ISSUE #1176: Raw license static url
Currently users have to visit the site, choose the license, copy it contents and put it into a new file.

I think it would be beneficial if only the content of a license would be available under a standalone uri.
It would allow for things like this:
```bash
$ curl https://choosealicense.com/licenses/agpl-3.0.txt > LICENSE
```

Happy new year 2024!
What do you think?

Dlurak:
This would close #1176.
This pull request includes changes to serve the raw text under a static URI.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ assets/vendor/hint.css/src
 /tmp
 Gemfile.lock
 .jekyll-metadata
+raw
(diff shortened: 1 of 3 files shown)


mlinksva:
This would only work for development purposes. choosealicense.com deploys as a static site, bootstrap is not run. As I said in #1176 actually serving a second URL for every license would require more restructuring.

PR #1202, opened 2024-03-04
Add CPAN to communities section

rawleyfowler:


mlinksva:
I don't think this is true as stated. As far as I can tell, eg https://libraries.io/search?order=desc&platforms=CPAN&sort=rank it's mostly Artistic 1.0 or GPL 1.0+. This is not a great example for others to learn from so I'd rather not feature it on the site.

rawleyfowler:
@mlinksva No worries. Thanks for your time :)

PR #1212, opened 2024-04-15
MIT.txt

Liam358:


mlinksva:
Thanks, we only need 3 examples per license, and would prefer they be relatively prominent.

PR #1229, opened 2024-07-06
Fix Typo in AGPL v3.0

GesundesWasser:


lambdakilo (commented):
Fair fix. There are some stronger theoretical copyleft licenses like Watcom-1.0 or SSPL but these have not gone through trial by fire in court or are only seemingly copyleft just for openwashing purposes.

CyberNFT (commented):
..

mlinksva:
Within the current scope of this project, it's the strongest.

PR #1234, opened 2024-07-20
Turn decor brackets in index into their own class.

TheBestWebsite:
I think it's a good idea to decorate the brackets on the home screen using class instead of span. This frees up span. The new class is decorBrackets.


TheBestWebsite:
Doesn't seem like community is interested - this pull request is STALE and has been closed

PR #1238, opened 2024-08-13
Improve readability of sticky header in appendix

### LINKED ISSUE #1110: Appendix table header is missing a background
The URL : https://choosealicense.com/appendix/

The bug :

image

matthijsgroen:
The header of the appendix table was transparent, making the text in the header hard to read when scrolling

Fixes #1110

### DIFF
--- a/appendix.md
+++ b/appendix.md
@@ -11,7 +11,7 @@ If you're here to choose a license, **[start from the home page](/)** to see a f
 
 <table border style="font-size: xx-small; position: relative">
 {% assign types = "permissions|conditions|limitations" | split: "|" %}
-<tr style="position: sticky; top: 0">
+<tr style="position: sticky; top: 0; background: color-mix(in srgb, var(--backgroundColor) 70%, transparent);">
   <th scope="col" style="text-align: center">License</th>
   {% assign seen_tags = '' %}
   {% for type in types %}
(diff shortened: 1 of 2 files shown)


matthijsgroen:
@mlinksva is this something you could verify?

mlinksva:
This does look nicer but it appears to me only the first column is affected:

Screenshot 2024-08-29 at 08 54 15

It isn't immediately clear to me why, any idea?

matthijsgroen:
It seems to be caused by:

https://github.com/github/choosealicense.com/blob/gh-pages/assets/vendor/hint.css/hint.css#L27

The position relative moves the dots out of the page flow and cause them to render on top of the sticky header.

Since this is a global stylesheet, I cannot forsee the impact of removing this `position: relative` for the rest of the site. It looks good without the position relative for the appendix page though.

**Update:** I just checked on other pages, but there the positioning of a hint following the dot will break.
They also work with really high z-index values. So I added a commit to set a z-index for this sticky header as well

matthijsgroen:
The z-index value is now based on  https://github.com/github/choosealicense.com/blob/3f5208ac0da4a3e9336013eadb0ff7f3e84e277b/assets/vendor/hint.css/hint.css#L62

mlinksva (approved):

mlinksva:
Tests are failing for unrelated reasons, will fix in a separate PR.

PR #1243, opened 2024-09-03
limit workflow permissions

mlinksva:
I think only read of git contents is needed to run tests.


majidkhosravi5:
Hello, work in the system, but do not try to change the main passwords, keep an eye on the site and Gmail

Issue #1256, opened 2024-11-25
Appendix: Legend does not correspond to colours used for same license and copyright notice


mschoettle:
The appendix uses different colour tones for the columns "same license" and "license and copyright notice".

However, in the legend on the bottom the same blue-ish colour tone is used for all items of those categories. It would be good to match what is used in the table.

mlinksva:
I see the same used for both columns:

![Image](https://github.com/user-attachments/assets/71a2a23f-d54d-4acb-8f7f-aee100c0250a)

Could you clarify?

mschoettle:
Thanks for the quick response. I should have added a screenshot in the issue right away.

<img width="692" alt="image" src="https://github.com/user-attachments/assets/6e97a61c-5387-4b19-afb7-cd33b70bf209">

See the different colour tones used in the "same license" column. 

The legend, however, looks as follows:

<img width="228" alt="image" src="https://github.com/user-attachments/assets/0dcbc0d5-884b-4167-beef-e7b0e863a94b">

Another small issue is that the same darker colour tone is used for "same license (file)" and "same license (library)" although that might be intended?

mlinksva:
Yes the muted colours are intended for lighter versions of conditions. They should be used in the key, thanks for noticing this discrepancy!

mlinksva:
Looks like the reason is there's a bit of special handling of conditions with `--` in their name only in the appendix table, should be in the key/legend as well. And probably elsewhere, but a start would be to make this page consistent.

https://github.com/github/choosealicense.com/blob/05070e5536dca04098bb3cc345dee859ec360ddf/appendix.md?plain=1#L45-L52

A pull request to fix would be welcome, or I'll get to it eventually.

mschoettle:
Thanks! I will see if I can create a PR for it.

Would it be ok to also add a third colour (in a separate PR) for "same license" so that each of the three are represented by a different one?

mlinksva:
Great!

Not sure I understand what you have in mind re adding a third colour. My intent is to have all conditions be blue, but lighter versions of conditions be a muted blue. Do you have a different scheme in mind?

mschoettle:
Ah, I see.

The issue I faced when reviewing differences between licenses was that "same license" has three different versions and only two different colours. The difference in "same license (file)" and "same license (library)" is very minimal so that it was difficult to figure out which of the two the darker dot belonged to. That's why I propose to introduce a third muted blue.

[...]

PR #1257, opened 2024-11-25
Show muted colors for conditions in legend

mschoettle:
As per discussion in #1256.

This PR adds the same classes for the different license conditions so that the colours match what is shown in the table.

image

### DIFF
--- a/assets/js/app.coffee
+++ b/assets/js/app.coffee
@@ -40,14 +40,15 @@ class Choosealicense
     # Dynamically add annotations as title attribute to rule list items
     for ruletype, rules of window.annotations
       for rule in rules
-        licenseLiElement = $(".license-#{ruletype} .#{rule["tag"]}")
+        # Only select license elements in table, not legend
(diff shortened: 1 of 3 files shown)


mlinksva (approved):

assets/js/app.coffee
```diff
@@ -40,14 +40,15 @@ class Choosealicense
     # Dynamically add annotations as title attribute to rule list items
     for ruletype, rules of window.annotations
       for rule in rules
-        licenseLiElement = $(".license-#{ruletype} .#{rule["tag"]}")
+        # Only select license elements in table, not legend
+        licenseLiElement = $("td.license-#{ruletype} .#{rule["tag"]}")
```
mlinksva, line 44: Belatedly noticing, I think this may have removed the tooltip from https://choosealicense.com/licenses/ and individual license pages.

mschoettle: Good catch! Fixed in #1259

PR #1258, opened 2024-11-25
Add rule type headers to legend

mschoettle:
Add a header above each category of rules to distinguish them. This follows the description above the legend.

image

Related to the discussion in #1256


mlinksva (approved):

mlinksva:
This does look nicer, but isn't strictly right as patent use can be a permission or limitation. Ideally the patent limitation would be rendered with the other limitations.

mschoettle:
Indeed, I missed that. Attempt of a fix in #1261 which makes the whole logic simpler at the same time.

PR #1261, opened 2024-11-26
Group rules by type in appendix legend

mschoettle:
Follow-up to #1258.

Changes the grouping of rules to group them by type such that patent use limitation is shown under the limitations heading.

### DIFF
--- a/appendix.md
+++ b/appendix.md
@@ -72,36 +72,24 @@ If you're here to choose a license, **[start from the home page](/)** to see a f
 
 <p>Most open source licenses also have <span class="license-limitations"><span class="license-sprite"></span></span> <b>limitations</b> that usually disclaim warranty and liability, and sometimes expressly exclude patents or trademarks from licenses' grants.</p>
 
-{% assign seen_tags = '' %}
 {% for type in types %}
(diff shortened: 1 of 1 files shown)


mschoettle:
One issue currently is that now there are two IDs `#patent-use`. The link in the table heading goes to the first one and highlights that one.

You can have a look here: https://opulent-journey-66577v5vp935w99-4000.app.github.dev/appendix/ Although it might become unavailable when the codespace is paused/disabled automatically.

mlinksva:
This is indeed simpler, but it loses muted coloring in the legend for `--` "light" features.

On the issue you note, I suppose ideally both would be highlighted. And for the same license column header, the whole range of same-license variations in the legend. However, these are just nice to haves in my book.

mschoettle:
> This is indeed simpler, but it loses muted coloring in the legend for -- "light" features.

Fixed.

> On the issue you note, I suppose ideally both would be highlighted. And for the same license column header, the whole 
> range of same-license variations in the legend. However, these are just nice to haves in my book.

That would indeed be nice. Unfortunately, it's not possible with anchors alone. Maybe with [text fragments](https://web.dev/articles/text-fragments) but it unfortunately cannot highlight all matches right now.

PR #1277, opened 2025-04-08
Mention that Chyrp Lite uses the BSD 3-Clause license

PPPDUD:
See title.
Chyrp uses the _BSD 3-Clause “New” or “Revised” License_, and I thought it would be helpful to put a link to them here.
Does not affect functionality.


mlinksva:
We have 3 more prominent examples already.

PR #1279, opened 2025-04-15
Fix repository url parser in LicenseSuggestion

### LINKED ISSUE #1278: Bug: License suggestion fails to parse URLs of repositories with ".git" in its name
License suggestion mistakenly truncates repository name when it contains ".git" (which is common in repositories for .github.io websites).

For example, if an user inputs repository urls like "https://github.com/octocat/octocat.github.io", license suggestion will falsely split url at the middle of repo name, drop ".github.io" and return the result of "octocat/octocat".

Image

Cycloctane:
Update regex. Replace `String.split` with `String.replace` to trim ".git" suffix.

Fixes #1278

### DIFF
--- a/assets/js/app.coffee
+++ b/assets/js/app.coffee
@@ -112,7 +112,7 @@ class LicenseSuggestion
 
   # Try to extract the repository full name from the user input
   parseUserInput: (userInput) ->
-    repository = /https?:\/\/github\.com\/([^\/]+)\/([^\/]+)\/?$/.exec userInput
+    repository = /https?:\/\/github\.com\/([^\/]+)\/([^\/]+)/.exec userInput
     [_, username, project] = repository
     return username + '/' + project.replace /(\.git)$/, ''
 
--- a/assets/js/app.coffee
+++ b/assets/js/app.coffee
@@ -112,14 +112,9 @@ class LicenseSuggestion
 
   # Try to extract the repository full name from the user input
   parseUserInput: (userInput) ->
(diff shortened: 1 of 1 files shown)


assets/js/app.coffee
```diff
@@ -112,14 +112,9 @@ class LicenseSuggestion
 
   # Try to extract the repository full name from the user input
   parseUserInput: (userInput) ->
-    repository = /https?:\/\/github\.com\/(.*?)\/(.+)(\.git)?$/.exec userInput
+    repository = /https?:\/\/github\.com\/([^\/]+)\/([^\/]+)\/?$/.exec userInput
```
mlinksva: Nice simplification but I don't see any reason to reject a URL with trailing garbage, say someone copies a repo URL that includes a filename or `?tab=...` which are pretty easy to get if not paying close attention. Wonder if simply matching up to the end of the second path item and ignoring anything after that wouldn't work.
```suggestion
    repository = /https?:\/\/github\.com\/([^\/]+)\/([^\/]+)/.exec userInput
```

mlinksva (changes requested):

assets/js/app.coffee
Cycloctane: I think it makes sense. I also added `?` and `#` to negated class in the second group to ignore query and fragment.

mlinksva (approved):

THAIuLINH01 (commented):
[](url)

PR #1280, opened 2025-04-15
Remove invalid html tags inside hint.css tooltips

Cycloctane:
html tags are not supported in tooltips after #1141 . This pr removes invalid tags.


mlinksva (approved):
Thanks, you're right &lt;b> is rendered as raw text without this change!
image

PR #1326, opened 2026-03-02
Change to recomment ISC over MIT license

jaalto:
Rationale:

The ISC license is functionally identical to the MIT and Simplified BSD licenses but excels in legal brevity: ISC is approx. 100 words, whereas MIT is approx. 170.

While established, the MIT license contains language specific to the Massachusetts Institute of Technology’s 1980s policies. It includes a specific grant for "sublicensing" that many legal scholars argue is redundant if the license already allows users to "deal in the software without restriction."

The Internet Systems Consortium (ISC) removed this and other unnecessary wording made redundant by the global adoption of the Berne Convention.

Bottom Line: ISC is essentially a stripped-down, modern refinement of the MIT license.

### DIFF
--- a/index.html
+++ b/index.html
@@ -32,10 +32,10 @@ permalink: /
       <h3>I want it simple and permissive.</h3>
     </a>
     <p>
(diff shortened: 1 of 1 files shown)


mlinksva:
Fair points but familiarity wins here, no intention to change.

PR #1341, opened 2026-06-01
Rename MIT to MIT Expat and change the license from CC-BY 3.0 to CC-BY 4.0

SapphireSlash:
According to the GNU website:

>Some people call this license “the MIT License,” but that term is misleading, since MIT has used many licenses for software. It is also ambiguous, since the same people also call the [X11 license](https://www.gnu.org/licenses/license-list.html#X11License) “the MIT License,” failing to distinguish them. We recommend not using the term “MIT License.”

This also means that the MIT No Attribution is renamed to the MIT Expat No Attribution.

Also from the GNU website:
> Some people call this license “the MIT No Attribution License,” but that term is misleading, since MIT has used many licenses for software and all of them contained requirements to preserve copyright notices and license notice. As the license is clearly based on the Expat License, we recommend to call it the No-attribution Expat License (but please note that Expat has never used this license).

Also changes CC-BY 3.0 license to 4.0.


mlinksva:
Do not resubmit the same PR, I already closed #1340 

Not going to rename, has been discussed years ago here and https://github.com/licensee/licensee/issues/98#issuecomment-220885877

PR #1230, opened 2024-07-08
Add new condition include-license-or-link

leoheitmannruiz:
Closes #1219

PR #1348, opened 2026-07-11
fix: correct Blue Oak 1.0.0 license notice condition

### LINKED ISSUE #1219: Blue Oak Model License 1.0.0 Conditions
Hello :)

The label for the conditions of the Blue Oak Model License 1.0.0 is "License and copyright notice", with the description being: "A copy of the license and copyright notice must be included with the licensed material."

If I'm not terribly misunderstanding the license, this label/description is not correct. The condition of the Blue Oak Model License 1.0.0 is stated in the [Notices](https://blueoakcouncil.org/license/1.0.0#notices) section.

No copyright notice must be included, as there is none. And, a copy of the license must not be included with the licensed material. A "link to https://blueoakcouncil.org/license/1.0.0" is fine.

Should a label with a description along the lines of "A copy of or link to the license must be included with the licensed material." be added?

PS: Do you know when GitHub will begin [detecting](https://github.com/git-learning-game/oh-my-git/blob/main/LICENSE.md) the Blue Oak Model License 1.0.0, now that licensee is [capable of doing so](https://github.com/licensee/licensee/releases/tag/v9.17.0)?

Tarekchehahde:
## Summary

Fixes #1219.

Blue Oak Model License 1.0.0 was tagged with `include-copyright`, which tells users they must include a copyright notice. The [Notices section](https://blueoakcouncil.org/license/1.0.0#notices) only requires a license copy or link — no copyright notice.

- Add `include-license` condition to `_data/rules.yml`
- Switch `BlueOak-1.0.0` to the new tag
- Document the rule in README

## Test plan

- [ ] `./script/cibuild` passes (license rules spec validates tags)
- [ ] Blue Oak license page shows "License notice" instead of "License and copyright notice"

Made with [Cursor](https://cursor.com)

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,7 @@ The license properties (rules) are stored as a bulleted list within the licenses
 
 * `include-copyright` - A copy of the license and copyright notice must be included with the software.
 * `include-copyright--source` - A copy of the license and copyright notice must be included with the software in source form, but is not required for binaries.
+* `include-license` - A copy of or link to the license must be included with the software.
 * `document-changes` - Changes made to the code must be documented.
 * `disclose-source` - Source code must be made available when the software is distributed.
 * `network-use-disclose` - Users who interact with the software via network are given the right to receive a copy of the source code.
--- a/_data/rules.yml
+++ b/_data/rules.yml
@@ -19,6 +19,9 @@ conditions:
 - description: A copy of the license and copyright notice must be included with the licensed material.
   label: License and copyright notice
   tag: include-copyright
+- description: A copy of or link to the license must be included with the licensed material.
+  label: License notice
+  tag: include-license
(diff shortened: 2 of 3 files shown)


leoheitmannruiz:
I think the proposed label and tag are rather poor. Half the point is that BlueOak-1.0.0 does not require a copy of the license and instead is fine with a link. The label "License notice" and tag "include-license" don't convey this at all. The majority of licenses require the license to be included but only very few are fine with a link.

I prefer something along the lines of #1230.

PS: Probably it makes sense to make a few larger changes, to better accommodate licenses such as BlueOak-1.0.0? I'm thinking perhaps remove the current include-copyright and then add these three:

- include-license: A copy of the license must be included with the licensed material.
- include-license--or-link: A copy of or link to the license must be included with the licensed material.
- include-copyright: A copy of the copyright notice must be included with the licensed material.

PR #1387, opened 2026-08-26
feat: collapse license text

Bishops-exe:
keep the text collapsed unless the user presses 'continue reading'. Translations done by Claude-code, no guarantee they are correct!


Bishops-exe:
here is how it looks
image

mlinksva:
I don't see any reason to add a click merely to read an entire license.

